PR #168, opened 2026-07-08
feat: add optional you.com search provider

mouse-value-add:
Why this is useful

Brigade already supports pluggable research search backends. You.com fits the existing provider pattern and gives the research flow another optional web source without changing defaults.

What changed

- Added a `youcom` provider in `src/brigade/research/sources/web.py`
- It calls You.com's Search API with plain HTTP, using `YDC_API_KEY` only when provided
- It merges web and news results into the existing `{url, title}` shape
- Added tests for the provider and factory aliases
- Documented the new provider in the README

Setup

```toml
[search]
research_search_provider = "youcom"
```

Optional env var:

- `YDC_API_KEY`

Usage example

Run Brigade research as usual after selecting the provider in `.brigade/research.toml`.

Validation

- `uv run pytest tests/test_research_web.py -q`
- `uv run ruff check src/brigade/research/sources/web.py tests/test_research_web.py`
- `uv run mypy src/brigade/research/sources/web.py tests/test_research_web.py`

Fallback behavior

The default provider stays `playwright`. The new provider is opt-in and only affects runs that explicitly select it.

Tracking issue

https://github.com/youdotcom-oss/integration-tracking/issues/87

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,7 +252,7 @@ The same review-and-receipt pattern covers the rest of an operator's day, and yo
 - **Daily loop**: `brigade work brief` shows pending work, imports, and warnings; `brigade daily status` keeps it bounded and cheap.
 - **Friction logs**: `brigade friction scan` mines recent notes, handoffs, and session artifacts for reviewable workflow friction.
 - **Security and scrub**: `brigade security scan` is a local read-only scanner for agent workspaces. `brigade scrub` gates content before it leaves the machine, and `brigade guard` exposes the embedded content guard CLI.
-- **Research**: `brigade research run` turns a question into a cited local report and a reviewable memory handoff.
+- **Research**: `brigade research run` turns a question into a cited local report and a reviewable memory handoff. Set `research_search_provider = "youcom"` in `.brigade/research.toml` to swap the web-search backend to You.com, or keep the default browser provider.
 - **Fleet and release**: health evidence across your local repos and release-readiness receipts, with no publish step.
 
 The full tour of every station lives in [docs/overview.md](docs/overview.md).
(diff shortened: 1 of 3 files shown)


solomonneas:
Thanks for the PR, but we are declining this and closing it.

### Why not

1. **No prior conversation on this repo.** CONTRIBUTING asks contributors to skim the guide so we spend time on the right things. There was no issue on `escoffier-labs/brigade` describing a user need, observed failure mode, or design for another search backend. The linked tracker is You.com's integration pipeline, not a Brigade request.

2. **This is not the kind of change CONTRIBUTING says lands easily.** Easy lands are bug fixes, harness/depth/include work, doctor checks, and tests for those. Optional third-party API backends are product surface; we would want a real operator story first, not a drive-by provider drop.

3. **Local-first defaults matter here.** Research already has Playwright (browser/DDG) and optional self-hosted SearxNG. Wiring a commercial SaaS search API as a first-class named provider needs an explicit maintainership decision (trust, secrets handling, doctor checks, docs), not a README one-liner.

4. **The PR description does not match the code.** It documents `YDC_API_KEY` as an env var, but Brigade only loads search settings from `.brigade/research.toml` via `search_settings()`. Nothing in this diff (or the config loader) reads that environment variable. Incomplete wiring is a review cost we do not want to take on for an unsolicited backend.

5. **Validation path does not match the project.** CONTRIBUTING and AGENTS.md point at `./scripts/verify` (ruff, format, version-sync, full pytest). Scoped `uv run` on two files is not our definition of done, and there is no research doctor / config check for the new provider the way SearxNG is treated.

6. **Pattern looks like mass integration spam.** Same author, same "optional You.com search" title, many repos, same tracking-org issue style. We are not a placement target for vendor OSS campaigns.

We are happy to review contributions that fix real Brigade bugs, improve harness adapters, or come from an issue we agreed to first. For a new research provider, open an issue with the operator use case before sending code.

solomonneas:
Closing as declined for the reasons in the review comment above.

PR #172, opened 2026-07-09
fix(agents): pass approval flags to headless grok so write tasks execute

solomonneas:
## Problem

Headless `grok -p` stops silently (exit 0) at the first tool-approval gate. Brigade's grok adapter passed no approval flag, so every grok worker on a write task reported `ok` while changing zero files - the run log showed only the model's preamble text ("Implementing the game and tests now.") and the repo diffstat was empty.

## Fix

- Write runs: append `--always-approve` (grok's equivalent of qwen's `--approval-mode yolo`).
- Read-only runs: use native `--permission-mode plan` instead of the soft prompt prefix.
- Reclassify grok in `READ_ONLY_ENFORCEMENT` as `hard`, since plan mode is a native guarantee rather than a prompt suggestion.

## Verification

- `python3 -m pytest tests/test_agents.py tests/test_agents_model_pin.py tests/test_read_only_enforcement.py -q` green via brigade verify receipt `20260709-050448-work-verify-0157e5` (full suite run before that: 1 failure = the stale model-pin assertion, updated here).
- Live repro: an all-grok roster (grok-4.5 chef + builder, grok-composer-2.5-fast speedrun) on a scratch repo. Before the fix: workers "ok", no files written. After: minesweeper.py + test_minesweeper.py landed, 11 pytest tests pass independently.

### DIFF
--- a/src/brigade/agents.py
+++ b/src/brigade/agents.py
@@ -98,8 +98,11 @@ def _openhands_argv(prompt: str, read_only: bool, sandbox: str | None) -> List[s
 
 
 def _grok_argv(prompt: str, read_only: bool, sandbox: str | None) -> List[str]:
-    task = _read_only_prompt(prompt) if read_only or sandbox == "read-only" else prompt
-    return ["grok", "-p", task]
+    # Headless `grok -p` stops silently (exit 0) at the first tool-approval
+    # gate, so a write task without an approval flag no-ops after its preamble.
+    if read_only or sandbox == "read-only":
+        return ["grok", "-p", prompt, "--permission-mode", "plan"]
+    return ["grok", "-p", prompt, "--always-approve"]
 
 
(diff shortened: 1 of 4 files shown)


solomonneas:
Empirical check of the read-only flag choice, since plan-mode-as-hard was the debatable call. Each mode was asked to create pwned_test.txt in a scratch dir using grok 0.2.77 headless:

| Invocation | File written? |
|---|---|
| `--permission-mode plan` | no |
| `--permission-mode plan --always-approve` | no |
| `--sandbox read-only` | **yes** |

Plan mode blocked the write even with `--always-approve` stacked on top, which points to a native gate rather than prompt compliance. The `--sandbox read-only` profile did NOT block workspace writes, so it is not a substitute here (it appears to scope network/filesystem outside the workdir, not workspace edits). Keeping `--permission-mode plan` and the `hard` classification.

PR #176, opened 2026-07-09
feat(research): add PageForge web provider with cached fetches

solomonneas:
## What

Adds a `pageforge` web-tier provider to the deep research lane. Search and page fetches route through the external PageForge CLI instead of headless Chromium.

## Why

The current web tier launches Chromium for every fetched page and throws the content away after the run. Routing fetches through PageForge gives research runs clean boilerplate-stripped markdown and banks every page in a local SQLite cache, so later runs on similar topics read from disk instead of the web.

## How

- `PageforgeProvider` in `sources/web.py`: `search` shells to `pageforge search_web --format json`, `fetch` chains `ingest_url` then `get --format markdown`. All subprocess failures degrade to empty results or a structured error instead of crashing the loop.
- Thin extractions (under 200 chars) fall back to the Playwright reader once when available; fallback failures keep the PageForge result.
- `build_provider` wires `research_search_provider = "pageforge"` with `pageforge_command`, optional `pageforge_db_path`, and `pageforge_timeout`.
- `research sources list/doctor` gains a pageforge row that validates the configured executable exists.
- Docs: config snippet and behavior notes in the technical guide.

## Verification

- `brigade work verify run --target . --command ".venv/bin/python -m pytest -q"`: 1729 passed, 1 skipped (receipt 20260709-061156-work-verify-040921)
- Live end to end against a local PageForge install and SearXNG: 3 search hits, fetch returned 165k chars of stored markdown
- The live smoke caught and fixed one integration bug the mocks missed: PageForge's CLI defaults search output to markdown, so the provider passes `--format json` explicitly

### DIFF
--- a/docs/technical-guide.md
+++ b/docs/technical-guide.md
@@ -476,6 +476,26 @@ The web tier needs the optional browser dependency, installed once:
 pip install 'brigade[research]' && playwright install chromium
 ```
 
+PageForge can serve as the web provider when you want searches and fetched pages banked in its local SQLite cache instead of reading each page through headless Chromium:
+
+```toml
+[search]
+research_search_provider = "pageforge"
+pageforge_command = ["node", "/path/to/pageforge/bin/pageforge.js"]
+pageforge_db_path = "/path/to/pageforge.sqlite"
+pageforge_timeout = 120
+```
+
+Then run:
+
+```bash
+brigade research run "latest on X" --web --provider pageforge
+```
+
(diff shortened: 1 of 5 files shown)


solomonneas:
Rebased onto origin/main: the branch had accidentally carried two unpushed local-main commits (the memory-doctor fold-in and a docs tweak), which caused 9 of the 10 mypy errors and the stale command-inventory failure. Also fixed the one genuine mypy error in the new provider (typed the fetch result dict in sources/web.py). PR now contains exactly the PageForge provider change; mypy, roadmap commands --check, and the research tests pass locally.

Issue #173, opened 2026-07-09
run: write-task workers that produce no file changes still report ok


solomonneas:
## What happened

An all-grok roster (before #172) ran a build task: "create minesweeper.py + tests and make them pass". Both workers hit the grok CLI's silent headless approval stop, wrote zero files, and the run still finished fully green:

- `worker-results.json`: both workers `ok: true`, `text` = model preamble only ("Implementing the game and tests now.")
- `ground_truth.diffstat`: empty except a pre-run roster edit; `changed_files` did not contain any of the files the task required
- synthesis: the chef also stopped at its first tool gate, so `final.txt` was one preamble sentence, and the run exited 0

So a run that did literally nothing was indistinguishable from success at every reporting layer, even though brigade had already captured the evidence (ground truth diffstat) that nothing happened.

#172 fixes the grok-specific cause, but the detection gap is generic: any worker CLI that dies quietly after emitting text (network drop mid-run, future headless regressions in other CLIs, models that just answer instead of acting) will hit the same blind spot.

## Suggested behavior

Brigade is "loud about exceptions". When a run is not `--read-only` and a worker's stage completes with:

1. `ok: true`, and
2. ground truth available, and
3. zero changed files across the whole run (or zero attributable to write-stage workers),

then at minimum warn in the run summary and `--inspect` output (e.g. `warning: write task completed with no file changes - worker output may be preamble-only`), and consider marking the worker result `detail: no-op` so synthesis sees it. A stricter option behind a flag could fail the run.

A cheap complementary heuristic: flag worker `text` that ends with an unfulfilled intent within the first N chars (very short output on a write task), since that is exactly what the silent-stop failure mode looks like.

## Repro

Roster with any CLI adapter stripped of its approval flag (pre-#172 grok reproduces it exactly), any write task, then compare `worker-results.json` `ok` fields against `ground_truth.diffstat`.

PR #180, opened 2026-07-09
feat(model): scorecard aggregating per-model outcomes from run artifacts

solomonneas:
## Why

Model choice across the fleet (gpt, opus, fable, grok, ...) has been judged by vibes. Every `brigade run` already records who sat in each seat (roster.json), whether each worker succeeded (worker-results.json), whether the run actually changed anything (ground_truth), and how long it took (run.json). This turns those artifacts into a per-model scoreboard, retroactively.

## Command

```
brigade model scorecard [--target DIR] [--runs-dir DIR ...] [--json] [--since YYYY-MM-DD]
```

Per (cli, model): runs participated, worker+orchestrator seats, worker ok rate, suspected no-op runs (worker reported ok but ground truth shows no changes outside `.brigade/` - the #173 signal), mean run duration, first/last seen. `--runs-dir` roots are additive to the target's default so one call sweeps multiple repos. Malformed dirs are counted and reported, never silently dropped. Read-only, no network.

## First real output (54 runs across 5 roots on this machine)

```
model                           runs  seats  w_ok  ok_rate  noop  mean_dur
codex/gpt-5.5                   45    85+45  85    100.0%   4     799s
codex/gpt-5.3-codex-spark       11    31+0   31    100.0%   4     1059s
grok/grok-4.5                   5     4+5    4     100.0%   1     290s
grok/grok-composer-2.5-fast     5     8+0    8     100.0%   2     290s
codex                           4     3+4    3     100.0%   0     53.6s
ollama:deepseek-v4-flash:cloud  2     1+0    0       0.0%   0     42.9s
```

The flagged grok no-op is the real pre-#172 silent-failure run, correctly detected.

## Caveat

For `--worktree` runs, brigade's ground truth watches the original cwd while changes land in the worktree patch, so such runs can flag as no-ops despite real output (one composer flag above is this). That is an artifact-recording question for #173 rather than a scorecard bug; the scorecard reports what the run recorded.

## Verification

- `python3 -m pytest tests/test_model_scorecard.py -q`: 20 passed
- `python3 -m pytest tests/ -q`: 1734 passed, 1 skipped (pre-existing codex appserver skip)

### DIFF
--- a/docs/technical-guide.md
+++ b/docs/technical-guide.md
@@ -377,6 +377,61 @@ When `--roster` is omitted, `brigade run` first reads `--cwd/.brigade/roster.tom
 if that file is missing, it falls back to `Path.home()/.brigade/roster.toml`.
 Passing `--roster` keeps using exactly that file.
 
+### Model scorecard
+
+`brigade model scorecard` aggregates run artifacts from `.brigade/runs` into a per-model summary. It is read-only: it never writes or networks. Rows group by `(cli, model)`; a missing or empty `model` is shown as the CLI alone (for example `codex`), otherwise as `cli/model` (for example `claude/claude-fable-5`). The scorecard reads the same run artifacts that `brigade runs show` inspects.
+
+Metrics per row:
+
+- `runs` - distinct run directories that seat the model.
+- `seats` - `worker_seats + orchestrator_seats` (printed as `workers+orchestrators`, for example `6+2`).
+- `w_ok` - worker results with `ok: true`.
+- `ok_rate` - `worker_ok / worker_seats` (0 if no worker seats); shown as a percentage.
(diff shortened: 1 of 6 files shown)


solomonneas:
Added orchestrator success columns (`orch_ok` / `orch_rate`) on this branch, closing the metric gap where a plan-stage failure (like the all-haiku DNF) dented no rate. Signal: `plan_attempts.json` records `parsed: true` only when an attempt produced accepted assignments. Models that never orchestrate render dashes. 21 scorecard tests, full suite 1787 passed / 1 skipped independently rerun (the worker's in-sandbox run showed the known 5 socket-permission failures, honestly reported). Built by a gpt-5.6-sol roster as part of the comparison series.

PR #218, opened 2026-07-12
feat: skill content fingerprints in the outcome ledger

solomonneas:
## Problem

`brigade outcome capture <skill-id>` records +1/-1 signals against a skill by name. When the skill's text is edited, the accumulated score keeps vouching for text that no longer exists (brigade-work on graphtrail: 109 helped / 14 hurt spanning many revisions). Promised publicly in the brigade.tools post "Applying CocoIndex to GraphTrail", section "One for the Python side".

## Fix

Modeled on CocoIndex's dual-fingerprint memo key (input hash + logic hash):

- **Capture stamps a fingerprint.** `outcome capture` and `outcome record` write `content_fingerprint` on each new record: the sha256 of the artifact's content at capture time. Skill content resolves registry-first (`.brigade/skills/registry/<id>/SKILL.md`), then harness installs (`.*/skills/<id>/SKILL.md`); cards use `memory/cards/<id>.md`. Unresolvable content means no fingerprint field, and the record loads exactly as before.
- **rank/explain are fingerprint-aware.** Default score covers only records matching the artifact's current fingerprint; lifetime counts stay visible. Records without a fingerprint form a distinct legacy cohort, fingerprinted records for other revisions a stale cohort. An edited skill earns its score back:

  ```
  - brigade-work score=0.207 helped=1 hurt=0 [rev ccc0fd7c52da; lifetime score=0.342 helped=2 hurt=0, stale=0 legacy=1]
  ```

  `explain` prints both scores and tags each trail entry `[current]`/`[stale]`/`[legacy]`. When the artifact content cannot be resolved locally, the split is unavailable, the default score falls back to lifetime, and both human and JSON output keep their pre-fingerprint shape.
- **Ledger stays append-only.** New field on new records only, riding the existing prev_digest/digest chain; old records are never rewritten. `receipts verify` passes over a mixed legacy+fingerprinted chain.
- **Ratchet untouched.** `reconcile` install/rollback/cooldown still fold over the full ledger with verified exit-code signals only, same posture as the delta-aware outcomes work in #163.

The core split (`split_by_fingerprint`, `fingerprint_cohort`) is pure in `outcome.py`; filesystem resolution lives in `outcome_cmd.py`. Cohort counts use the same evidence-dedup rules as scoring, so a re-captured receipt cannot inflate any cohort.

Known caveat, same one CocoIndex documents for undecorated helpers: the hash sees the skill's text, not the harness around it.

Also reformats `tests/test_aboyeur.py` (blank-line drift from #216) so `ruff format --check .` passes repo-wide.

## Verification

- Full suite through `brigade work verify run`: 2055 passed, 3 skipped, exit 0 (receipt `20260712-132124-work-verify-50d4b4`); outcome captured against `brigade-work` with the new code stamping the fingerprint.
- 13 new tests: cohort split/dedup/fallback in `test_outcome.py`; capture stamping (registry, harness fallback, absent artifact), card fingerprints via `record`, digest-chain verify, rank earn-back/legacy/unpinned shape, explain cohort tags in `test_outcome_cmd.py`.
- `ruff check .` and `ruff format --check .` clean.


solomonneas:
Review round before merge, four models plus CodeRabbit on the original diff:

- grok 4.5: sound against the five hard requirements (append-only capture, decide() untouched, legacy in lifetime totals, capture-time stamping, clean fixtures).
- GPT 5.5 (codex): one High, confirmed real. capture fingerprinted the registry copy while the verified run exercised the harness-installed copy, so a drifted install would pin signals to text that never ran. Fixed in d015849: installed copy resolves first, registry is the fallback, and capture and rank/explain share the resolver.
- deepseek-v4-pro: inconclusive, no findings.
- CodeRabbit: pass.

Also fixed in d015849: the lint failure (mypy could not narrow str | None through the pinned property) and a rollout defect the original cohort semantics carried: records captured before fingerprints existed were excluded from the current cohort, which would have zeroed every existing artifact's score at migration (brigade-work: 0.818 to 0.000 with no edit anywhere). Legacy records cannot be proven stale, so they now grandfather into the current cohort, and the rank line stays byte-identical until an artifact is actually edited. Full suite green through brigade verify (receipt 20260712-133052 and 20260712-134-series in the worktree ledger), 159 tests in the touched files.

PR #247, opened 2026-07-14
docs(design): spec route as an outcome cohort axis

solomonneas:
Spec, not implementation. Answers "how do we keep track of the route feature's performance now that 0.22.0 is live."

## The gap

The deterministic route brief (#210, #227, #232) ships in 0.22.0 and is exercised in live runs, but the outcome ledger is route-blind: nothing records whether a run followed a composed route, so `outcome rank` can't show whether routed runs verify greener than unrouted ones. The feature is used and asserted, not measured.

## The spec

Route is a third cohort axis, following the exact phased pattern the ledger already used twice:
- `content_fingerprint` (record #218, score #219)
- `capability_fingerprint` (record #228, score later, `docs/design/context-blind-spot.md`)
- **`route_fingerprint` (this spec)** — a coarse manifest (path + size + sorted signals, never the stage sequence) stamped on each `OutcomeRecord`, plus `followed` and `coverage`.

The linkage already exists in the plumbing: a verify receipt carries `run_id` → `run.json` → the `route` payload, and `outcome_cmd._resolve_run_receipt` already reads it. Phase 1 records + surfaces (no scoring change), Phase 2 splits `outcome rank --by-route` via the existing `split_by_fingerprint` + Wilson lower bound, Phase 3 defers route-arm paired attribution (the counterfactual, two exit codes no judge).

Full doc: `docs/design/route-outcome-cohort.md`. Sits beside `context-blind-spot.md` as its sibling axis. Cited functions (`split_by_fingerprint`, `_resolve_run_receipt`) verified present. Awaiting a green-light to build Phase 1 as its own PR.

### DIFF
--- a/docs/design/route-outcome-cohort.md
+++ b/docs/design/route-outcome-cohort.md
@@ -1,0 +1,116 @@
+# Route as an outcome cohort axis
+
+Status: spec, awaiting green-light. Not implemented. Sibling to
+`context-blind-spot.md`: a third cohort axis on the outcome ledger.
+
+## The problem
+
+The deterministic route brief (brigade #210, #227, #232) composes the stages a
+`brigade run` must cover, records the route in `run.json`, and ships in 0.22.0.
+It is exercised in live runs. What it is not is measured. The outcome ledger
+scores skills and cards by verified exit codes, but nothing in the ledger knows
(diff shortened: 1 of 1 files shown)


solomonneas:
Superseded by #248, which includes this design doc alongside the Phase 1 implementation.

PR #411, opened 2026-07-21
fix(agents): load proxy seat secrets from env files

solomonneas:
## Problem

Proxy-routed seats (`k3`, `luna`, `glm`, `muse`, `k27fast`) authenticate to the local CLIProxyAPI with `ANTHROPIC_AUTH_TOKEN_REF`. `resolve_env_overrides` only read the referenced value with `os.environ.get()`, so the token came from whatever the launching shell had exported.

The CLIProxyAPI key lives only in `/run/.../rocinante-secrets/workspace.env` (tmpfs, rotated every boot; the proxy restarts onto the new key). A shell that exported an older boot's key sent a stale token, so `brigade run` on the K3 seat failed with:

```
Failed to authenticate. API Error: 401 {"error":"Invalid API key"}
```

Observed in run `20260720-184120-e6e3ddba`, `logs/worker-001-k3_claudex.stdout.log`. The auth path itself is sound: a headless `claude -p --model kimi-k3` with the live key + the claudex config dir returns normally, and a direct request to the proxy with the current key returns 200 (a wrong key returns 401).

The rosters had already been switched to an `env-file:PATH#KEY` ref to read the live file, but that scheme was never implemented, so the seat instead failed even earlier with an `env-ref-missing` dispatch error.

## Fix

`resolve_env_overrides` now understands `env-file:/absolute/path#VARIABLE` refs and reads the value live from the file at dispatch, immune to per-boot key rotation. Anything else is still treated as a parent env var name. `roster.py` validates the ref shape at load time so a malformed pointer fails with a clear message instead of a runtime 401. Errors never contain the secret value.

## Tests

- `tests/test_agents.py`: env-file resolution, missing file/key, malformed ref.
- `tests/test_roster.py`: env-file ref validation.

`test_agents.py` and `test_roster.py` pass. Note: the full `./scripts/verify` run on this host currently flakes on two `test_work_cmd_sweeps.py` cases with `BlockingIOError: [Errno 11] Resource temporarily unavailable` from `subprocess` spawns under host fork pressure; both pass in isolation and are unrelated to this change.

## Follow-up

After merge, refresh the live pipx install (`pipx install --force`) so `brigade run` picks up the resolver.

## Summary by CodeRabbit

* **New Features**
  * Added support for `env-file:/absolute/path#VARIABLE` references to resolve environment overrides from referenced env files.
  * Per-agent and roster `env` entries now accept `env-file:`-formatted values when correctly formed.

* **Bug Fixes**
  * Improved failure messaging when a referenced env-file value can’t be read/found.
  * Prevented malformed `env-file:` strings from falling back to parent environment lookups.
  * Ensured only the exact `env-file:` syntax is treated as an env-file reference (e.g., `env-filed` is handled normally).

* **Tests**
  * Added coverage for success, missing referenced files/values, malformed formats, and output scrubbing.


solomonneas:
@coderabbitai review

solomonneas:
@coderabbitai review

Applied: validation and resolution now share one classifier (agents.is_env_file_reference + ENV_FILE_REF_RE); a malformed env-file reference fails dispatch with env-ref-missing instead of falling through to a parent-environment lookup, with a regression test planting a matching parent variable to prove it is never read. Also merged main into the branch (ordinary merge). ./scripts/verify green through brigade work verify run.

solomonneas:
@coderabbitai review

Applied: the discriminator is now the exact `env-file:` prefix. An `env-file`-prefixed, non-colon parent variable reference resolves from the environment (new regression test with `env-filed`), while malformed colon-prefixed references keep failing dispatch with env-ref-missing. ./scripts/verify green through brigade work verify run.

solomonneas:
@coderabbitai resolve

PR #414, opened 2026-07-21
fix(hooks): reconcile and remove legacy standalone work-loop hooks (#397)

### LINKED ISSUE #397: Legacy Claude work-loop hook can OOM on large untracked files after upgrade
## What happened

An upgraded Brigade installation reported `brigade 0.24.0`, but Claude Desktop was still running the legacy standalone work-loop hook registered as:

```text
~/.claude/hooks/brigade-work-loop.py
```

A read-only shell command mentioned several absolute directories. `_repo_for_event()` selected the first path that resolved to a Git repository, even though that repository was incidental to the session's working directory. On the first event for that repository, `_work_fingerprint()` enumerated its untracked files and hashed each one with:

```python
digest.update(path.read_bytes())
```

One untracked model-cache blob was 10,672,390,984 bytes. The hook buffered it into memory, entered `mem_cgroup_handle_over_high`, and stalled Claude long enough for Electron's hung-child watchdog and systemd-oomd to kill the application.

Observed sequence:

```text
11:33:15 legacy brigade-work-loop.py started
11:34:43 FATAL:content/child/child_thread_impl.cc:377] Crashing because hung
11:35:05 systemd-oomd: Current Memory Usage: 8.2G
11:35:05 systemd-oomd: Killed ... due to memory pressure
```

The Claude scope had `MemoryHigh=8G` and `MemoryMax=10G`. A controlled reproduction of `_work_fingerprint()` against the same repository reached 8.45 GB RSS before termination. No network filesystem or disk error was involved.

I expected a read-only hook invocation to use bounded memory and an upgrade to remove or replace obsolete hook registrations. Instead, a stale legacy hook survived the upgrade and allocated memory in proportion to the largest untracked file in a repository mentioned by the command.

## Reproduction

1. Start with an installation that registered the legacy standalone `brigade-work-loop.py`, then upgrade Brigade to 0.24.0.
2. Confirm the old hook command remains in `~/.claude/settings.json`.
3. Create a Git repository with a large untracked file, for example a sparse file under `cache/`.
4. From another Brigade-wired working directory, run a Bash tool command that mentions the repository path.
5. Observe the legacy hook select that repository and call `_work_fingerprint()` before the tool runs.
6. Observe hook RSS grow toward the logical size of the untracked file.

## Contributing defects

1. Legacy fingerprinting reads each untracked file into one `bytes` object instead of hashing it in bounded chunks.
2. Upgrade or wiring reconciliation leaves the obsolete standalone hook registered after the managed v0.24 hook implementation has replaced that code path.
3. Multi-path shell-command attribution can enroll the first incidental repository mentioned by a command, causing an expensive baseline fingerprint outside the session's main target.

## Acceptance criteria

- [ ] Upgrade and wiring reconciliation detect legacy standalone Brigade hook entries and replace or remove them without duplicating active hooks.
- [ ] The migration is idempotent and preserves unrelated user hooks.
- [ ] Any retained file fingerprinting hashes through bounded-size chunks. Peak RSS does not scale with the largest tracked or untracked file.
- [ ] A regression test covers a large or mocked-large untracked file without allocating its full size.
- [ ] A regression test covers a Bash command that mentions multiple repositories and proves incidental path arguments do not enroll the wrong repository.
- [ ] Doctor or status reports a stale legacy hook registration with an exact repair command.
- [ ] Documentation names model caches, virtual environments, and generated databases as directories that should normally be ignored.

## Brigade version

`brigade 0.24.0`

## Python version

`Python 3.12.3`

## OS

`Ubuntu 24.04.4 LTS`

## Pre-flight

- [x] I have scrubbed personal hostnames, tokens, session IDs, usernames, and private paths from this report.

solomonneas:
## What

Fixes #397. An upgraded install kept the obsolete standalone `brigade-work-loop.py` hook registered because `is_managed_handler` never recognized it, so install/update reconciliation preserved it as a foreign user hook. That stale hook read each untracked file fully into memory (`digest.update(path.read_bytes())`) and OOMed on a ~10 GB untracked cache blob.

## Changes

- **Legacy detection + removal** (`package.py`, `install_cmd.py`): `is_legacy_handler` recognizes standalone `brigade-work-loop.py` registrations, **anchored to executable position** (tokens[0] is the script, or a python interpreter followed by the script as first non-flag arg) so commands that merely *mention* the name (`grep -r brigade-work-loop.py .`, a trailing comment, a quoted path) are never matched and never deleted. Install/update reconciliation drops legacy handlers idempotently while preserving genuine foreign hooks and dropping emptied groups.
- **Status/doctor reporting** (`install_cmd.py`, `doctor.py`): `legacy_handler_count` + `legacy_events` in hook status (disjoint from `foreign_handler_count`); doctor emits a WARN that preserves the base state signal and appends the settings file + exact repair command `brigade work hooks install --target <path>`.
- **Attribution guard** (`runtime.py`): `wired_target_from_payload` prefers the session cwd and explicit `cd`/`--target`/`git -C` targets over incidental bare-path arguments, and only enrolls cwd when the payload actually supplies one, so a Bash command that mentions another wired repo no longer enrolls it.
- **Bounded fingerprinting**: confirmed by inspection — the managed path hashes via `git hash-object` (streamed, bounded) and the directory fallback is stat-only; no in-process whole-file reads. Peak RSS does not scale with file size. A regression test locks this in.
- **Docs** (`technical-guide.md`): names model caches, virtualenvs, and generated databases as snapshot-skipped directories.

## Tests

Regression tests (using `tmp_target`/`tmp_path`, never the real workspace): legacy anchoring incl. false-positive guards, legacy+managed coexistence in one group, multi-event removal, disjoint foreign/legacy counts, uninstall preserving foreign hooks, doctor normal-vs-legacy behavior, no-cwd attribution, and a mocked-large untracked file fingerprinted without reading its bytes.

## Verification

`./scripts/verify` green (ruff, format, mypy, version_sync, managed_snapshot, pytest coverage 82.42% ≥ 78 floor). Captured receipt: `20260721-153145-work-verify-ab0bf2`.

> Note for local reviewers on this machine: `TMPDIR` under a `.cache`-named path makes `tests/guard/` spuriously fail (the content-guard scanner excludes `.cache`); run verify with `TMPDIR=/tmp`. CI uses `/tmp` and is unaffected.

---

> [!NOTE]
> **Medium Risk**
> Changes Claude hook reconciliation and Bash repo attribution in the work-loop runtime; mistakes could drop wrong hooks or mis-attribute sessions, but behavior is heavily tested and legacy matching is narrowly anchored.
> 
> **Overview**
> **Fixes stale `brigade-work-loop.py` hooks** that survived install/update and could OOM on large untracked files. Hook install/update/uninstall now strips **legacy** standalone script registrations (executable-position matching only) alongside managed handlers, while keeping foreign hooks and dropping empty groups idempotently.
> 
> **Status and doctor** expose `legacy_handler_count` / `legacy_events` (separate from foreign handlers) and **WARN** with `brigade work hooks install --target <path>` when legacy entries remain.
> 
> **Runtime** tightens which repo a Bash tool call is attributed to: session `cwd` is preferred over incidental paths in the command, and `cwd` is only considered when the payload actually provides it—so mentioning another wired repo in `rg` no longer enrolls that repo.
> 
> Docs note that work-loop snapshot fingerprints skip common gitignored dirs (caches, `.venv`, `node_modules`, etc.). Tests cover legacy removal, doctor messaging, attribution edge cases, and large-file fingerprinting without in-process full reads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0bcb6d27652d1a726ca953de763a949c57e3161a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Claude hook installation, updates, and removal now clean up legacy standalone handlers while preserving unrelated handlers.
  * Hook status reports identify legacy handlers and affected events.
  * `brigade doctor` now warns when legacy handlers are detected and provides a repair command.
  * Improved command target resolution when an explicit working directory is provided.
* **Tests**
  * Added coverage for legacy handler detection, cleanup, status reporting, diagnostics, and command resolution.

### DIFF
--- a/docs/technical-guide.md
+++ b/docs/technical-guide.md
@@ -625,7 +625,7 @@ Start-of-day commands:
 - `brigade work brief` shows branch state, active sessions, pending tasks, import counts, latest dogfood run, and the command to continue.
 - `brigade work status` is the quick dashboard for branch state, dogfood readiness, paths, latest run, and extracted next step.
 - `brigade work doctor` checks dogfood config, security config, evidence bundles, Codex CLI, artifact paths, handoff inbox, task acceptance, issue-backed tasks, stale active sessions, ignore coverage, and latest run context.
-- `brigade work hooks install|update|status|uninstall` manages the project-scoped Claude Code work-loop package while preserving foreign settings and hooks. The runtime entrypoint is `brigade work hook-run`; it is called by Claude Code rather than by operators.
+- `brigade work hooks install|update|status|uninstall` manages the project-scoped Claude Code work-loop package while preserving foreign settings and hooks. The runtime entrypoint is `brigade work hook-run`; it is called by Claude Code rather than by operators. The work-loop snapshot fingerprint skips normally-gitignored directories such as model caches, virtualenvs (`.venv`), `__pycache__`, `node_modules`, and generated databases.
 - `brigade work resume` shows the active or latest session, latest dogfood run, extracted next step, and suggested command.
 - `brigade work inbox` groups pending scanner imports by source, kind, priority, age, and acceptance coverage, then suggests plan, promote, dismiss, or run commands.
 - `brigade work backup status` reads local backup health summaries and reports snapshot, check, prune, and restore rehearsal risk without running backup commands.
--- a/src/brigade/claude_hooks/install_cmd.py
+++ b/src/brigade/claude_hooks/install_cmd.py
@@ -9,7 +9,7 @@ from typing import Any
 
 from .. import localio
 from ..config import load_config
-from .package import MANAGED_EVENTS, PACKAGE_ID, PACKAGE_VERSION, is_managed_handler, managed_groups
+from .package import MANAGED_EVENTS, PACKAGE_ID, PACKAGE_VERSION, is_legacy_handler, is_managed_handler, managed_groups
 
 SETTINGS_REL_PATH = Path(".claude/settings.json")
 SIDECAR_REL_PATH = Path(".brigade/claude-hooks.json")
@@ -35,7 +35,12 @@ def _load_settings(path: Path) -> tuple[dict[str, Any] | None, str | None]:
     return payload, None
 
 
-def _without_managed(groups: object, event: str) -> list[object]:
+def _without_managed_or_legacy(groups: object, event: str) -> list[object]:
+    """Drop managed and legacy standalone handlers, preserving genuine foreign hooks.
+
+    Groups that become empty after dropping handlers are removed so install/update
+    reconciliation is idempotent and does not leave stale empty arrays behind.
+    """
     if not isinstance(groups, list):
         return []
     kept: list[object] = []
@@ -47,7 +52,9 @@ def _without_managed(groups: object, event: str) -> list[object]:
         if not isinstance(handlers, list):
             kept.append(group)
             continue
-        foreign = [handler for handler in handlers if not is_managed_handler(handler)]
+        foreign = [
+            handler for handler in handlers if not is_managed_handler(handler) and not is_legacy_handler(handler)
+        ]
         if foreign:
             updated = dict(group)
             updated["hooks"] = foreign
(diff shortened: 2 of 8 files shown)


solomonneas:
@coderabbitai review

solomonneas:
bugbot run

solomonneas:
@coderabbitai full review

PR #416, opened 2026-07-21
fix(work): name the remedy when verify rejects a shell interpreter

### LINKED ISSUE #412: verify run: high-risk-executable rejection gives no remedy (unlike the metacharacter branch)
## Summary

`brigade work verify run` correctly rejects shell interpreters (`bash`, `sh`, ...) as verification executables, but the rejection message is a dead end: it names the offending executable and gives no path forward. Its sibling check (shell metacharacters) already models the better behavior by naming the remedy. This is a message/DX fix, not a request to relax the policy.

## The policy is correct, keep it

`verify run` executes argv directly with `shell=False` (`src/brigade/work_cmd/verification.py:362`), and `SCANNER_HIGH_RISK_COMMANDS` (`src/brigade/work_cmd/constants.py:204`) blocks `bash/sh/zsh/fish/powershell/pwsh/ssh/scp/rsync`. This is documented intent (`docs/work-closeout.md:19`, `docs/scanner-registry.md:47`) and blocking shells is what keeps `bash -c "a; b | c"` from smuggling a shell past the adjacent metacharacter guard. No objection to the block itself.

## The gap

The two rejection branches in `_verify_parse_command` are inconsistent about guidance:

- Metacharacter branch (`verification.py:61-69`):
  `"high-risk verification command contains shell metacharacters (use --argv-json to pass a pre-parsed argv list instead)"` — actionable.
- High-risk executable branch (`verification.py:59-60`, and `_verify_parse_argv:91-92`):
  `"high-risk verification command: bash"` — no remedy.

A user or agent whose check is `bash ./scripts/check.sh` hits the second message and has nowhere to go. `--argv-json` is not the answer here, because `_verify_parse_argv` re-applies the same high-risk set (`verification.py:91`), so `["bash","-c",...]` is rejected there too (correctly). The actual supported pattern - make the script executable and invoke it by its path so a resolvable non-shell binary runs via its shebang - is discoverable only by reading the source.

## Repro

```
$ brigade work verify run --target . --command "bash ./check.sh" --capture brigade-work
- bash ./check.sh [rejected] exit=2
  stderr_summary: high-risk verification command: bash
```

Real-world friction: burned three attempts before reading `verification.py` to learn the supported shape.

## Proposed fix

Give the high-risk-executable rejection the same courtesy the metacharacter branch already has: name the supported pattern. Something like:

```
high-risk verification command: bash (verify runs argv directly with no shell;
invoke a resolvable executable instead, e.g. a chmod +x script by its path
like ./scripts/check.sh)
```

Optionally, add one line to `docs/work-closeout.md` stating the supported shape for multi-step or shell-shaped checks (a `chmod +x` script invoked by path). Message-only change; no behavior change to what is accepted or rejected.

## Secondary note (lower priority)

The only route for a shell-shaped check is to make a script executable and call it by path. On a fresh checkout, or a script tracked without the `+x` bit, that means mutating the working tree with `chmod +x` before verification will run. Not blocking, but if there is appetite, an explicit "run this file as a script" affordance would avoid the mode change. The message fix above is the piece worth doing regardless.

solomonneas:
Closes #412.

## What

`verify run` rejects shell interpreters (`bash`, `sh`, `zsh`, ...) as verify
executables, which is correct: verify runs argv directly with `shell=False`, so a
shell interpreter is never a valid executable. But the rejection message
dead-ended at `high-risk verification command: bash`, while the sibling
shell-metacharacter branch already names its remedy (`use --argv-json`). A caller
whose check was `bash ./script.sh` had no pointer to the supported pattern.

## Change

- Add a shared `_high_risk_command_message` helper, used by both the `--command`
  (`_verify_parse_command`) and `--argv-json` (`_verify_parse_argv`) rejection
  sites so the two stay in sync.
- The message names the supported shape (a resolvable executable, e.g. a
  `chmod +x` script invoked by its path) and deliberately does **not** point at
  `--argv-json`, which applies the same high-risk block and so is not a shell
  escape hatch.

Message-only: the set of accepted and rejected commands is unchanged.

## Tests

Adds `test_work_verify_run_rejects_shell_interpreter_with_remedy`, parametrized
over `--command` and `--argv-json`, asserting the rejection carries the remedy
text and does not mention `--argv-json`.

`./scripts/verify` green: ruff, ruff-format, mypy, version-sync, managed-snapshot,
and `3502 passed, 3 skipped`, coverage 82.42%.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved and standardized rejection messaging when high-risk shell interpreters are used as verification commands.
  * Ensured consistent, actionable guidance across both supported input formats for specifying the verify executable.

* **Tests**
  * Added CLI coverage to confirm `work verify run` rejects these high-risk commands in both `--command` and `--argv-json` modes, and that the error details include the recommended remedy without referencing the `--argv-json` flag.

### DIFF
--- a/src/brigade/work_cmd/verification.py
+++ b/src/brigade/work_cmd/verification.py
@@ -41,6 +41,22 @@ def _receipt_git_snapshot(target: Path) -> dict[str, Any] | None:
     return {"head": head, "branch": branch, "dirty_files": len(status.stdout.splitlines())}
 
 
+def _high_risk_command_message(executable: str) -> str:
+    """Rejection message for a shell/remote executable used as a verify command.
+
+    Mirrors the shell-metacharacter branch by naming the remedy: verify runs the
+    command directly with no shell, so a shell interpreter is never a valid
+    executable. ``--argv-json`` is deliberately not suggested here because that
+    path applies the same high-risk block, so the fix is a resolvable non-shell
+    executable, e.g. a chmod +x script invoked by its path.
+    """
+    return (
+        f"high-risk verification command: {executable} "
+        "(verify runs with no shell; use a resolvable executable, e.g. a "
+        "chmod +x script run by its path like ./scripts/check.sh)"
+    )
+
+
 def _verify_parse_command(command: str, target: Path) -> tuple[list[str] | None, dict[str, str], str | None]:
     try:
         parts = shlex.split(command)
@@ -57,7 +73,7 @@ def _verify_parse_command(command: str, target: Path) -> tuple[list[str] | None,
         return None, env, "command contains only environment assignments"
     executable = Path(argv[0]).name
     if executable in constants.SCANNER_HIGH_RISK_COMMANDS:
-        return None, env, f"high-risk verification command: {executable}"
+        return None, env, _high_risk_command_message(executable)
     if any(constants.SCANNER_SHELL_META_RE.search(part) for part in argv):
         return (
             None,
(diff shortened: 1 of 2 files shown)


solomonneas:
@coderabbitai review

Issue #403, opened 2026-07-21
Detect competing memory-care queue writers before migration


solomonneas:
## Outcome

Brigade should warn when more than one enabled memory-maintenance producer can overwrite the same `scan-latest.json` or `refresh-queue.json` files. The warning should include a read-only migration plan that selects one scanner while leaving card edits with the operator's existing policy-gated refresh worker.

## Why this is worth tracking

Brigade's current boundary is right: knowledge cards are canonical Markdown owned by the operator, and `brigade memory care scan` detects review work without editing those cards. The work inbox already deduplicates imported findings through stable keys and fingerprints.

One real OpenClaw deployment now has two enabled scanner jobs a few minutes apart:

- `brigade memory care scan`
- the older `scripts/card-decay-scanner.py`

Both can write `scan-latest.json` and `refresh-queue.json`. The later job wins, so the source of a queue depends on schedule order rather than an explicit ownership decision.

There is related path drift inside Brigade:

- `src/brigade/memory_cmd.py` writes new default state under `.brigade/memory-care/decay/` and reads the old `memory/cards/decay/` location as a compatibility fallback.
- `docs/memory-care.md` still presents `memory/cards/decay/` as the current write location.
- `src/brigade/doctor.py::_check_openclaw_cron_jobs` still expects the legacy `Card Decay Scanner (Daily)` job, so doctor treats the competing producer as healthy wiring.

The reliability case for idempotent downstream work is established. Temporal recommends idempotent activities because retries can execute the same operation again, and AWS gives the same guidance for retryable mutating operations. Brigade already follows that pattern for imports. This issue covers the remaining producer-ownership gap.

- Temporal: https://docs.temporal.io/activity-definition
- AWS Well-Architected: https://docs.aws.amazon.com/wellarchitected/latest/framework/rel_prevent_interaction_failure_idempotent.html

## Proposed slice

Add one read-only producer-ownership check to the existing memory-care/OpenClaw doctor path.

The check should:

- identify enabled Brigade and known legacy memory scanner producers;
- resolve their configured output locations without printing raw commands or private absolute paths;
- warn when two producers can write the same scan or queue artifact;
- distinguish scanners from downstream consumers such as `Card Decay Auto-Refresh (Safe)`;
- suggest a migration order: verify the Brigade output, point consumers at it, then disable the legacy scanner;
- leave cron jobs and knowledge cards unchanged.

[...]

PR #417, opened 2026-07-21
feat(doctor): detect competing memory-care queue writers (#403)

solomonneas:
## What
Implements #403. Warns when more than one enabled producer can write the same memory-care `scan-latest.json` / `refresh-queue.json`, with a read-only migration plan. Diagnostic only — never disables a cron job, edits a card, or mutates a queue file.

## Design (producers classified by what they write)
- **P1** Brigade memory-care **scan writer** (`brigade memory care scan` → `_output_dir`, default `.brigade/memory-care/decay/`).
- **P2** legacy OpenClaw cron **"Card Decay Scanner (Daily)"** → `memory/cards/decay/`.
- Reader scanners (`import-issues`, `import memory-refresh`) and the **"Card Decay Auto-Refresh (Safe)"** consumer are never counted as writers (the earlier suffix heuristic misclassified them and false-flagged the repo's own shipped `scanners.toml`).

Directory-level collision (`expanduser`+`resolve`, reusing `_scanner_output_path`) → exactly one WARN per shared dir. Host-global legacy cron producer gated on the target being a memory-care workspace.

## Tests
Collision → one warning; consumer scanner → none; shipped `scanners.toml` → none; custom colliding path → one; safe consumer → none; redaction asserts raw cron command/schedule/`jobs.json`/paths never leak.

## Verification
`./scripts/verify` green (receipt `20260721-173606-work-verify-c2a322`). Local reviewers: run with `TMPDIR=/tmp` (a `.cache`-path TMPDIR spuriously fails `tests/guard/`).

## Summary by CodeRabbit

* **New Features**
  * Added diagnostics to detect conflicting memory-care producers writing to the same artifacts.
  * Added support for custom memory-care output locations when checking for conflicts.
  * Collision warnings include migration guidance and confirm that checks are read-only.

* **Bug Fixes**
  * Improved recognition of memory-care scan commands.
  * Sensitive scheduling and command details are excluded from diagnostic warnings.

### DIFF
--- a/docs/memory-care.md
+++ b/docs/memory-care.md
@@ -63,9 +63,11 @@ Safe fix plans are copied into memory-care imports as metadata so the work inbox
 `brigade memory care scan` writes:
 
 ```text
-memory/cards/decay/scan-latest.json
-memory/cards/decay/refresh-queue.json
+.brigade/memory-care/decay/scan-latest.json
+.brigade/memory-care/decay/refresh-queue.json
 ```
+
+Before 0.9.1, scans wrote to `memory/cards/decay/`. Readers still fall back to that location when the default path has no `scan-latest.json` and the legacy path has existing output, so existing workspaces keep their queue continuity.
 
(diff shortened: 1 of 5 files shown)


solomonneas:
@coderabbitai review

PR #419, opened 2026-07-21
fix(engines): resolve managed engine binaries by absolute path everywhere

solomonneas:
## Problem

`brigade setup` installs the managed engine set (graphtrail, graphtrail-mcp, miseledger, sessionfind) to `~/.local/share/brigade/bin` and records it in `installed.json`, but every resolver and config generator still emitted bare PATH command names. On any machine with stale standalone binaries on PATH (cargo graphtrail, an old miseledger in `~/.local/bin`), everything kept running the standalone set and the managed engines sat dormant. This is the fleet-wide root cause behind the v0.25.0 consolidation not taking effect.

## Change

- New `component_bins` module, one shared resolver with a fixed order: env override (`GRAPHTRAIL_BIN`, `MISELEDGER_BIN`, ...) -> managed executable from `installed.json` -> PATH -> legacy standalone location (kept only as a migration-window fallback).
- All consumers route through it: the three duplicated per-module resolvers (`context_cmd`, `graphtrail_delta`, `evidence_brief`), receipts miseledger export, `search_cmd` status, station surface execution (`stations_cmd`), `ManagedTool.detect` plus the graphtrail/miseledger doctors, and the Cursor MCP generator (`cursor_user_cmd`), which now writes the managed absolute path into generated config.
- Stale `install_args` in `managed.py` (`cargo install graphtrail`, `go install .../miseledger`) become `brigade setup`.
- Broken env overrides no longer silently fall through to PATH; an explicit override that resolves nowhere is a miss.

**Deliberately unchanged:** `templates/stations/managed-snapshot.json` stays name-based. The snapshot is a committed, fleet-portable contract; name-to-absolute-path resolution happens at generation/execution time on each machine.

## Test isolation fix

The bare-host conftest fixture moved to the same seam. Before, it nulled `managed.proc.which` while other modules used `shutil.which`; with one resolver, that split is gone. The fixture now pins resolution to none while keeping the two sandboxed channels tests legitimately use (env overrides, test-created PATH dirs). Without this, the suite on a wired dev host resolved and **executed the real managed miseledger against the real evidence archive** during receipts tests.

## Verification

`./scripts/verify` green through `brigade work verify run` (captured receipt): ruff, format, version-sync, mypy, full pytest 3549 passed / 3 skipped, coverage 82.44% (floor 78%). GraphTrail impact run on `_graphtrail_bin`/`_miseledger_bin`/`_mcp_servers` callers before the change: aboyeur briefs, code_cmd, evidence_cmd, operator lifecycle, work verification graph delta, cursor install/doctor/adoption - all resolve-then-run or emit-config sites.

## Summary by CodeRabbit

* **New Features**
  * Added centralized executable discovery for managed tools, honoring per-binary env overrides, managed installs, system `PATH`, and legacy fallback locations.
  * GraphTrail/MiseLedger-related commands (including doctor/surface and MCP server entries) now use resolved executable paths.

* **Bug Fixes**
  * Improved detection/execution when binaries aren’t on `PATH`, and ensured health/doctor behavior aligns with resolved availability.

* **Tests**
  * Added coverage for resolution precedence, `~` expansion, and argv rewriting.
  * Updated existing command tests/mocks to use the new resolution behavior.

### DIFF
--- a/src/brigade/add.py
+++ b/src/brigade/add.py
@@ -148,7 +148,7 @@ def _run_manifest_add(ref: str, *, install_manifest: bool) -> int:
         return 0
     rc = 0
     for tool in manifest.tools:
-        installed = tool.kind == "executable" and managed.proc.which(tool.command) is not None
+        installed = tool.kind == "executable" and managed.component_bins.resolve(tool.command) is not None
         marker = "installed" if installed else "missing"
         if tool.kind == "skill-roster":
             marker = "skill-roster"
--- a/src/brigade/component_bins.py
+++ b/src/brigade/component_bins.py
@@ -62,23 +62,23 @@ def managed_path(name: str, *, env: Mapping[str, str] | None = None) -> str | No
 def resolve(name: str, *, env: Mapping[str, str] | None = None) -> str | None:
     """Resolve ``name`` to an executable path, or None when nowhere to be found."""
     environment = env if env is not None else os.environ
+    search_path = environment.get("PATH")
     override = environment.get(ENV_OVERRIDES.get(name, ""))
     if override:
         candidate = Path(override).expanduser()
         if _is_executable(candidate):
             return str(candidate)
-        found = shutil.which(override)
-        if found:
-            return found
-        return None
+        return shutil.which(override, path=search_path)
     managed = managed_path(name, env=env)
     if managed:
         return managed
-    found = shutil.which(name)
+    found = shutil.which(name, path=search_path)
     if found:
         return found
+    home = environment.get("HOME")
+    home_path = Path(home).expanduser() if home else Path.home()
     for relative in _LEGACY_RELATIVE.get(name, ()):
-        legacy = Path.home() / relative
+        legacy = home_path / relative
         if _is_executable(legacy):
             return str(legacy)
     return None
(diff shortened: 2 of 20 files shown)


solomonneas:
@coderabbitai All four findings applied in the follow-up commit: resolve() honors the supplied env's PATH and HOME, search sync plans and unhealthy-status guidance emit the resolved graphtrail path, the bare-host fixture searches only test-introduced PATH entries, and the doctor/context tests now assert the resolved absolute path reaches subprocess argv. ./scripts/verify green through brigade work verify run.

solomonneas:
@coderabbitai review

solomonneas:
@coderabbitai review

All three round-2 findings applied: tilde overrides expand against the supplied env's HOME (with a distinct-homes regression test), the Cursor MCP generator persists absolute paths, and the resolver tests now exercise the supplied env's PATH. The bare-host fixture additionally pins the component data root to a temp dir so the suite never reads the developer's real installed.json. ./scripts/verify green through brigade work verify run.

solomonneas:
Cross-model review (composer-2.5 lane) found four more real gaps, now fixed: manifest-based `brigade add` detection went through PATH-only `proc.which`; `brigade evidence crawl plan` still emitted bare `miseledger` argv heads; the legacy fallback missed `.exe` names on native Windows; and the bare-host test fixture did not clear engine override env vars exported by the developer shell. Two other flags were intentional behavior: a broken explicit override resolves to nothing rather than silently falling through (documented in the PR body), and the bare-name fallback in generated Cursor config is the documented pre-setup state. ./scripts/verify green through brigade work verify run.

solomonneas:
@coderabbitai review

solomonneas:
@coderabbitai resolve

PR #469, opened 2026-07-24
fix(harness): migrate legacy cursor install-state into profile sync

### LINKED ISSUE #468: Cursor: legacy harness install and user-profile sync deadlock on shared install-state.json
Found while landing #467 (issue #438 slice 2).

## Problem

`brigade harness install cursor --scope user` (legacy, `cursor_user_cmd.py`) and `brigade harness sync --target cursor --scope user` (profile layer, `harness_profile_cmd.py`) both use `~/.cursor/brigade/install-state.json` as their ownership state file, with incompatible schemas:

- legacy writes `{"version": 1, "files": ..., "hooks": ..., "mcp": ...}`
- the profile layer requires `schema_version: 2` with `workspace`/`harness` keys

Each side refuses the other's state as a conflict and neither offers a recovery path:

- After a legacy install, `harness sync --target cursor` fails with `unsupported ownership state version: None` (the legacy file has no `schema_version` key) and writes nothing.
- After a profile sync, legacy `harness install cursor` reports `unsupported ownership state version: 2` and skips all writes.

The only way out is manually deleting `~/.cursor/brigade/install-state.json`, which also destroys the ownership attestations uninstall depends on.

## Expected

One of:

1. The profile sync detects a version-1 legacy state and migrates/supersedes it deliberately (the legacy installer is documented as superseded by sync in --help), or
2. The two surfaces use distinct state filenames so they can coexist, or
3. At minimum, the conflict detail names the recovery step (`rm ~/.cursor/brigade/install-state.json` after `brigade harness uninstall` from the owning surface) instead of a bare version error.

## Repro

```bash
HOME=$(mktemp -d) brigade harness install cursor --scope user --write
HOME=<same> brigade harness sync --target cursor --scope user --workspace . --write
# -> conflict: unsupported ownership state version: None; nothing written
```

## Notes

- Safe today: both directions fail closed (conflict, no clobber), so this is a UX/migration bug, not data loss.
- `--adopt` does not cover ownership-state conflicts, only instruction/MCP surfaces.

solomonneas:
## Summary

Closes #468.

`brigade harness install cursor --scope user` (legacy, `cursor_user_cmd.py`, schema v1) and `brigade harness sync --target cursor --scope user` (profile layer, `harness_profile_cmd.py`, schema v2) share `~/.cursor/brigade/install-state.json` with incompatible schemas. Each refused the other's state as a conflict with no recovery path — a deadlock whose only manual escape (`rm` the state file) also destroys the ownership attestations uninstall needs.

Root-cause fix (option 1 from the issue): the legacy installer is already documented as superseded by sync, so the profile layer now **detects a legacy version-1 state and deliberately migrates it**:

- `harness_profile_cmd.load_profile_state` detects `{"version": 1, ...}` on the cursor harness and adopts every legacy attestation into the v2 ownership model — instruction digest, generated plugin/hook artifacts, the `hooks.json#sessionStart` entry fingerprint, skill files, and MCP projections (the legacy digest prefix is `stable_hash`-compatible, so unedited live entries still attest). Adoption is in-memory and read-only until a sync `--write` persists the migrated state; the result payload reports `"migration": {"from": "legacy-install-v1", ...}`.
- The sync then proceeds normally: legacy-owned artifacts reconcile as current, the superseded `brigade/mcp.json` catalog and legacy-only skill files are removed through the standard removed-profile/removed-registry paths, and nothing user-owned is clobbered.
- Reverse direction: legacy `harness install cursor` after a profile sync now fails closed with a conflict that names the recovery step — run `brigade harness uninstall --target cursor --scope user` from the owning surface first — instead of a bare `unsupported ownership state version: 2`.

No other harnesses' behavior changes (migration is gated on `harness == "cursor"` and the legacy v1 shape; non-cursor and malformed states keep the existing fail-closed error).

## Tests

- `test_cursor_legacy_install_migrates_into_profile_sync`: legacy install → profile sync migrates cleanly (rule survives byte-for-byte, attestations carried), a second sync is a no-op, and `harness uninstall --scope user` removes exactly the artifacts — originally-legacy and newly-synced alike.
- `test_cursor_profile_sync_then_legacy_install_names_recovery`: profile sync → legacy install gives the recovery-named conflict, writes nothing, and leaves the v2 state untouched.
- `test_load_state_migrates_legacy_cursor_install_state_read_only` / `test_load_state_rejects_non_dict_legacy_cursor_sections`: unit coverage for the adoption mapping, read-only guarantee, and fail-closed edges.

All tests drive a temp HOME/`tmp_path`; the real `~/.cursor` is never touched.

## Verification (brigade-wired)

```
brigade work verify run --target . --command "pytest -q tests/test_harness_user_scope_slice2.py tests/test_harness_profile_cmd.py" --capture brigade-work
# status: completed — exit=0 (55 passed)

# status: completed — exit=0 (ruff, ruff format, mypy, version sync; 4186 passed, 3 skipped)
```

Receipts: `.brigade/work/verify-runs/20260724-002316-work-verify-322e6a/` and `.brigade/work/verify-runs/20260724-002456-work-verify-019210/`.

### DIFF
--- a/src/brigade/cursor_user_cmd.py
+++ b/src/brigade/cursor_user_cmd.py
@@ -7,7 +7,7 @@ import json
 from pathlib import Path
 from typing import Any
 
-from . import __version__, component_bins, localio
+from . import __version__, component_bins, harness_profiles, localio
 
 STATE_VERSION = 1
 MANAGED_MCP_NAMES = ("brigade", "graphtrail", "miseledger")
@@ -138,7 +138,14 @@ def _load_state(root: Path) -> dict[str, Any]:
         empty["_read_error"] = error or "ownership state is unreadable"
         return empty
     if payload.get("version") != STATE_VERSION:
-        empty["_read_error"] = f"unsupported ownership state version: {payload.get('version')!r}"
+        if payload.get("schema_version") == harness_profiles.PROFILE_STATE_VERSION:
+            empty["_read_error"] = (
+                "ownership state is managed by `brigade harness sync` (profile schema "
+                f"v{harness_profiles.PROFILE_STATE_VERSION}); recover by running `brigade harness "
+                "uninstall --target cursor --scope user` from the owning surface first"
+            )
+        else:
+            empty["_read_error"] = f"unsupported ownership state version: {payload.get('version')!r}"
         return empty
     for key in ("files", "hooks", "mcp"):
         if not isinstance(payload.get(key), dict):
--- a/src/brigade/harness_profile_cmd.py
+++ b/src/brigade/harness_profile_cmd.py
@@ -35,6 +35,7 @@ class SurfacePlan:
 class LoadedProfileState:
     state: dict[str, Any]
     error: str | None
+    migration: dict[str, Any] | None = None
 
 
 def digest_text(text: str) -> str:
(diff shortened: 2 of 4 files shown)


solomonneas:
Superseded by #510, which is the later green fix for the same #468 Cursor install-state migration. Closing this duplicate without merging.

PR #535, opened 2026-07-26
feat(agents): add the oracle browser researcher lane

solomonneas:
## What

Adds `oracle` as a Brigade CLI adapter pinned to its browser engine, so the roster's `researcher` role can run on a subscription cookie lane with no API key.

[steipete/oracle](https://github.com/steipete/oracle) (MIT) drives a real `gemini.google.com` session using Chrome cookies. Agent Pantry already syncs that cookie jar between machines and warns before it expires, so the auth substrate already exists. This PR is the adapter that connects the two.

| Layer | Component | Change |
|---|---|---|
| Auth | Agent Pantry cookie sync | none |
| Driver | `oracle --engine browser` | none, external binary |
| Adapter | `src/brigade/agents.py` | new |
| Timeout | `src/brigade/research/llm.py` | one floor |

Backend selection needed no change: `research/llm.py` already returns a `CliBackend` for any researcher declaring `cli`, and `engine.py` only calls `.complete()`.

Brigade does not install oracle, bundle Node, or touch a browser. It shells out and reads text back, the same as every other adapter.

## Why the timeout change rides along

`research/engine.py:138` asks for `timeout=30` on its planning call and `research/types.py:50` defaults to 60, both hardcoded at the call site. A browser round trip will not meet 30 seconds, so the lane fails on its first request without a floor. `Caps` has no timeout field, and `Caps.build` filters overrides with `hasattr`, so a config-only attempt would look applied and silently do nothing.

This is not a timeout widened to hide a symptom. It adds the missing mechanism for a backend to declare how slow it legitimately is, reusing the roster's existing per-agent `timeout_seconds` that `resolve_backend` was discarding. It only raises, never lowers, and seats that declare no `timeout_seconds` keep today's behavior exactly.

Happy to split this into its own PR if you would rather review it separately.

## Where to look hardest

**`failure_kind` is scored, so it has to be right.** The first version of the cookie-expiry hint shared the `provider-preflight` branch and inherited `failure_kind="workspace-trust"`, so every stale-cookie failure was being scored as a workspace trust refusal by `outcome capture` and `brigade model scorecard`. Now `"browser-auth"`, fixed at both failure paths in `run_agent`, with a codex regression guard proving the original kind still fires. Unit tests on the detail function alone had passed. Only a test through `run_agent` caught it.

**Read-only is `hard` by construction.** Oracle has no filesystem write path, so it is the only adapter reporting `hard` without a sandbox or a prompt instruction. Registering it also makes `cli = "oracle"` assignable to any seat rather than only the researcher. That is why the enforcement contract is tested rather than assumed.

## How to verify

```bash
./scripts/verify
```

4220 passed, 3 skipped, coverage 82.68% against the 78% floor.

Oracle-specific tests:

```bash
pytest tests/test_agents_oracle.py tests/test_research_llm.py tests/test_research_engine.py -q
```

| Claim | Proof |
|---|---|
| argv shape, model pin, never `--heartbeat`, never an API path | unit assertions |
| argv survives a real `exec` | stub binary on a narrowed PATH records what it received |
| stdout is the answer channel | stub returns markdown, `run_agent` returns it verbatim |
| auth failure on both `run_agent` paths | `failure_kind="browser-auth"` |
| auth beats workspace-trust when both patterns appear | stub emits both |
| non-oracle seats keep `workspace-trust` | codex regression guard |
| timeout floor survives the real engine | `DeepResearcher` asserts 300, not the engine's 30 |
| roster accepts `cli = "oracle"` | clears `is_known` and `limits.allow_models` |

## What is not proven

`oracle` is not installed on the machine this was written on, so no real browser round trip ran. Two claims rest on reading oracle's source rather than executing it:

- **stdout carries only the answer.** Based on `src/cli/renderOutput.ts` returning unrendered markdown when `richTty` is false, which it is under captured pipes, plus never passing `--heartbeat`. If a real run shows progress chrome, the fix is an extraction function following the `_parse_grok_final_output` precedent, not looser `validate_final_output`.
- **the browser session completes against synced cookies.**

The full proof and blocker record is in `docs/phase-oracle-browser-researcher.md`.

## Risk

Oracle's browser mode is labelled experimental by its author, and cookie automation of a consumer web session is grey against provider terms. The docs scope it to a single operator machine and explicitly not a fleet default. The `cli = "oracle"` seam keeps the driver replaceable without touching `research/`.

@coderabbitai review

## Summary by CodeRabbit

* **New Features**
  * Added support for the Oracle browser-based CLI adapter.
  * Oracle runs are enforced as read-only and support per-run model selection.
  * Added clearer guidance when Oracle browser sessions or cookies require reauthentication.
* **Improvements**
  * Research calls now honor the configured minimum timeout for the researcher.
  * Provider authentication and workspace-trust failures are reported separately.
* **Tests**
  * Added coverage for Oracle execution, authentication errors, model selection, read-only behavior, roster validation, and research timeouts.

### DIFF
--- a/docs/phase-oracle-browser-researcher.md
+++ b/docs/phase-oracle-browser-researcher.md
@@ -141,10 +141,34 @@ Models the browser engine accepts: `gemini-3.5-flash`, `gemini-3.1-pro`, and
 - [x] Test the `min_timeout` floor raises a short engine timeout, never lowers
       a generous one, and leaves seats without `timeout_seconds` unchanged.
 - [x] Test the expired-cookie preflight detail string.
-- [ ] Run focused tests and `./scripts/verify` through
+- [x] Run focused tests and `./scripts/verify` through
       `brigade work verify run`.
-- [ ] Live smoke: one `brigade research run` against real synced cookies,
+- [x] Live smoke: one `brigade research run` against real synced cookies,
       recording the result or the environmental blocker.
+
+## Live result
+
+The adapter dispatches correctly and fails cleanly when the binary is absent.
+`agents.run_agent("oracle", "hello")` returned `ok=False`,
+`failure_phase="dispatch"`, `failure_kind="command-not-found"`, detail
+`oracle not installed`, and `build_argv` produced exactly
+`['oracle', '--model', 'gemini-3.1-pro', '--engine', 'browser', '-p', 'hello']`.
+
+A full browser round trip could not run, blocked twice on this machine:
+
+- `oracle` is not installed (`command not found`, and no global npm package).
+  Brigade does not install it by design.
+- `brigade pantry status` reports the agentpantry build **rejected by version
+  policy** (unreleased or non-semver build; expected released >= 0.5.0), so the
+  cookie substrate is not healthy here either.
+
+Consequently the settled stdout decision is **not yet empirically confirmed**.
+It rests on reading oracle's `src/cli/renderOutput.ts` (`if (!richTty) return
+markdown;`, `richTty` defaulting to `process.stdout.isTTY`) plus never passing
+`--heartbeat`. The first real run should check whether stdout carried only the
+answer; if it did not, add an extraction function following the
+`_parse_grok_final_output` precedent rather than loosening
+`validate_final_output`.
 
 ## Resolved during planning
 
--- a/docs/phase-oracle-browser-researcher.md
+++ b/docs/phase-oracle-browser-researcher.md
@@ -106,10 +106,13 @@ Models the browser engine accepts: `gemini-3.5-flash`, `gemini-3.1-pro`, and
 
 - **oracle absent.** `resolve_agent_executable` already returns
   `failure_kind="command-not-found"` at `failure_phase="dispatch"`. No work.
-- **cookies expired.** This must not surface as a generic nonzero exit. Add an
-  `oracle` case to `_provider_preflight_detail` (`agents.py:493`) that
-  recognises oracle's login prompt and points the operator at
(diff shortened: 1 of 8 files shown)


solomonneas:
@coderabbitai review

solomonneas:
Retargeted rather than renamed, in \`a38f7c7\`.

The finding is correct: \`VALID\` gives \`agents.coder\` the role \`"write code"\`, so the test never built a researcher seat despite its name.

Renaming it to \`..._coder\` would have made the name honest but left the researcher path uncovered at the roster layer. Instead the fixture now also swaps the role, and the test asserts through \`find_role("researcher")\` — which is the exact lookup \`research/llm.py:resolve_backend\` performs — so the seat this PR actually wires is the one under test:

```python
text = (
    VALID.replace('cli = "ollama:llama3.3"', 'cli = "oracle"')
    .replace('role = "write code"', 'role = "researcher"')
    .replace('allow_models = ["codex", "ollama:*"]', 'allow_models = ["codex", "oracle"]')
)
loaded = roster_mod.load_roster(_write(tmp_path, text))
assert loaded.agents["coder"].cli == "oracle"
assert loaded.find_role("researcher").cli == "oracle"
```

\`"write code"\` occurs exactly once in \`VALID\`, so the replace is unambiguous. \`./scripts/verify\` green: 4220 passed, 3 skipped.

Note for anyone reading the diff: \`docs/**\` is path-excluded from this review, so \`docs/phase-oracle-browser-researcher.md\` was not looked at. It carries the design rationale, the recorded blockers, and the proof/unproven split.

Issue #503, opened 2026-07-24
outcome ledger: dual-criterion promotion (effectiveness AND utility retention)


solomonneas:
## What
A skill may only promote when it passes both an effectiveness criterion and a utility/cost-retention criterion.

## Why (evidence)
"A Critical Evaluation of Defenses against Prompt Injection" (ACM, 2026) finds published defenses report inflated success because they are evaluated only against static attacks and without measuring utility loss; a defense that rejects everything "passes". The same failure shape applies to any gate or skill in the ledger: promotion on a single success metric rewards degenerate strategies. Small, principled change to the promotion rule in `outcome rank`.

## Effort
S-M

Issue #502, opened 2026-07-24
outcome ledger: multi-dimension skill scorecards from verifier receipts only, with exploration caps


solomonneas:
## What
Score skills/cards on separate dimensions (verified reliability, competence, receipt completeness) populated exclusively from verifier-authored receipts, never self-reports. Add exploration caps so unproven skills cannot gain broad write authority, and record score inputs plus policy version in routing receipts.

## Why (evidence)
The MAMA framework (Chalkiadakis, Liu, Polyzos, FLLM 2025, doi:10.1109/fllm67465.2025.11391149) reports a 10.6% MRR gain from expertise-and-trust routing over a hash-linked verifiable ledger, with the authors stressing that self-reported scores are the failure mode. Locally, outcome ledgers show the symptom this addresses: hundreds of records, single-digit scored, `promoted=0` on nearly every wired repo. Scalar promotion is not converging; dimensional scoring gives the ranker something to rank.

## Effort
L

solomonneas:
Planning pass is in draft PR #570: `docs/proposals/skill-scorecards.md`.

Ordered executor slices:

1. #571 (M) - verifier-owned subject, work-owner, patch/fixture, check-role, and failure-taxonomy binding.
2. #572 (M) - receipt-only effectiveness, verifier-cost, retry-stability, and evidence-integrity projection.
3. #573 (M) - bounded route exploration and score inputs on route receipts.
4. #503 (S-M, existing) - dual effectiveness and utility-retention promotion, folded into this design rather than duplicated.
5. #574 (S) - audit-only backfill plus work-brief and operator scorecard surfaces.

Important dependencies/composition points: #474 supplies typed infrastructure failures, #507 supplies generated-patch quarantine/independence, #499 supplies declared obligation ids, #546 prevents no-work ownership pollution, #557 supplies adversarial fixtures, #560 preserves failed-attempt ordering, #562 defines additive receipt evolution, and #563 supplies the patch identity tuple.

#502 remains open as the parent tracking issue.

PR #570, opened 2026-07-26
docs: propose verifier-bound skill scorecards

solomonneas:
## Summary

- Defines multi-dimension skill scorecards whose eligibility and values come only from verifier receipt v2 fields.
- Adds a sharp subject, patch, work-owner, verifier, check-role, and failure-taxonomy trust boundary.
- Specifies bounded exploration, dual effectiveness and utility promotion, immediate routing demotion, strict cold start, and audit-only backfill.
- Decomposes the implementation into four new S/M issues plus existing #503.

## Current failure mode

`scored_artifact_count` is only the number of distinct artifact ids with parsed outcome rows. Capture is opt-in, rank is read-only, and only explicit `brigade outcome reconcile --apply` advances promotion state. The proposal does not treat the reported 1,601 existing rows as trustworthy training data.

## Verification

- `brigade work verify run --target . --command "git diff --cached --check" --command "brigade handoff lint --target . .claude/memory-handoffs/2026-07-26-skill-scorecards-planning.md" --capture brigade-work`
- Receipt: `20260726-223449-work-verify-3713cb`
- MiseLedger evidence: `68adffeab4dd404c3b95b47d`
- Both commands exited 0.

Refs #502 and #503.

### DIFF
--- a/docs/proposals/skill-scorecards.md
+++ b/docs/proposals/skill-scorecards.md
@@ -1,0 +1,529 @@
+# Proposal: multi-dimension skill scorecards from verifier receipts (#502)
+
+Status: draft, planning pass only. Not implemented.
+
+Parent issues: [#502](https://github.com/escoffier-labs/brigade/issues/502) (scorecards),
+[#503](https://github.com/escoffier-labs/brigade/issues/503) (dual-criterion promotion).
+
+Schema contracts: verify receipt `brigade.work_verify_receipt` `schema_version: 2` after merged
+[#563](https://github.com/escoffier-labs/brigade/pull/563); additive `schema_version` evolution from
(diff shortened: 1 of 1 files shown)


solomonneas:
Blocking readiness on one stale dependency reference:

- The opening schema-contract section says #562 is pending and points implementers to `origin/pr-562`. #562 merged as `ad5541f532abfc53ce480b5bcf2e23b5cca3b4de`, and that remote ref does not exist.
- Update the reference to `docs/receipt-schemas.md` on `main`, state that #562 is merged, and rebase so CI validates the proposal against the merged contract.

I found no other issue in the scorecard decomposition or its four filed implementation slices.

Issue #474, opened 2026-07-24
run: unify seat preflight into a general health probe with a typed worker-failure taxonomy


solomonneas:
## What
1. Generalize the piecemeal seat preflights (#394 Cloudflare env vars, #284 Cursor ACP auth, #299 model ID validation) into one per-seat health probe that runs before dispatch and marks degraded seats.
2. Classify worker failures into a typed taxonomy recorded in run receipts, so `friction scan` and `outcome rank` can aggregate by cause.

## Why (evidence)
A 30-day friction scan (2026-07-24) shows at least 10 distinct seat-failure modes that all surface as opaque worker failures in `.brigade/runs/*/worker-results.json`: model at capacity, empty output, cursor exit-0-without-output (trust), effort value `'max'` rejected by the Anthropic API, wrong CLI flag (`--print` vs `--prompt` on kimi), Fable data-policy acknowledgment required, missing env vars, provider 403 subscription, acpx exit 5, and a permission-prompt hang. None of these are attributable today without reading raw snippets.

## Effort
M-L

solomonneas:
Planning pass is in PR #575: `docs/proposals/2026-07-26-seat-health-probe-and-worker-failure-taxonomy.md`.

Implementation decomposition:

1. #576, M, add the closed worker-failure taxonomy and receipt compatibility
2. #577, M, implement the adapter-owned unified seat health probe, blocked by #576
3. #578, M, route from seat health and persist retry decisions, blocked by #576 and #577
4. #579, M, detect and stop worker isolation breaches, blocked by #576 and #577 and parallel with #578
5. #580, S, keep worker infrastructure failures neutral, blocked by #576 with final fixtures after #578 and #579

#474 remains the tracking issue. Proposed order: #576, then #577, then #578 and #579 in parallel, then #580.

PR #575, opened 2026-07-26
docs(run): propose seat health failure taxonomy

solomonneas:
Tracks #474.

## Summary

- inventory current roster, doctor, transport, dispatch, and receipt behavior
- define a closed worker-failure taxonomy and adapter-owned seat health probe
- specify additive receipt shapes, routing rules, migration, and compatibility
- separate the implementation into five ordered S/M tracker items

This is a planning-only change. It does not implement the probe or alter run behavior.

## Verification

- `brigade work verify run --target . --command "./scripts/verify" --capture brigade-work`
- completed: `20260726-223358-work-verify-e5e891`

### DIFF
--- a/docs/proposals/2026-07-26-seat-health-probe-and-worker-failure-taxonomy.md
+++ b/docs/proposals/2026-07-26-seat-health-probe-and-worker-failure-taxonomy.md
@@ -1,0 +1,634 @@
+# Proposal: seat health probes and typed worker failures
+
+Status: proposed for issue #474. Planning only, no implementation.
+
+## Decision summary
+
+Replace the current collection of roster, doctor, transport, and dispatch
+checks with one adapter-owned health probe for every requested seat. Run the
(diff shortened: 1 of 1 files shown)


solomonneas:
Blocking merge on a receipt-contract mismatch with merged #562:

- The `brigade.seat_health.v1` example omits integer `schema_version`.
- The nested `brigade.worker_failure.v1` example also omits integer `schema_version`.
- The proposal says `seat-health.json` is updated in place after re-probes, but does not classify it as a latest-state snapshot or preserve probe events. That conflicts with #562s write-once/append-only policy unless the mutable snapshot semantics are explicit.
- The new writers do not specify canonical sorted-key serialization, absent-versus-null rules, or reader compatibility.

Please rebase onto `main` and define those contract points before implementation is split from this proposal. An implementer following the current examples would produce receipts that fail the policy merged in #562.

Issue #568, opened 2026-07-26
runs: add an append-only lifecycle journal and replayable run.json projection


solomonneas:
## Problem

Brigade rewrites `run.json` as a run moves through planning, dispatch, synthesis, recovery, and terminal states. Atomic replacement keeps polling readers from seeing a truncated file, but each write erases the prior state.

After a process dies, Brigade cannot reliably answer:

- Which lifecycle transition committed last?
- Did a dispatch start, finish, or stop between snapshot writes?
- Is a recovery request a retry of the same transition or a new request?
- Can a missing, corrupt, or stale `run.json` be rebuilt from durable facts?

Receipts remain useful proof artifacts, and the outcome ledger remains the scored-outcome chain. Neither provides ordered lifecycle history for a run.

Brigade already defines `brigade.run_event.v1`. This issue gives that event type a storage, replay, and migration contract.

## Inspiration

The immediate prompt for this issue was Yohei Nakajima's observation that long-running agent systems tend toward immutable event logs:

- [“if you are trying to solve long-running agent problems, you will eventually start using immutable event logs”](https://x.com/i/status/2081411733235863605)
- [A follow-up connecting the pattern to Git history, transaction journals, medical and aviation audit trails, critical-infrastructure event logs, and versioned scientific records](https://x.com/i/status/2081441852709093670)

These posts are design prompts, not dependencies or an argument for distributed log infrastructure. Brigade applies the principle at the per-run, local-file boundary defined below.

## Concurrency and storage boundary

The current run guard permits one lifecycle writer for a run. `watch`, `show`, and similar commands are concurrent readers. Recovery may become the next writer only after it claims a stale run lock.

Use that constraint instead of adding a database:

- Append lifecycle records to `<run-dir>/events/lifecycle.jsonl`.
- Keep provider and worker event streams separate.
- Require the active run-lock owner for normal lifecycle appends.
- Use one write call per canonical JSON line, flush it, and call `fsync` before refreshing `run.json`.
- Treat a partial final line as an interrupted append. Recovery may truncate only that incomplete suffix after preserving diagnostic evidence.
- Store the journal with the same private permissions and gitignore behavior as other local run artifacts.

This is a per-run journal, not a repo-wide broker. If Brigade later needs to atomically append an event and create work for multiple competing consumers, that should be a separate SQLite-backed queue issue with evidence that the single-writer boundary is no longer enough.

## Event contract

Each line uses an allowlisted event-type schema and contains:

[...]

PR #615, opened 2026-07-30
feat: add crash-safe run recovery checkpoints

solomonneas:
Part of #568.

## Summary

- Publish content-addressed checkpoints before lifecycle status events and the atomic `run.json` replacement.
- Retain the run lock when lifecycle activation publishes recovery state but the initial receipt write fails.
- Require bounded N/N+1 checkpoint coverage before stale-run recovery, with exact claimed-lock restoration on every validation failure.
- Recover a matching persistent stale claim without mutating a foreign or invalid current lock.
- Share stale-lock receipt construction between runtime recovery and Doctor reconstruction.
- Add a read-only Doctor check for repairable and unsafe recovery states across the 50 newest run directories.

## Why

The lifecycle journal proves event ordering, but it cannot restore the canonical run record after an interrupted write by itself. Activated journals now fail closed unless checkpoint validation proves a safe recovery path. Legacy runs without an activated journal keep their existing behavior.

## Verification

- `./scripts/verify`
- Ruff lint and format passed.
- Mypy passed across 333 source files.
- 5,126 tests passed, 3 skipped, with 83.03% coverage.
- Brigade receipt: `20260730-050126-work-verify-2ab7e1`
- Independent Brigade review: `20260730-045649-2637de6b`, ready with no actionable findings.

## Review focus

- Checkpoint and status tail coverage at sequences N and N+1.
- Lock retention after partial initial lifecycle publication.
- Persistent stale-claim recovery while preserving a different current lock.
- Shared stale-lock receipt construction and provenance parity.
- Exact stale-lock restoration when validation or terminalization fails.
- Doctor reconstruction of terminal recovery receipts without filesystem mutation.

### DIFF
--- a/docs/phase-568-slice-5-recovery-checkpoints-plan.md
+++ b/docs/phase-568-slice-5-recovery-checkpoints-plan.md
@@ -1,0 +1,368 @@
+# Issue #568 Slice 5: Recovery Checkpoints Implementation Plan
+
+Implementation plan for the slice 5 boundary of GitHub issue #568
+(https://github.com/escoffier-labs/brigade/issues/568), paired with the design spec
+`docs/phase-568-slice-5-recovery-checkpoints.md`. Prior slice docs:
+`docs/phase-568-slice-3-run-projector.md` and `docs/phase-568-slice-4-shadow-comparison.md`.
+
+Every task is RED first: write the named failing tests that pin the behavior, implement the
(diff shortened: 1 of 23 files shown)


solomonneas:
@coderabbitai review

solomonneas:
@coderabbitai review

PR #633, opened 2026-07-31
fix(journal): close run-journal hardening gaps

### LINKED ISSUE #632: run-journal: four hardening residuals from the #568 grading review
Residual hardening items from the grader's adversarial review of the #568 stack (slices 7-11, PRs #624-#627 and #630). None is a correctness or integrity break; all survived attack attempts only as hardening gaps. Ordered by priority.

1. Mid-run enrollment shedding on unreadable run.json. The `_payload` helper in `aboyeur.run()` (aboyeur.py:3057-3076) swallows `(OSError, JSONDecodeError)` and omits both `lifecycle_journal_requested` and `run_journal_authority_requested` from later status writes, so `_payload_requests_authority` resolves legacy and the snapshot silently sheds enrollment. `record_run_start` already fails closed on the same condition; this path should too. This is the last sliver of the slice 7 D3 fix.

2. Redaction records are mutable side files, not chained journal events. A second overlapping redaction rewrites the first operation's `redaction:<op>:<seq>` idempotency markers, leaving op1 attributable only through `record.json` lineage files. The lineage validator fails closed on tampering it can see but cannot detect wholesale replacement of the record set together with a consistent journal. Consider appending a chained redaction event or anchoring the lineage tip digest in the journal.

3. `localio.write_text_atomic` (localio.py:45-64) performs no parent-directory fsync after `os.replace`, so the run.json rename can be lost on power failure. Recovery via checkpoint plus journal makes this an ordering rather than durability gap, but the journal and checkpoint paths already fsync their parent directories; this is the one inconsistency left.

4. `recover_partial_tail` (run_journal.py:915-980) truncates the journal outside `_APPEND_LOCK` and SIGTERM deferral. It is only reachable from the lock-gated recovery path today, but nothing in the function enforces that; take the lock and defer signals inside the function.

Refs #568.

solomonneas:
Fixes #632.

This follow-up closes four hardening gaps found while grading #568. It keeps existing event canonicalization, envelope fields, and projector output unchanged for existing event types. No existing golden fixture was regenerated.

## Fix and regression map

| Item | Fix | Named regression tests |
| --- | --- | --- |
| Mid-run enrollment shedding | `record_run_start` now reports lifecycle or authority enrollment. Later status writes inspect `run.json` without following symlinks and raise `refusing to overwrite unknown durable enrollment state` for missing, unreadable, non-object, or non-regular receipts. | `test_record_run_start_treats_lifecycle_only_enrollment_as_durable`<br>`test_run_payload_fails_closed_when_enrolled_run_json_becomes_corrupt` |
| Chained redaction record | Adds status-neutral `run.redaction.recorded` events with the five reference-only fields from #632. Anchors hash the exact `record.json` bytes, survive later overlapping rewrites, bind child records to the parent anchor observed in the child's quarantined input journal, and are checked against the record inventory after rewrite and cleanup. Structural-only ranges now fail before mutation. | `test_overlapping_redactions_preserve_both_chained_operation_anchors`<br>`test_replaced_redaction_record_set_fails_chained_anchor_validation`<br>`test_redaction_after_cleaned_parent_uses_active_anchor_reference`<br>`test_redaction_inventory_rejects_tampered_parent_record_anchor_reference`<br>`test_redaction_rejects_preserved_structural_only_range_without_mutation` |
| Atomic text-write directory durability | `localio.write_text_atomic` fsyncs the parent directory after `os.replace` on supported platforms, opens it with directory and no-follow flags, and preserves the primary fsync error if close also fails. | `test_write_text_atomic_fsyncs_parent_after_replace`<br>`test_write_text_atomic_opens_parent_without_following_symlinks`<br>`test_write_text_atomic_reports_directory_fsync_failure_after_replace`<br>`test_write_text_atomic_skips_parent_fsync_when_platform_has_no_support` |
| Partial-tail recovery serialization | `recover_partial_tail` now owns the same append lock and SIGTERM deferral critical section as `append_event`. | `test_recover_partial_tail_reentrancy_matches_append_event` |

## Verification

```text
receipt: 20260731-164708-work-verify-193280
5554 passed, 3 skipped in 542.02s
Total coverage: 82.99%
```

The follow-up independent review found no remaining Critical or Important findings. Its two Minor findings, structural-only no-op redactions and duplicate resume verification, were fixed before the full gate above.

### DIFF
--- a/src/brigade/aboyeur.py
+++ b/src/brigade/aboyeur.py
@@ -2791,7 +2791,7 @@ def record_run_start(
     codex_transport: str | None = None,
     started_at: datetime | None = None,
     scheduler: str | None = None,
-) -> None:
+) -> bool:
     """Write the minimal typed receipt needed before optional or blocking work.
 
     The requested scheduler is recorded here so a run that dies before dispatch
@@ -2870,6 +2870,7 @@ def record_run_start(
     except (OSError, run_lifecycle.LifecycleJournalError, run_checkpoint.CheckpointError) as exc:
         raise runguard.RetainRunLockError(f"failed to write initial run receipt: {exc}") from exc
     _write_json(output_dir / "roster.json", _roster_payload(roster))
+    return lifecycle_requested or authority_requested
 
 
 @contextmanager
@@ -3037,6 +3038,7 @@ def run(
     output_dir = output_dir.expanduser() if output_dir is not None else None
     handoff_inbox = handoff_inbox.expanduser() if handoff_inbox is not None else None
     direct_worker = worker is not None
+    durable_enrollment_expected = False
 
     # What the roster/flag asked for vs what dispatch actually ran. `used` stays
     # None until dispatch resolves it, so a run that dies before dispatch reads
@@ -3058,11 +3060,30 @@ def run(
             "lifecycle_journal_requested" not in kwargs or "run_journal_authority_requested" not in kwargs
         ):
             run_path = output_dir / "run.json"
-            if run_path.is_file():
+            try:
+                run_info = os.lstat(run_path)
+            except FileNotFoundError as exc:
+                if durable_enrollment_expected:
+                    raise runguard.RetainRunLockError("refusing to overwrite unknown durable enrollment state") from exc
+                run_info = None
+            except OSError as exc:
+                if durable_enrollment_expected:
+                    raise runguard.RetainRunLockError("refusing to overwrite unknown durable enrollment state") from exc
+                run_info = None
+            if run_info is not None and (not os.path.isfile(run_path) or os.path.islink(run_path)):
+                if durable_enrollment_expected:
+                    raise runguard.RetainRunLockError("refusing to overwrite unknown durable enrollment state")
+            elif run_info is not None:
                 try:
                     existing = json.loads(run_path.read_text())
-                except (OSError, json.JSONDecodeError):
+                except (OSError, ValueError, RecursionError) as exc:
+                    if durable_enrollment_expected:
+                        raise runguard.RetainRunLockError(
+                            "refusing to overwrite unknown durable enrollment state"
+                        ) from exc
                     existing = None
+                if not isinstance(existing, dict) and durable_enrollment_expected:
+                    raise runguard.RetainRunLockError("refusing to overwrite unknown durable enrollment state")
                 if isinstance(existing, dict):
                     if (
                         "lifecycle_journal_requested" not in kwargs
@@ -3139,7 +3160,7 @@ def run(
             print(f"error: {worker_error}", file=sys.stderr)
             return 2
     if output_dir is not None:
-        record_run_start(
+        durable_enrollment_expected = record_run_start(
             output_dir,
             task=task,
             cwd=cwd,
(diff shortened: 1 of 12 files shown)


solomonneas:
Grader review: APPROVE-WITH-CHANGES. All four #632 items are correctly implemented and the closeout map checks out (receipt `20260731-164708-work-verify-193280` verified on disk, 5,554 passed; no golden fixture regenerated, confirmed by captured check `20260731-183223-work-verify-a31ae9`). Two changes required before merge, two minors that can ride.

Required before merge:

1. `localio._fsync_parent_directory` opens the parent with `O_NOFOLLOW` on the final component, so a target directory reached through a symlink now raises ELOOP after a successful `os.replace`. `write_text_atomic` has 30+ call sites outside the journal, and symlinked repo or workspace directories are a normal deployment shape. Resolve the parent before opening (or catch ELOOP and retry on `path.parent.resolve()`), and add a regression test that writes through a symlinked parent directory. The journal-side `_fsync_directory` keeps its strict no-follow contract; this generic helper should not inherit it.
2. Add the missing test for redaction of a range that contains a prior `run.redaction.recorded` anchor event. That case is the core attribution guarantee this PR exists to provide; it needs a pinned regression test either proving the anchor is preserved or proving the operation is refused.

Ride-along follow-ups (no merge block): idempotent-cleanup journal churn in `_refresh_chained_anchors`, and the dropped split-retirement digest cross-check flagged in review.

Re-request review after the two changes with a fresh full-suite receipt.

solomonneas:
Implemented the two required changes from the grader review.

1. `b81ed9ac` (`fix(io): support symlinked atomic-write parents`)
   - `localio._fsync_parent_directory` resolves the generic atomic-write parent before opening it with the existing directory and no-follow flags.
   - `run_journal._fsync_directory` is unchanged and still refuses a symlinked final directory.
   - Regression tests:
     - `test_write_text_atomic_allows_directory_symlink_parent`
     - `test_run_journal_fsync_directory_refuses_directory_symlink`

2. `54bce3db` (`test(redaction): pin overlapping anchor preservation`)
   - The regression reads the first operation's actual `run.redaction.recorded` sequence and includes that sequence in the second redaction range.
   - It requires a complete chain with no chain errors or partial tail, both operation anchors, matching `record_sha256` values, and the expected parent operation link.
   - Regression test: `test_redaction_preserves_first_anchor_when_second_range_contains_it`

Verification:

- Focused receipt `20260731-185508-work-verify-2ca7c7`: 3 passed.
- Full receipt `20260731-185534-work-verify-67a837`: `./scripts/verify` exited 0, with 5,556 passed, 3 skipped, and 82.99% coverage.

No fixture, golden, event-canonicalization, projector, `run_journal.py`, or `run_redaction.py` changes were made. The two ride-along minors remain out of scope.

solomonneas:
Grader final pass: both required changes verified. The localio fix resolves the parent before the no-follow open (one line, correct minimal blast radius), the journal-side strict helper is untouched and now has a pinning test, and the overlapping-anchor regression proves the first redaction anchor survives a second redaction whose range contains it, with matching record_sha256 and parent link. Receipts verified on disk: focused 2ca7c7 (3 passed) and full 67a837 (5,556 passed, 3 skipped, 82.99% coverage) at head 54bce3db with a clean tree. CI fully green. APPROVE - merging.

PR #641, opened 2026-08-01
feat(runs): durable cursors and idempotent live control (#604)

### LINKED ISSUE #604: runs: add durable cursors and idempotent live control on the lifecycle journal
## Problem

Issue #146 and PR #154 added artifact-only `runs watch` plus live `runs steer` and `runs interrupt` for app-server runs. The current reader keeps byte offsets only inside one `watch` process. A new process starts without a durable cursor, so a wrapper cannot reconnect at a known lifecycle event.

Live control has a related retry gap. PR #335 retries the short `no-active-turn` race, but callers cannot attach a stable request identity. If a client loses the response after Brigade applies a steer or interrupt, it cannot tell whether retrying will repeat the action.

Issue #568 defines the missing storage primitives: contiguous per-run sequence numbers, event IDs, request fingerprints, idempotency keys, and requested/observed pairs for external actions. It does not define the consumer or live-control CLI contract that uses them.

## Proposal

Add a lifecycle-only event reader and idempotent request identities for live control.

### Durable event cursor

Add:

```text
brigade runs events <run|latest> [--after <cursor>] [--follow] --json
```

- Read only the verified lifecycle journal from #568. Do not mix raw worker streams into this command.
- Emit newline-delimited lifecycle records in journal order.
- Include an opaque cursor on each record. The cursor binds the run ID, sequence, and event digest so it cannot be reused against another run or a rewritten prefix.
- `--after <cursor>` emits only committed events after that cursor.
- `--follow` waits for appended events and exits after a terminal lifecycle event.
- A cursor with the wrong run ID, missing sequence, or digest mismatch fails with a bounded diagnostic. It never falls back silently to `run.json` or raw worker streams.
- Legacy runs without a lifecycle journal report `legacy-no-journal`. Existing `runs watch` behavior remains available.

The cursor is client-owned. Brigade does not need a subscriber registry, acknowledgement store, broker, or daemon.

### Idempotent live control

Add a caller request identity to the existing commands:

```text
brigade runs steer <run> <worker> <text> --request-id <id>
brigade runs interrupt <run> [worker] --request-id <id>
```

- When omitted, Brigade generates the request ID and prints it in human and JSON output.
- Scope the ID to one run.
- Fingerprint the operation kind, run, worker, active turn identity when known, and a digest of the steering text. Do not place steering text in the lifecycle journal.
- Append and sync `control.requested` before contacting the live control transport.
- Append `control.observed` or `control.failed` after the transport returns.
- Reusing the same ID with the same fingerprint returns the committed result without calling the transport again.
- Reusing the same ID with a different fingerprint is a conflict.
- A request with `control.requested` but no terminal observation is `indeterminate`. Brigade reports it and does not reissue the control action automatically.

The transport cannot make an app-server side effect transactional. This contract provides safe retry inspection and an explicit indeterminate state. It does not claim exactly-once delivery.

## First slice

- Lifecycle events only, backed by #568.
- `runs events` with `--after`, `--follow`, and newline-delimited JSON.
- Request identities for app-server `steer` and `interrupt`.
- No resume, approval, provider transcript, worker-stream, network API, or UI changes.

## Acceptance criteria

- [ ] Stop a live `runs events --follow` reader, restart it with the last cursor, and receive each later lifecycle event once and in order.
- [ ] Reading a completed run with the same cursor produces the same suffix.
- [ ] A cursor for another run, an unknown sequence, a digest mismatch, or an unsupported schema fails without emitting an unverified suffix.
- [ ] Reusing a request ID and identical fingerprint returns the first terminal result without a second control-transport call.
- [ ] Reusing a request ID with a changed worker, text digest, operation, or turn identity returns a conflict.
- [ ] Crash tests cover exit before `control.requested`, after its sync, after the transport call, and after the terminal control event.
- [ ] A requested action with no terminal observation reports `indeterminate` and is not reissued automatically.
- [ ] Lifecycle events contain bounded metadata and digests, not steering text, model output, provider bodies, credentials, or raw tool arguments.
- [ ] Unknown event types and fields follow #568's fail-closed registry behavior.
- [ ] Existing `runs watch`, control commands without `--request-id`, and legacy run artifacts remain compatible.
- [ ] No daemon, network service, broker, subscriber database, or runtime dependency is added.

## Research notes

- The WHATWG [`Last-Event-ID`](https://html.spec.whatwg.org/multipage/server-sent-events.html#the-last-event-id-header) model keeps reconnect position with the reader and sends it back when reconnecting. Brigade can use the same client-owned pattern over a local append-only journal without adopting Server-Sent Events.
- The IETF HTTPAPI [`Idempotency-Key` draft](https://datatracker.ietf.org/doc/html/draft-ietf-httpapi-idempotency-key-header) distinguishes a completed duplicate, a concurrent request, and reuse of one key with a different payload. Brigade can apply those semantics to local CLI control requests without adopting an HTTP API.

These are protocol patterns, not dependencies.

## Dependencies and related work

- **Blocked by #568:** owns the lifecycle journal, sequence, digest chain, idempotency behavior, projection, recovery, and privacy boundary.
- **Builds on closed #146 and merged PR #154:** introduced detached dispatch, artifact watch, steer, and interrupt.
- **Builds on merged PR #335:** handles the transient no-active-turn race but does not give callers durable request identity.
- **Related #489:** resume can adopt the same request identity after its lifecycle contract lands. It is outside this first slice.
- **Related #592:** classifies and scrubs raw worker event streams. `runs events` does not consume those streams.
- **Related #595:** offline coordinator audit can consume the same verified lifecycle suffix. It does not block this reader contract.
- **Related #593:** budget lifecycle events should appear automatically once registered under #568.
- **Related #594:** each durable child run receives its own cursor namespace because it has its own run ID and journal.
- **Related #567:** preserves rewritten worker and synthesis sidecars. This issue reads lifecycle events only.
- **Related #578:** seat-health retries and fallback decisions may publish lifecycle events, but routing policy is separate.

## Non-goals

- Replay raw provider or worker traffic.
- Re-run historical control actions.
- Make provider calls exactly once.
- Persist subscriber acknowledgements.
- Replace `runs watch` or `run.json`.
- Add a web transport or shared control server.
- Recompose the route or choose fallback seats.

solomonneas:
## Summary

- Adds lifecycle-only `brigade runs events` with opaque durable cursors (`--after`, `--follow`, NDJSON).
- Registers status-neutral `control.requested` / `control.observed` / `control.failed` event types (reference fields and digests only; no steering text).
- Makes app-server `runs steer` / `runs interrupt` idempotent via `--request-id`, with replay, conflict, and indeterminate semantics on the journal.

Closes #604.

## Verify

```bash
```

Receipt ID: `20260801-032545-work-verify-0782e0`

## Acceptance criteria → tests

| #604 criterion | Test |
|---|---|
| Stop follow, restart with last cursor, later events once in order | `test_events_follow_restart_with_cursor_emits_each_later_event_once` |
| Completed run + same cursor → same suffix | `test_events_completed_run_same_cursor_produces_same_suffix` |
| Wrong run / unknown sequence / digest mismatch / unsupported schema fails closed | `test_events_cursor_failures_emit_no_unverified_suffix` |
| Same request id + fingerprint → no second transport call | `test_idempotent_request_id_replays_without_second_transport_call` |
| Same request id + changed worker/text/op/turn → conflict | `test_request_id_fingerprint_conflict` |
| Crash before requested / after sync / after transport / after terminal | `test_control_crash_points` |
| Requested without terminal → indeterminate, not reissued | `test_indeterminate_control_request_is_not_reissued` |
| No steering text / private payloads in lifecycle events | `test_control_payload_never_contains_steering_text`, `test_control_requested_rejects_*` |
| Unknown event types/fields fail closed (#568 registry) | `test_control_event_types_registered_with_closed_payload_keys`, `test_control_requested_rejects_unknown_payload_key_with_bounded_diagnostic` |
| `runs watch`, control without `--request-id`, legacy artifacts compatible | existing `tests/test_run_control.py` + `test_cli_explicit_request_id_on_legacy_run_fails` / `test_events_legacy_run_reports_legacy_no_journal` |
| No daemon / broker / runtime dependency | stdlib-only modules `run_event_cursor.py`, `run_control_journal.py` |

## Notes for reviewers

- Cursor codec salvaged from abandoned `codex/issue-604-*` worktrees; `run_id` tightened to the journal alphabet.
- Control appends use `append_event` under `checkpoint_event_pair` and sync shadow afterward so authoritative runs do not trip `journal-ahead-of-evidence`.
- Golden lifecycle fixtures are unchanged.

## Test plan

- [x] `./scripts/verify` via Brigade (receipt above)
- [ ] External grader review before merge

### DIFF
--- a/docs/command-inventory.md
+++ b/docs/command-inventory.md
@@ -56,7 +56,7 @@ enabled: run `brigade extras on` once, or set `BRIGADE_EXTRAS=1`.
 - `brigade route`: 1 command path(s)
 - `brigade run`: 1 command path(s)
 - `brigade runbook` (extras): 5 command path(s)
-- `brigade runs`: 9 command path(s)
+- `brigade runs`: 10 command path(s)
 - `brigade scrub`: 1 command path(s)
 - `brigade search`: 6 command path(s)
 - `brigade security`: 15 command path(s)
@@ -439,6 +439,7 @@ enabled: run `brigade extras on` once, or set `BRIGADE_EXTRAS=1`.
 - `brigade runbook plan` (extras)
 - `brigade runbook resume` (extras)
 - `brigade runbook run` (extras)
+- `brigade runs events`
 - `brigade runs interrupt`
 - `brigade runs latest`
 - `brigade runs list`
--- a/src/brigade/cli/runs.py
+++ b/src/brigade/cli/runs.py
@@ -62,6 +62,45 @@ def register(sub: argparse._SubParsersAction) -> None:
         default=1.0,
         help="Polling interval in seconds.",
     )
+    p_runs_events = runs_sub.add_parser(
+        "events",
+        help="Read verified lifecycle journal events with durable cursors.",
+    )
+    p_runs_events.add_argument("run", help="Run directory path, run id under --runs-dir, or 'latest'.")
+    p_runs_events.add_argument(
+        "--cwd",
+        type=Path,
+        default=Path("."),
+        help="Workspace whose default .brigade/runs directory should be used for run ids.",
+    )
+    p_runs_events.add_argument(
+        "--runs-dir",
+        type=Path,
+        default=None,
+        help="Explicit runs directory for run ids. Defaults to .brigade/runs under --cwd.",
+    )
+    p_runs_events.add_argument(
+        "--after",
+        default=None,
+        help="Opaque cursor; emit only committed lifecycle events after this position.",
+    )
+    p_runs_events.add_argument(
+        "--follow",
+        action="store_true",
+        help="Wait for appended lifecycle events and exit after a terminal event.",
+    )
+    p_runs_events.add_argument(
+        "--json",
+        action="store_true",
+        default=True,
+        help="Emit newline-delimited JSON records (default).",
+    )
+    p_runs_events.add_argument(
+        "--interval",
+        type=float,
+        default=1.0,
+        help="Polling interval in seconds when --follow is set.",
+    )
     p_runs_steer = runs_sub.add_parser("steer", help="Send steering text to an active app-server worker turn.")
     p_runs_steer.add_argument("run", help="Run directory path, run id under --runs-dir, or 'latest'.")
     p_runs_steer.add_argument("worker", help="Worker name to steer.")
@@ -78,6 +117,11 @@ def register(sub: argparse._SubParsersAction) -> None:
         default=None,
         help="Explicit runs directory for run ids. Defaults to .brigade/runs under --cwd.",
     )
+    p_runs_steer.add_argument(
+        "--request-id",
+        default=None,
+        help="Caller request identity for idempotent live control (generated when omitted).",
+    )
     p_runs_interrupt = runs_sub.add_parser("interrupt", help="Interrupt active app-server worker turns.")
     p_runs_interrupt.add_argument("run", help="Run directory path, run id under --runs-dir, or 'latest'.")
     p_runs_interrupt.add_argument("worker", nargs="?", default=None, help="Optional worker name to interrupt.")
@@ -92,6 +136,11 @@ def register(sub: argparse._SubParsersAction) -> None:
         type=Path,
         default=None,
         help="Explicit runs directory for run ids. Defaults to .brigade/runs under --cwd.",
+    )
+    p_runs_interrupt.add_argument(
+        "--request-id",
+        default=None,
+        help="Caller request identity for idempotent live control (generated when omitted).",
     )
     p_runs_recover = runs_sub.add_parser(
         "recover",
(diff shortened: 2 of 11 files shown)


solomonneas:
Post-merge grading review found one Critical and two Major defects in the control-journaling half of this PR - the cursor codec and runs events reader are sound and unaffected. Until #651 lands, treat live 'runs steer'/'interrupt' against journal-backed runs as unsafe: the control CLI is a second OS process appending to the lifecycle journal under a process-local lock, and a collision with the run owner forks the hash chain with no recovery path (details, file:line evidence, and the fix contract are in #651). If the cross-process lock cannot land promptly, revert the control-journaling half and keep the reader.

PR #655, opened 2026-08-01
fix(redaction): survive crash between parent retirement and child realign (#652)

### LINKED ISSUE #652: run-redaction: crash between parent retirement and child realign permanently fail-closes the lineage
Post-merge grading review of #644 (PR for #638) found one Major crash window introduced by the new retirement digest cross-check.

In `_retire_rewritten_digest_aliases` (src/brigade/run_redaction.py) the parent record is rewritten (adding `rewritten_digest_retired_by`) before `_realign_child_parent_record_digest` runs. A crash between those two fsync'd writes leaves the child's `parent_record_sha256` pinned to the parent's pre-retirement digest. The new cross-check in `_load_operation_inventory` (`child_record.get("parent_record_sha256") != record_digests.get(operation_id)`) then raises "redaction transaction state/record mismatch" on every subsequent inventory load, including the retry that would have realigned it. Pre-#644 this state was recoverable; now it requires manual record surgery. Quarantine data stays intact, so this fails closed, but redaction is the safety-critical path and a permanent fail-closed lineage with no automated repair is not acceptable.

Fix: tolerate the parent's pre-retirement record digest during the split window (accept either digest while a retirement marker is mid-flight), or realign-on-load. Add a crash-injection test that kills between the two writes and asserts the next inventory load succeeds and completes the realignment.

Also fold in (Minor): the no-op early return added to `_refresh_chained_anchors` happens before `_assert_active_owner`, dropping the ownership assertion on the read-only pass.

From the post-merge grading review of PR #644. Refs #638.

solomonneas:
## Summary

- Keep run-redaction lineage retryable if cleanup stops after the parent retirement record write and before child digest realignment.
- Accept the exact pre-retirement parent record digest only when validated record and state markers identify that in-flight split.
- Assert active ownership before `_refresh_chained_anchors` returns from its no-op path.

## Issue mapping

- Crash between parent retirement and child realignment: `test_cleanup_retry_after_parent_retirement_before_child_realign_crash`
- Ownership assertion on no-op anchor refresh: `test_refresh_chained_anchors_asserts_owner_on_noop_early_return`
- Existing tamper rejection remains covered by `test_tampered_split_retirement_record_fails_digest_cross_check`.

## Verification

- Brigade receipt `20260801-185244-work-verify-dad88f`: `./scripts/verify`, exit 0
- Result: 5,678 passed, 3 skipped, 83.04% coverage
- Outcome capture completed with signal `+1`. It emitted the known warning that `brigade-work` is not installed, but the receipt was written successfully.

The passing receipt came from a fresh Linux checkout at `5d6e81a6f964c4f014eda869d0a396fc2e84e701` with byte-identical copies of the two changed files. The exact command was first attempted locally, where Windows returned `[WinError 193] %1 is not a valid Win32 application` for the POSIX script.

Fixes #652.

### DIFF
--- a/src/brigade/run_redaction.py
+++ b/src/brigade/run_redaction.py
@@ -1481,20 +1481,97 @@ def _refresh_chained_anchors(
         limit=run_checkpoint.MAX_JOURNAL_BYTES,
         category="journal",
     )
+    # Ownership is asserted even on the no-op path so a stale lock cannot
+    # claim a successful refresh without re-checking active ownership.
+    _assert_active_owner(workspace, run_dir)
     if current == data:
         # Anchor payloads already match the record digests; skip journal and
         # projection rewrites so a second verify/cleanup pass is a no-op.
         return _digest(data)
-    _assert_active_owner(workspace, run_dir)
     _replace_journal(journal_path, data)
     snapshot = _load_json_object(run_dir / "run.json", limit=MAX_RUN_JSON_BYTES, category="run projection")
     _replace_projection(run_dir, _projection(snapshot, rewritten))
     return _digest(data)
 
 
+def _preretirement_parent_record_digest(
+    record: Mapping[str, Any],
+    *,
+    operation_id: str,
+    run_id: str,
+    rewritten_sha256: str,
+) -> str:
+    """Reconstruct the parent record digest from before retirement was recorded.
+
+    During the mid-flight window after the parent record gains
+    ``rewritten_digest_retired_by`` but before the child is realigned, the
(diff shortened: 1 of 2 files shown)


solomonneas:
Grader: APPROVE. The pre-retirement digest is accepted only inside the validated crash window and is reconstructed from validated fields rather than trusted from the child, so tamper detection is not weakened; ownership assertion correctly hoisted above the no-op return; existing tamper tests unchanged. CI matrices green. Note for the record: the cited receipt (dad88f) was produced on a remote Linux checkout not inspectable from the grading host; CI is the verification of record for this merge.

Issue #656, opened 2026-08-01
run: a failed run deletes its worktree, leaving uncommitted work recoverable only from changes.patch


solomonneas:
Observed during the #651 fix wave on 2026-08-01. Highest-value item from the orchestration failure record.

## What happened

Run `20260801-182926-facaecff` ended `incomplete` (5 workers failed or skipped). Its worktree was then removed: `~/.cache/brigade/worktrees/` has zero entries matching `facaecff`.

That worktree held 13 good, uncommitted regression tests - including a real two-process append race that reproduces the bug it was written for (`AssertionError: cross-process append race forked the journal: ['sequence gap/duplicate: expected 202, got 201']`). Nothing had been committed, because the seat that wrote the tests failed before the commit stage and the downstream seats recorded `skipped: prerequisite failed`.

The work survived only because `brigade run` independently writes `changes.patch` into the run artifact directory. Recovery was:

```
git worktree add /tmp/wt --detach <base-sha>
cd /tmp/wt && git apply .brigade/runs/<run-id>/changes.patch
```

Without that artifact the red suite would have been lost with no warning that anything was gone.

## Why it matters

A failing run is exactly the case where uncommitted work is most valuable: it is the evidence of what the run learned before it died. Deleting the worktree silently on failure means the default behavior destroys the most useful output of the least successful runs.

## Requested change

1. Retain the worktree when a run reaches a non-success terminal state and the worktree has uncommitted changes. Prune it on the next successful run of the same target, on an explicit `brigade runs prune`, or after a documented retention window.
2. If retention is refused for disk reasons, the run's terminal output must say explicitly that uncommitted work exists only in `changes.patch` and print the exact `git apply` recovery command with the base SHA filled in.
3. Either way, print the changed-file count from `changes.patch` in the failure summary so a non-empty patch is visible at a glance.

## Acceptance

- A run that fails with uncommitted changes leaves a recoverable worktree, or emits the recovery command with a real base SHA and file count.
- A run that fails with a clean worktree still prunes as it does today.
- Regression test covering both branches.

PR #660, opened 2026-08-01
fix(run-control): serialize journal appends across processes (#651)

### LINKED ISSUE #651: run-control: cross-process journal appends race the owner and can fork the chain (urgent fix-forward for #641)
Post-merge grading review of #641 (PR for #604) found one Critical and two Major defects in the control-journaling half. The cursor codec and `runs events` reader are sound and stay. Until this fix lands, treat live `runs steer`/`interrupt` against journal-backed runs as unsafe.

## Critical: cross-process journal appends have no mutual exclusion

`run_journal._append_critical_section` is a `threading.Lock` plus signal mask (run_journal.py:75, 114-126) - process-local. #641 makes `runs steer`/`interrupt` append `control.*` events from an external CLI process (run_control_journal.py:220-266) while the run owner appends dispatch facts and transitions mid-run (run_lifecycle.py:434-441, 642-649, 756-763, with no retry on StaleSequenceError). Failure modes:

- Both processes read tail sequence N and both append N+1 to the O_APPEND file: forked hash chain. Every subsequent read reports chain errors and a journal-authoritative run is permanently fail-closed. No recovery path exists for a forked chain (recover_partial_tail handles partial lines only).
- The CLI append lands between the owner's read and append: the owner's lifecycle write dies with LifecycleJournalError, un-retried.
- `interrupt` co-schedules the two writers by design (CLI appends control.observed at the same moment the owner reacts to the aborted turn).
- Same race against redaction: an external append between redaction's _verified_events read and _replace_journal (run_redaction.py:1430, 1492) silently drops the control event or forks the chain.

The docstring claim at run_control_journal.py:229-233 that checkpoint_event_pair prevents pair-splitting is false across processes; run_lifecycle.py:154-166 documents it as process-level.

Fix: an fcntl lock file beside the journal, taken inside _append_critical_section by EVERY append path (owner, control CLI, redaction, recovery), plus owner-side retry on stale sequence. If this cannot land promptly, revert the control-journaling half of #641 and keep the read-only cursor/events half.

## Major: no status or ownership guard on control journaling

execute_control_request checks only journal_present (run_control_journal.py:312-313) and control_transport_from_run reads persisted metadata that survives completion (run_control.py:436-454). `runs steer` against a completed run appends control.requested + control.failed to a terminal authoritative journal and rewrites shadow evidence as a non-owner (run_control_journal.py:206-217). Guard on run status and lock ownership before any control append.

## Major: unfiltered transport/provider text in control.failed payloads

detail is filled from str(exc) on transport exceptions (run_control_journal.py:363-369) and verbatim from the app-server response["error"] (:423-427), truncated to 512 chars; the allowlist legitimizes it (run_events.py:124). #604's criterion was bounded metadata and digests, not provider bodies. Replace with a bounded error-class enum plus digest reference, matching the established payload discipline.

## Also fold in (Minor, same surface)

- Concurrent same-request-id callers both see "absent" and both invoke the transport; compare the committed event against the submission to detect the idempotent replay.
- `runs events` stops at the first terminal event, making post-terminal events (run.redaction.recorded, control.*) unreachable through the cursor reader.
- Follow mode never revalidates chain continuity between polls; a redaction rewrite mid-follow emits a suffix that does not chain onto previously emitted events.

Regression tests must include a real two-process append race (fork or subprocess, not monkeypatch) proving the lock serializes writers, and a steer-against-completed-run refusal.

From the post-merge grading review of PR #641. Refs #604.

solomonneas:
Fixes #651.

Closes the Critical defect found in the post-merge grading review of #641: the `runs steer`/`interrupt` control CLI appends to a run's lifecycle journal from a second OS process, while `_append_critical_section` held only a `threading.Lock` plus signal mask. Two processes could each read tail sequence N and append N+1 to the O_APPEND file, forking the hash chain and permanently fail-closing a journal-authoritative run with no recovery path.

## Fix and regression map

| Item | Fix | Named regression tests |
| --- | --- | --- |
| Cross-process append race | Sibling `lifecycle.jsonl.lock` acquired with `fcntl.flock(LOCK_EX)` inside the existing critical section, taken by every append path: owner lifecycle writes, the control CLI, redaction, and recovery. Nesting is SIGTERM mask, then process lock, then file lock; unlock precedes close and a primary exception is preserved through cleanup. Hosts without `fcntl` or `pthread_sigmask` keep the prior in-process guards. | `test_cross_process_append_race_serializes` (real `subprocess` children, not monkeypatch) |
| Owner-side stale retry | Lifecycle writes retry on `StaleSequenceError` instead of dying when a control append lands between read and write. | `test_owner_retries_stale_sequence_after_external_append` |
| Status / ownership guard | Control appends refuse on a completed run and when the caller is not the active lock owner, instead of journaling into a terminal chain and rewriting shadow evidence as a non-owner. | `test_steer_against_completed_run_refuses_without_journaling` |
| Bounded `control.failed` payloads | Free transport and app-server error text replaced with a bounded error-class enum plus digest reference, matching the payload discipline in `run_events.EVENT_TYPES`. | `test_control_failed_payload_carries_bounded_error_class` |
| Minor: duplicate control effect | Concurrent same-request-id callers detected by comparing the committed event against the submission. | covered in `tests/test_run_events_cursor_control.py` |
| Minor: post-terminal events unreachable | `runs events` no longer stops at the first terminal event, so `run.redaction.recorded` and `control.*` are reachable through the cursor reader. | covered in `tests/test_run_events_cursor_control.py` |
| Minor: follow-mode chain continuity | Follow mode revalidates chain continuity between polls, so a redaction rewrite mid-follow cannot emit a suffix that fails to chain onto emitted events. | covered in `tests/test_run_events_cursor_control.py` |

No changes to event canonicalization, existing event types, or golden fixtures.

## Verification

- `brigade work verify run --target . --command "./scripts/verify" --capture brigade-work`
- Receipt `20260801-200416-work-verify-135e03`, exit 0: **5,675 passed, 3 skipped**
- Targeted receipt `20260801-200338-work-verify-138809`: 308 passed across the four affected suites, including all 13 previously-red tests

## Provenance note for the reviewer

This branch was assembled in two salvage steps, both from `changes.patch` artifacts of `brigade run` invocations that terminated non-successfully after producing good uncommitted work and then had their worktrees pruned (see #656):

- `0dcbe970` red suite, from run `20260801-182926-facaecff` (ended `incomplete`; a worker misread its base commit and cascaded `skipped: prerequisite failed`)
- `711e1fef` implementation, from run `20260801-191101-6612a08a` (ended `timeout` at 1500s after the implementation was complete)

The red suite was verified failing against the pre-fix tree before the implementation was applied: `AssertionError: cross-process append race forked the journal: ['sequence gap/duplicate: expected 202, got 201']`.

### DIFF
--- a/docs/issue-651-run-control-journal-race-plan.md
+++ b/docs/issue-651-run-control-journal-race-plan.md
@@ -1,0 +1,59 @@
+# Issue #651 implementation plan: cross-process journal serialization and safe live control
+
+Status: planned. This plan applies on top of #641, which introduced lifecycle-journal-backed control requests. It is intentionally ordered so the journal serialization primitive lands before callers depend on it.
+
+## Scope and invariants
+
+- Keep the existing event canonicalization and digest algorithms unchanged.
+- Keep the existing event-type registry unchanged: this work does not add, remove, or rename event types.
+- Do not edit golden journal fixtures. New tests must construct their own journals and assert their current canonical output through the existing APIs.
+- Continue to fail closed on an invalid, partial, or forked journal. This change prevents new forks; it does not invent recovery for an already forked journal.
+
+## Ordered implementation steps
+
+1. **Add one cross-process mutation lock to `run_journal`.**
+
+   - Derive the lock file as `journal_path.with_name(f"{journal_path.name}.lock")`, so a lifecycle journal at `events/lifecycle.jsonl` uses the private sibling `events/lifecycle.jsonl.lock`. Open or create that regular file with mode `0o600`, reject symlinks using the module's existing no-follow/identity safeguards, and leave the file in place after use.
+   - Change `_append_critical_section` to accept the journal path. Inside that context manager, retain the current SIGTERM deferral and in-process reentry protection, then acquire an exclusive advisory `fcntl.flock(fd, LOCK_EX)` on the sibling lock descriptor before any tail read, idempotency inspection, canonical build, write, fsync, truncate, or replace. Release with `LOCK_UN` and close the descriptor in `finally`, including when the protected mutation raises. A lock-release failure must not mask the primary exception.
+   - Provide one journal-owned mutation context for non-append mutations and make it enter `_append_critical_section(journal_path)` rather than creating another locking protocol. It must cover the full read/verify-to-replace window, not only `os.replace`.
+   - Route all mutation paths through that section: `append_event` and idempotency lookup, `recover_partial_tail`, owner lifecycle and dispatch/checkpoint append calls, external `runs steer` and `runs interrupt` control appends, the redaction anchor append, and every redaction journal re-chain or replacement path. Redaction must hold the section from its verified read through replacement so an external control append cannot be lost; reorganize nested calls around the locked journal primitive rather than taking a second lock descriptor recursively.
+   - Test: add `test_append_event_fcntl_lock_serializes_two_subprocess_writers` in `tests/test_run_journal.py`. Start two real Python subprocesses from a barrier, have each append after reading the same initial tail, and assert the resulting report has contiguous unique sequences, one digest-linked chain, and no `chain_errors`. Add `test_append_critical_section_releases_fcntl_lock_after_exception`, which injects a mutation failure and proves a later subprocess can acquire the lock and append. Extend `tests/test_run_redaction.py` with a barrier-backed redaction-versus-external-append test proving the final verified journal contains neither a dropped control line nor a fork.
+
+2. **Make owner lifecycle writes retry a stale tail without changing their request identity.**
+
+   - Add a single owner append helper in `run_lifecycle` and route `record_dispatch_fact`, `record_lifecycle_event`, and `record_lifecycle_transition` through it. Preserve the originally derived event payload and idempotency key for all attempts.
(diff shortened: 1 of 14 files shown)


solomonneas:
Grader: APPROVE, merging.

Verified independently rather than trusting the green suite, since I assembled this branch from two salvaged changes.patch artifacts and the tests share provenance with the implementation:

1. The race test is honest. Disabling `fcntl.flock` in a scratch copy makes `test_append_event_fcntl_lock_serializes_two_subprocess_writers` FAIL at test_run_journal.py:1990; restoring it makes the test pass. It spawns two real subprocess children synchronized through a filesystem barrier so both pass the stale check on the same sequence, across multiple rounds. It is not a vacuous pass.
2. Every append path takes the cross-process lock. Four call sites of `_append_critical_section`: append_event, lookup_idempotent_event, recover_partial_tail, and journal_mutation. Redaction routes through journal_mutation (run_redaction.py:1424, :2071) and the control CLI through run_control_journal.py:305. recover_partial_tail is now covered, closing an open Minor from the slice 1-6 review.
3. Full gate green: receipt 20260801-200416-work-verify-135e03, 5,675 passed, 3 skipped, exit 0. CI fully green.

This closes the Critical introduced by #641. Live steer/interrupt on journal-backed runs is safe again.

PR #662, opened 2026-08-01
fix(dispatch): bound ledger root, name missing interpreters, surface stale inventory (#657)

### LINKED ISSUE #657: dispatch ergonomics: nested outcome ledger creation, missing-interpreter error, stale command-inventory CI message
Three separate traps hit during one dispatch wave on 2026-08-01. All present as opaque failures that cost a worker seat its budget or an operator a debugging cycle. Grouped because each fix is a better error message plus a guard.

## 1. `brigade` creates an outcome ledger in a nested subdirectory without complaint

Run `20260801-184731-c3c933c1` dispatched the `grok` seat to do Go work under `stations/notify/`. The seat timed out after 900s, and its `changes.patch` contains exactly two new files, both wrong:

```
stations/notify/memory/outcome/.records.lock
stations/notify/memory/outcome/records.jsonl
```

The seat ran brigade commands with cwd inside `stations/notify/`, and brigade initialized a fresh outcome ledger there rather than resolving to the repo root or refusing. A ledger inside a Go station directory is never correct. Requested: resolve the outcome ledger to the target/repo root, or refuse with a bounded error naming the resolved root, when invoked from a nested directory that is not itself a configured target.

## 2. A missing interpreter is reported as a rejected receipt, not a missing interpreter

`brigade work verify run --command "python -m pytest ..."` is rejected before pytest starts on any machine that has only `python3` on PATH. The result reads as an infrastructure/rejected receipt with no obvious cause. In run `20260801-182926-facaecff` this contributed directly to a seat's confusion and the loss of the run (receipt `20260801-190656`). Requested: name the missing interpreter in the error and suggest the available one (`python3`).

## 3. `repo-metadata` CI fails with a bare exit code when the command inventory is stale

Adding a CLI subcommand without regenerating `docs/command-inventory.md` fails the `repo-metadata` job with `##[error]Process completed with exit code 1`. The actual cause (`roadmap_command_inventory_current: docs/command-inventory.md missing or stale`) is buried in the step log. This hit two independent PRs in one day (#645 adding `outcome repair`, #649 adding `runs audit`). Requested: surface the stale-file name and the regeneration command in the job's failure output.

## Also worth checking

Every run in this wave logged `warning: dag scheduler: plan not fully covered; falling back to wave scheduler`. Either plan coverage is genuinely incomplete for ordinary plans or the coverage check is too strict; the fallback is silent enough that nobody would notice the DAG scheduler is effectively never used.

Context and evidence: `.claude/memory-handoffs/2026-08-01-orchestration-failure-record.md`.

solomonneas:
## What and why

Bound outcome writes to configured targets, make missing Python interpreters actionable, and annotate stale command inventory failures with the regeneration command. The DAG investigation also found that admission rejected safe coversless root assignments, so that narrow check is corrected here.

Closes #657

## Regression coverage

- Nested outcome ledger: `test_outcome_capture_reports_bounded_error_for_nested_git_target` verifies the bounded nested-target error and absence of a misplaced ledger. `test_append_records_allows_configured_nested_git_target` protects configured nested targets.
- Missing interpreter: `test_verify_run_rejects_missing_python_with_python3_suggestion` pins the actionable Python 3 text. `test_verify_run_rejects_missing_python_with_inline_path_and_python3_suggestion` covers the command's effective PATH.
- Stale command inventory: `test_repo_metadata_command_inventory_failure_is_actionable_and_preserves_status` pins the GitHub error annotation, `docs/command-inventory.md`, the exact regeneration command, and preserved exit status.
- DAG fallback: `test_empty_covers_runs_as_independent_dag_root` confirms a coversless assignment runs as an independent root without a wave fallback. Existing and added unknown-cover tests keep malformed plans on the fallback path.

## Verification

`brigade work verify run --target . --command "./scripts/verify" --capture brigade-work`

Receipt: `20260801-211405-work-verify-0b99de`

Result: 5,687 passed, 3 skipped; coverage 83.04%. Ruff, Ruff format, version sync, mypy, managed snapshots, and the full pytest suite passed.

## Checklist

- [x] Bug fix
- [x] Added regression tests covering each error message
- [x] No command inventory regeneration in this branch
- [x] Conventional commit

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -296,7 +296,15 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install -e ".[dev]"
       - name: Command Inventory
-        run: brigade roadmap commands --check
+        run: |
+          set +e
+          brigade roadmap commands --check
+          status=$?
+          set -e
+          if [ "$status" -ne 0 ]; then
+            echo "::error file=docs/command-inventory.md::docs/command-inventory.md is stale; regenerate with BRIGADE_EXTRAS=1 brigade roadmap commands --write"
+          fi
+          exit "$status"
         env:
           BRIGADE_EXTRAS: "1"
       - name: Template Audit
--- a/src/brigade/outcome_cmd.py
+++ b/src/brigade/outcome_cmd.py
@@ -858,6 +858,17 @@ def append_records(target: Path, records: list[core.OutcomeRecord]) -> None:
     interrupted batch may leave earlier rows from that call on disk while the
     partial tail is removed before validation and the next append proceeds.
     """
+    resolved_target = target.expanduser().resolve()
+    from .work_cmd import helpers as work_helpers
+
+    repo_root = work_helpers._git_value(resolved_target, "rev-parse", "--show-toplevel")
+    if repo_root is not None:
+        resolved_root = Path(repo_root).resolve()
+        if resolved_target != resolved_root and not (resolved_target / ".brigade" / "config.json").is_file():
+            raise OutcomeLedgerError(
+                f"outcome ledger target {resolved_target} is inside git repository {resolved_root} "
+                f"but lacks .brigade/config.json; rerun with --target {resolved_root}"
+            )
     path = _records_path(target)
     lock_path = _records_lock_path(target)
     path.parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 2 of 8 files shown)


solomonneas:
Grader: APPROVE, merging. All three traps fixed with named regression tests pinning the improved error text, not just the behavior: nested ledger creation now fails with a bounded error naming the resolved root, a missing interpreter names python3, and the repo-metadata job annotates the stale file plus the exact regeneration command. The DAG investigation went further than asked and found a real defect (admission rejected safe coverless root assignments), fixed narrowly with test_empty_covers_runs_as_independent_dag_root. Receipt 20260801-211405-work-verify-0b99de, 5,687 passed, CI green.

PR #663, opened 2026-08-01
fix(run): retain the worktree of a failed run with uncommitted work (#656)

### LINKED ISSUE #656: run: a failed run deletes its worktree, leaving uncommitted work recoverable only from changes.patch

solomonneas:
Closes #656.

## Summary

- Retain a `brigade run --worktree` checkout when the run returns a non-success
  code and its validated `changes.patch` is non-empty.
- Keep the existing cleanup path for successful runs and non-success runs with
  clean worktrees.
- Print the patch file count and retained checkout path in terminal output.
- Document the retained-worktree lifecycle in the technical guide.

## Retention lifecycle

This PR prunes older retained worktrees on the next successful `--worktree`
run for the same target. Cleanup reads Git's registered worktrees from the git
root and removes only older Brigade cache entries with that target's prefix.
If Git cannot list worktrees, the successful run stays successful and the older
checkout remains available for a later cleanup attempt.

## Acceptance coverage

| Criterion | Regression test |
| --- | --- |
| A non-success run with uncommitted changes leaves a recoverable checkout and prints the real patch file count. | `test_run_cli_worktree_retains_failed_checkout_with_changes` |
| A non-success run with a clean checkout prunes it as before. | `test_run_cli_worktree_prunes_failed_checkout_without_changes` |
| The next successful run for the same target prunes an older retained checkout. | `test_run_cli_success_prunes_retained_worktree_for_same_target` |
| Same-target cleanup also works when `--cwd` is a repository subdirectory. | `test_run_cli_success_prunes_retained_worktree_for_same_target_from_subdirectory` |

## Verification

Command:

```text
```

Receipt: `20260801-211802-work-verify-fe92bd`

Result: 5,682 passed, 3 skipped, 83.04% coverage.

### DIFF
--- a/docs/technical-guide.md
+++ b/docs/technical-guide.md
@@ -233,9 +233,14 @@ takes a local `.brigade/run.lock` (stale locks from dead processes are
 replaced automatically) so two runs do not mutate the same checkout at once.
 Use `--worktree` to run agents in a detached checkout from `HEAD` under
 `~/.cache/brigade/worktrees/`; Brigade writes the resulting diff to
-`changes.patch` in the run artifacts, removes the temporary checkout, and
-leaves the original checkout unchanged. `--worktree` requires run artifacts,
-so it cannot be combined with `--no-artifacts`.
+`changes.patch` in the run artifacts and leaves the original checkout
+unchanged. Successful runs and clean non-success runs remove the temporary
+checkout; failed, timed-out, canceled, and incomplete runs with uncommitted
+changes retain it for recovery. The next successful `--worktree` run for the
+same target prunes older retained entries. Terminal output shows the
+`changes.patch` file count and, when a checkout is retained, its path.
+`--worktree` requires run artifacts, so it cannot be combined with
+`--no-artifacts`.
 
 Start with a roster:
 
--- a/src/brigade/cli/run.py
+++ b/src/brigade/cli/run.py
@@ -599,7 +599,7 @@ def dispatch(args) -> int:
                         tracked_count=summary.tracked_count,
                         untracked_count=summary.untracked_count,
                     )
-                    keep_worktree = False
+                    keep_worktree = rc != 0 and summary.changed
                     if summary.changed:
                         print(
                             f"changes: {summary.path} ({summary.tracked_count + summary.untracked_count} file(s))",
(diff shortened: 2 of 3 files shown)


solomonneas:
Grader: APPROVE, merging. Retention now covers timed-out, canceled, and incomplete runs holding uncommitted changes, prunes retained entries on the next successful worktree run for the same target, and reports the retained path plus patch file count in the failure summary. The session also caught a gap in its own first fix (prune invoked from a subdirectory) and added test_run_cli_success_prunes_retained_worktree_for_same_target_from_subdirectory to cover it. Receipt 20260801-211802-work-verify-fe92bd, 5,682 passed, CI green.

This closes the defect that cost two recoveries today: both the 13 red tests and the 35KB implementation for #651 lived only in changes.patch after their runs failed and their worktrees were pruned.

Issue #654, opened 2026-08-01
grading follow-ups: fixture pinning (#640), export temp-file gap (#646), outcome margins and dedup durability (#648/#650)


solomonneas:
Non-urgent findings from the post-merge grading review of PRs #640, #646, #648, #650. Batched; none blocks anything.

## #640 reader fixture (test-only)

- The fixture omits v0.25.0's finished_at-implies-terminal rule, so the paused variant never pins the one compat property it exists for (deleting the finished_at pop at aboyeur.py:2491 stays green). Add the property.
- The "legacy" snapshot variant is a four-key dict the v0.25.0 writer could never have produced (it always emitted cwd, dry_run, read_only, ...). Generate it from the real writer shape.
- parse_behavior asserts are tautological (fixture constants asserted against the fixture).
- derived_from.commit records the tag object SHA (949d6df7) instead of the commit (5c1dcba5).

## #646 checkpoint export stripping

- Strip and assert cover only *.json: a crashed checkpoint temp (.checkpoint.*.tmp, run_checkpoint.py:568) holding a full private body is copied and passes the assert. Extend both to the temp pattern.
- A checkpoint whose filename does not match its content hash makes _expected_verify_archive_manifest raise, leaving that verify run permanently unprunable (fail-closed but unbounded retention). Bound it.

## #648 regression forgiveness

- The non-scorecard decide path installs on a margin of one (helped=2, hurt=1) with no statistical gate. Consider helped >= hurt + install_min_helped or a Wilson floor, especially while pre-#647 ledgers still hold mixed-signal generic buckets.

## #650 reused-receipt dedup

- Legacy-row canonicalization depends on the reused receipt file still existing on disk; once receipts are pruned the double count silently returns. Persist the canonical ref (or a resolved marker) instead of re-deriving per read.
- Every rank/score/reconcile now performs O(rows) receipt-file opens (1,600+ on the live ledger); add a cache.
- The ledger row for a reused capture no longer references the reused receipt itself; that provenance lives only in the receipts dir.

From the post-merge grading review. Refs #637, #636, #629, #634.

solomonneas:
## Scope narrowing: the #646 section is done (verified 2026-08-03 on `ad5ae46f`)

Both `#646 checkpoint export stripping` bullets have landed via PR #668. Striking that section; the other three remain.

[...]

PR #668, opened 2026-08-02
fix(checkpoint): harden verification archive export

solomonneas:
## Summary

- Replace exported checkpoint bodies with bounded artifact references.
- Keep raw checkpoint paths, temp names, bodies, and body digests out of the verification archive index.
- Make existing archive destinations immutable and fail closed on legacy raw archives.
- Add byte-and-mode immutability tests for valid bodies and crashed temp files across retries.

## Verification

- `tests/test_work_cmd_verification.py`: 125 passed
- `tests/test_run_checkpoint.py`: 138 passed
- `./scripts/verify` through Brigade before the current-main merge: pass
- Focused archive/checkpoint suites after the current-main merge: pass
- Independent security review: no blocking findings

Refs #654

This is the checkpoint-export slice only and does not close the broader batch issue.

### DIFF
--- a/.claude/memory-handoffs/2026-08-02-0739-pr668-fix-temp-decoding-retention-gap.md
+++ b/.claude/memory-handoffs/2026-08-02-0739-pr668-fix-temp-decoding-retention-gap.md
@@ -1,0 +1,55 @@
+# Memory Handoff
+
+## Type
+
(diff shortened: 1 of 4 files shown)


solomonneas:
@coderabbitai review

PR #716, opened 2026-08-05
fix(update): run the background version refresh on piped stderr

### LINKED ISSUE #698: update refresh is also TTY-gated, so an agent-only install never populates the version cache
## Problem

Split out from #693 rather than widening its fix.

`update_notify.maybe_notify` returns early on the TTY gate **before** it ever spawns the background refresh:

```python
if _gated(list(argv), exit_code, environment, err):
    return

path = cache_path(environment)
state = localio.read_json_dict(path) or {}
...
checked_at = state.get("checked_at")
stale = not isinstance(checked_at, (int, float)) or current_time - float(checked_at) >= CHECK_INTERVAL_SECONDS
if stale:
    spawn_refresh()
```

`_gated` is True whenever stderr is not a TTY. Every agent harness pipes stderr, so on an install where no human ever runs `brigade` in a terminal, `spawn_refresh()` is unreachable and `~/.cache/brigade/update-notify.json` is never written.

The consequence is worse than a missing notice: the cache holds no data at all, so any consumer of that cache reports nothing regardless of how it presents it.

## Why it is not visible on this machine

The operator does occasionally run `brigade` interactively, so the cache here is populated:

```
/home/clawdbot/.cache/brigade/update-notify.json
{"checked_at": 1785717226.3494148, "latest": "0.25.0"}
```

That masks the defect locally. A pure agent-only install has no such path.

## Why the two gates should be separated

The TTY gate exists to keep stray bytes out of agent stderr, which is a real hazard this repo has already been bitten by. That reasoning applies to **printing**. It does not apply to **refreshing a cache in a detached child process**, which writes nothing to stderr and is already documented as never blocking the caller.

## Proposed fix

Split the gate. Keep the print gated on TTY. Let the refresh run whenever the cache is stale and `BRIGADE_NO_UPDATE_CHECK` and `CI` are unset, regardless of TTY.

Preserve every existing guarantee: never raise, never block on the network, never alter the exit code, no network activity at all when `BRIGADE_NO_UPDATE_CHECK` is set.

## Relationship to #693

#693 adds `available_update()` and surfaces it in `brigade work brief`. That change is correct and independently useful, since it reports whatever the cache knows. But on an agent-only install the cache stays empty until this issue is fixed, so the two together are what actually deliver update awareness to agents.

solomonneas:
Fixes #698. Splits the single update-notice gate into a notice gate (unchanged: TTY, exit 0, CI, opt-out, skip commands) and a refresh gate (opt-out, CI, update/completions only), so agent-only installs finally populate the version cache that the work brief's update_available surface (#699) reads.

Verified: focused tests replaced the monolithic gate test (26 pass), full ./scripts/verify green via brigade work verify run (fresh receipt, reused_from null). The deliberate call: the refresh is NOT gated on exit code, so failed agent commands still warm the cache; rationale in the module docstring.

Part of the 0.26.0 release-readiness push; recommend milestone 0.26.0 per the 2026-08-04 assessment.

### DIFF
--- a/docs/update-channels.md
+++ b/docs/update-channels.md
@@ -42,8 +42,11 @@ background process sends one HTTPS GET to
 `https://check.brigade.tools/v1/version`. The request has no query
 parameters, no body, and no install id. The User-Agent carries the brigade
 version and OS name. Raw IPs are never stored server-side (a weekly-salted
-hash backs an aggregate weekly-active count). The notice is skipped entirely
-when stderr is not a TTY, when `CI` is set, or when the command failed.
+hash backs an aggregate weekly-active count). The stderr notice is skipped
+when stderr is not a TTY, when `CI` is set, or when the command failed. The
+background cache refresh still runs when stderr is piped (agent harnesses) so
+`work brief` can surface available updates; it is skipped when `CI` is set or
+`BRIGADE_NO_UPDATE_CHECK` is set.
 
 Opt out completely (no notice, no network, ever):
 
--- a/src/brigade/update_notify.py
+++ b/src/brigade/update_notify.py
@@ -5,6 +5,10 @@ detached child process so no user command ever waits on it. The endpoint
 request carries no parameters and no install id - the User-Agent (version +
 OS) plus the request itself is the entire signal. BRIGADE_NO_UPDATE_CHECK
 disables both the notice and all network activity.
+
+The stderr notice is TTY-only (interactive shells). The background cache
+refresh runs even when stderr is piped so agent harness installs populate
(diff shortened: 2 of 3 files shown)


solomonneas:
Correction for the record: the PR was opened before the first full-gate result was read - that run failed on a ruff-format violation (tests all passed). The format fix is now pushed and the full ./scripts/verify gate is green on this exact head: receipt 20260805 in the lane worktree, reused_from null, 5920 passed. One unrelated flake (test_sync_write_then_resync_is_idempotent[grok]) failed once while concurrent agent seats were mutating user scope, and passes in isolation.

Issue #721, opened 2026-08-05
center: read-only operator dashboard served locally over existing --json contracts


solomonneas:
## Problem

The operator loop is spread across a dozen CLI invocations: `brigade work brief`, `handoff lint`, `outcome rank`, `memory care status`, run receipts under `.brigade/work/verify-runs/`. Each is fine alone, but there is no single glanceable surface for "what is the state of this workspace right now". The roadmap already names a workspace UI as a view over the CLI; this issue scopes the first slice.

## Proposal

A local HTTP dashboard started by an explicit command (proposed spelling: `brigade center serve`), stopped with ctrl-c, never a daemon. Architecture rules:

- **Pure presentation layer.** The server renders JSON that existing commands already emit (`work status --json`, `handoff lint --json`, `outcome rank --json`, `memory care status --json`, run receipts). No parallel implementation, no new state, no writes. If a view needs data the CLI cannot produce, the JSON contract grows first and the dashboard consumes it second.
- **Zero build step.** One HTML file plus a small stdlib HTTP server. No node toolchain, no bundler, no frontend dependency tree to audit. Refresh via polling (30s is plenty for this data); no websockets in the first slice.
- **Read-only.** No mutating endpoints at all in v1. Approvals, ingest, and reconcile stay in the terminal where the gate semantics already live.

### Views for the first slice

1. **Status grid**: pending handoffs, cards at refresh risk (memory-care), latest verify receipt and its signal, outcome ledger totals, open work-inbox items.
2. **Handoff inbox**: the pending/reviewed queue with lint state per file.
3. **Memory cards**: table with title, tags, reviewed date, freshness; text filter backed by the existing search scoring.
4. **Outcomes**: `outcome rank` rendering with capability cohort and recency columns.
5. **Run timeline**: verify runs and receipts in reverse-chronological order, expandable to the receipt JSON.

A 52-week activity heatmap over verify runs and captures is a cheap sixth view and makes dormancy (the "installed-but-dormant" failure mode) visible at a glance.

## Security floor (non-negotiable, day one)

Local dashboards get popped through exactly three doors, so close them before writing the first view:

- Bind 127.0.0.1 by default. Refuse to bind anything else unless both a bearer token and an explicit allowed-hosts list are configured; fail startup with a clear error otherwise.
- Validate the `Host` header against an allowlist on every request (DNS rebinding turns "localhost-only" into "any website in your browser can read it"). Reject unknown hosts with 403 even for static routes.
[...]

PR #729, opened 2026-08-05
feat(center): add a read-only local dashboard server with a security floor

solomonneas:
First slice of #721. This lands the transport and its security floor before any
view exists, so the guards are in place before there is anything to attack.

Deliberately no dashboard views here. `GET /` returns a placeholder page whose
only job is to prove the nonce and delegated-listener wiring works. The six
views follow in a second PR, which will close the issue.

## What it does

`brigade center serve` starts on an explicit command, binds `127.0.0.1`, prints
the URL, and stops on ctrl-c. Python standard library only: no new dependency,
no build step, no daemon, no mutating route. Registered under the existing
extras-gated `center` group, following the `register`/`dispatch` pattern and the
`center_cmd` facade's `_MODULE_NAMES` tuple.

## Security floor

Threat model is DNS rebinding plus XSS from card and handoff text, which are
attacker-influenced input.

- **Host allowlist on every request**, enforced in `handle_one_request` after
  `parse_request()` and before verb dispatch. Placement matters: a guard called
  from each `do_*` method leaves any verb without a handler falling through to
  `BaseHTTPRequestHandler`'s built-in 501 responder, which consults no allowlist
  and emits none of the security headers. `OPTIONS` and `TRACE` were answerable
  with a forged Host under the per-method arrangement.
- **Port-pinned entries.** Allowlist values derived from the bind carry the bound
  port, so `Host: 127.0.0.1:1` does not pass. Bare any-port matching is reserved
  for operator-supplied `--allowed-host` values, where a reverse proxy makes it
  meaningful. Loopback binds also accept `localhost` and `::1`, which are not
  attacker-controllable names.
- **Bind refusal.** A non-loopback `--host` is refused unless both `--token` and
  at least one `--allowed-host` are set. `0.0.0.0` and `::` count as non-loopback.
  Refusal is a clear stderr message and exit 2, not a traceback.
- **CSP**: `default-src 'none'`, `script-src 'nonce-<per-response>'`,
  `script-src-attr 'none'`, plus `nosniff`, `no-referrer`, `no-store`. The nonce
  is generated per response and shared by the header and the `<script>` tag.
- **No inline event handlers.** The page uses one delegated listener keyed off a
  `data-action` attribute.
- **Constant-time token comparison** via `hmac.compare_digest` over encoded
  bytes, so a non-ASCII token cannot turn a wrong password into a 500.
- **Read-only**: every verb other than GET/HEAD returns 405.

`ThreadingHTTPServer` rather than `HTTPServer`, so one stuck client cannot wedge
the polling refresh the views will rely on.

## Tests

`tests/test_center_serve.py`, 27 cases. The Host tests drive a **raw socket**
because `urllib`/`http.client` rewrite the Host header and would pass against a
broken guard.

- forged Host returns 403, with a good-Host 200 control
- the guard applies to unknown paths and to `OPTIONS`/`TRACE`/`POST`, and those
  403s still carry CSP
- `localhost` accepted on a loopback bind; a different port rejected; a missing
  Host header rejected
- bind-refusal matrix, including the command-level exit code and stderr text
- CSP assertions, and that the nonce differs between two responses
- HTML scanned for `on\w+\s*=`, asserting zero matches, and the body nonce
  matching the header nonce
- token: missing 401, wrong 401, correct 200
- every non-GET verb returns 405

## Docs

`docs/command-inventory.md` regenerated with the repo's generator
(`BRIGADE_EXTRAS=1 brigade roadmap commands --write`), run from this worktree's
venv so it reads the local parser. `--check` reported the inventory stale
beforehand. The diff is exactly the new `brigade center serve` path and the
group count moving 29 to 30.

## Verification

```
```

`./scripts/verify [completed] exit=0` - 5990 passed, 3 skipped, 68 subtests
passed in 684.64s. Coverage 83.09% against the 78% floor.
Receipt: `20260805-163215-work-verify-f0749b`.

Refs #721.

### DIFF
--- a/docs/command-inventory.md
+++ b/docs/command-inventory.md
@@ -17,7 +17,7 @@ enabled: run `brigade extras on` once, or set `BRIGADE_EXTRAS=1`.
 
 - `brigade add`: 1 command path(s)
 - `brigade budgets` (extras): 2 command path(s)
-- `brigade center` (extras): 29 command path(s)
+- `brigade center` (extras): 30 command path(s)
 - `brigade chat` (extras): 7 command path(s)
 - `brigade code`: 16 command path(s)
 - `brigade completions`: 1 command path(s)
@@ -104,6 +104,7 @@ enabled: run `brigade extras on` once, or set `BRIGADE_EXTRAS=1`.
 - `brigade center report show` (extras)
 - `brigade center reviews` (extras)
 - `brigade center schema` (extras)
+- `brigade center serve` (extras)
 - `brigade center status` (extras)
 - `brigade center templates` (extras)
 - `brigade chat surfaces doctor` (extras)
--- a/src/brigade/center_cmd/__init__.py
+++ b/src/brigade/center_cmd/__init__.py
@@ -15,6 +15,7 @@ _MODULE_NAMES = (
     "reports",
     "report_review",
     "actions",
+    "serve",
 )
 _MODULE_ALIASES = {
     "schema": "schema_ops",
@@ -89,6 +90,7 @@ _module_readiness = importlib.import_module(f"{__name__}.readiness")
 _module_reports = importlib.import_module(f"{__name__}.reports")
 _module_report_review = _facade_module("report_review")
 _module_actions = importlib.import_module(f"{__name__}.actions")
+_module_serve = importlib.import_module(f"{__name__}.serve")
 
 from .schema_ops import *
 from .core import *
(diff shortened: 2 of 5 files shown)


solomonneas (commented):
Rebased onto current main. The Opus reviewer run completed, but Brigade did not preserve a verdict or findings in the run artifacts, so approval cannot be established. Leaving this open instead of merging.

PR #730, opened 2026-08-05
feat(handoff): lint durable facts for unresolved references

### LINKED ISSUE #725: handoff lint: flag unresolved references so durable facts read standalone
## Problem

Handoffs are written at the end of a session by an author (human or agent) who still has the session in their head. The result is durable-facts bullets like "the fix was reverting it", "that file also needs the guard", "same error as before". Filed as-is, these become cards that are unintelligible without the session that produced them, and search can never surface them usefully because the referent ("it", "that file", "before") is not in the text. This is the single most common way a memory system silently rots: every entry is true and none is retrievable.

## Proposal

Add a **standalone-readability check** to `brigade handoff lint`, mechanical first, no LLM required:

- Scan `Durable facts` bullets (and `Suggested card content` bodies) for unresolved references:
  - leading or bare pronouns with no in-sentence antecedent: "it", "this", "that", "they" as subject;
  - deictic phrases: "that file", "this repo", "the above", "as before", "the same error", "the earlier fix";
  - relative time in durable facts: "yesterday", "last week", "recently" (absolute dates are already the rule; lint should enforce it).
- Emit a warning per hit with the line and the pattern matched, e.g. `[warn] durable fact has unresolved reference ("that file"): name the path`.
- Warnings, not errors, by default: prose legitimately uses pronouns with in-bullet antecedents, and the patterns will have false positives. A `--strict` flag can promote them for CI-style use.
- Heuristic scope is deliberately narrow: only the sections that get filed into canonical memory, not Summary or Evidence, which are allowed to lean on session context.

### Later (separate slice, optional)

An enrichment mode that rewrites flagged references using the rest of the handoff plus session artifacts as context, proposing the resolved text for the author to accept. That needs a model in the loop and belongs behind the same explicit-command gate as everything else. The lint rule is valuable alone and ships first.

## Acceptance criteria

- `handoff lint` warns on the pattern classes above within Durable facts and Suggested card content only.
- Relative-date detection covers the common English forms and suggests converting to an absolute date.
- Zero new dependencies; pure regex/token heuristics with a documented pattern list.
- `--strict` turns the warnings into lint failures.
- Fixture handoffs cover: bare-pronoun fact (warns), deictic phrase (warns), pronoun with clear in-bullet antecedent (does not warn), relative date (warns), compliant handoff (clean).

solomonneas:
Closes #725.

## What

Adds a standalone-readability check to `brigade handoff lint`. It scans only the sections that get filed into canonical memory, `Durable facts` and `Suggested card content`, and warns when a durable fact cannot be read without the session that produced it.

Three pattern classes, all pure stdlib regex with the list documented in `src/brigade/handoff_cmd/readability.py`:

- **bare-pronoun** - a bullet that opens with `it`, `this`, `that`, `they`, `them`, `their`, `these`, `those`. Only the first word of a bullet can lack an antecedent, so a mid-bullet pronoun does not warn.
- **deictic** - `that file`, `this repo`, `the above`, `as before`, `the same error`, `the earlier fix`, and similar pointers.
- **relative-date** - `yesterday`, `last week`, `recently`, `3 days ago`, and similar, with a suggestion to use an absolute date.

Zero new dependencies.

## Compatibility

Findings are advisory by default. `lint_file` reports them as warnings and never marks a handoff invalid, because nine production call sites gate on its `valid` flag (`handoff_cmd/drafts.py`, `daily_cmd/closeout_ops.py`, `phases_cmd/checks_actions.py`, `work_cmd/imports.py`, `research_cmd.py`, `handoff_cmd/migrate_ops.py`). `--strict` promotes them to failures inside the `lint` command only, leaving `lint_file` untouched.

JSON output changes are additive: `readability` on each result, `readability_flagged_count` and `strict` at the top level. Every pre-existing key keeps its name and type, and exit codes stay 0 on valid / 1 on invalid.

## Notes

Fixed along the way: this repo allows `Suggested card content` to be wrapped in a markdown fence (`test_handoff_lint_accepts_fenced_card_content`), and a fence-skipping scanner silently skipped the entire card body. Fence state now resets at each `## ` heading, and a fence that opens a section is treated as a content wrapper rather than a code block. A fenced card containing `It was reverted last week because that file was wrong.` went from 0 findings to 3.

## Tests

`tests/test_handoff_readability.py` plus five fixture handoffs under `tests/fixtures/handoff_lint/readability/`, covering the acceptance criteria: bare-pronoun warns, deictic warns, relative date warns, a pronoun with a clear in-bullet antecedent does not warn, and a compliant handoff is clean. Also covered: out-of-scope sections are ignored, fenced card content is scanned, `--strict` exits 1, and the default JSON shape stays compatible.

## Verification

```
- ./scripts/verify [completed] exit=0
5973 passed, 3 skipped, 68 subtests passed in 679.98s (0:11:19)
Required test coverage of 78% reached. Total coverage: 83.14%
```

Receipt: `.brigade/work/verify-runs/20260805-165417-work-verify-92a54c/receipt.json`


solomonneas (commented):
Rebased onto current main. The repo-metadata check is failing, and Brigade did not preserve the Opus review verdict or findings in the run artifacts. Leaving this open instead of merging.

PR #750, opened 2026-08-06
ci: make the publish workflow rerunnable end to end

### LINKED ISSUE #749: publish: rerunning the workflow after a post-upload step fails cannot succeed
Observed on the v0.26.0 release (run 31065438598). The endpoint-verify step failed on first run because it executed minutes after the PyPI upload, inside the version-check worker's 15-minute PyPI cache window. Rerunning the failed job then re-executed the entire build-and-publish job, and the PyPI upload step failed with `400 Bad Request from https://upload.pypi.org/legacy/` - the artifacts already exist, so a rerun can never turn the run green.

Two fixes, both worthwhile:
1. The PyPI publish step should tolerate already-uploaded artifacts (`skip-existing` on pypa/gh-action-pypi-publish), so post-upload verification steps are rerunnable.
2. The endpoint verify step should retry with backoff for up to ~20 minutes before failing, since check.brigade.tools self-updates from PyPI on a 15-minute cache and a fresh publish is always inside that window.

Release impact of the v0.26.0 instance: none - PyPI serves 0.26.0, the GitHub release is Latest with all 27 assets, the endpoint now reports 0.26.0 (verified with scripts/verify_version_check_endpoint.py, exit 0), and a live `brigade update --channel stable` succeeded. The workflow run just cannot be made green retroactively.

solomonneas:
Fixes #749.

The v0.26.0 publish run went red twice for process reasons while every artifact shipped correctly: the endpoint verify step ran inside the version-check worker's 15-minute PyPI cache and failed on the stale value, and the rerun then re-uploaded to PyPI, which returns `400 File already exists` - so the run could never be greened.

- `pypa/gh-action-pypi-publish` gains `skip-existing: true`, making reruns idempotent past the upload.
- `scripts/verify_version_check_endpoint.py` gains `--wait-seconds`/`--poll-interval` polling; the workflow passes 1500s so the verify rides out the cache window instead of failing on the first read. Smoke-tested live against the real endpoint: matching version exits 0 immediately, mismatched version polls the window then exits 1 with the corrected remediation text.

Gate: full `./scripts/verify` green in the lane worktree (Brigade receipt); workflow YAML strict-validated (duplicate-key-safe); `test_publish_workflow.py` assertions updated.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -406,10 +406,12 @@ jobs:
         run: python -m build
       - name: Publish to PyPI
         uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b
+        with:
+          skip-existing: true
       - name: Verify check.brigade.tools reports the published version
         env:
           EXPECTED_VERSION: ${{ github.ref_name }}
-        run: python scripts/verify_version_check_endpoint.py --expected-version "${EXPECTED_VERSION#v}"
+        run: python scripts/verify_version_check_endpoint.py --expected-version "${EXPECTED_VERSION#v}" --wait-seconds 1500
 
   published-artifact-acceptance:
     name: published-artifact-acceptance (${{ matrix.platform }})
--- a/scripts/verify_version_check_endpoint.py
+++ b/scripts/verify_version_check_endpoint.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 import argparse
 import json
 import sys
+import time
 import urllib.request
 
 ENDPOINT = "https://check.brigade.tools/v1/version"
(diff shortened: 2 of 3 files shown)


solomonneas:
## Merge shepherd full review (pre-touch)

Reviewed against current main after a clean rebase:

- **Intent:** make publish reruns idempotent (`skip-existing: true` on `pypa/gh-action-pypi-publish`) and ride out the 15-minute check.brigade.tools PyPI cache via `--wait-seconds 1500` polling.
- **Diff size:** 3 files, focused. Workflow test updated to assert both knobs.
- **Local verify:** `tests/test_publish_workflow.py` passed after rebase.
- **Risk notes:** `skip-existing` means a corrupted prior upload of the same filename would not be overwritten (acceptable for PyPI immutability). Polling can hold the publish job up to ~25 minutes on a stuck worker — matches the stated cache window.
- **Prior tip had no fresh required checks** (only stale CodeRabbit) because the branch was 139 commits behind; force-pushed rebased tip to refresh CI.

Will merge only after required checks go green on this tip.

PR #773, opened 2026-08-09
fix(handoff): classify salvageable notes distinctly in lint

### LINKED ISSUE #680: handoff lint: a well-formed note with non-Brigade headings is indistinguishable from garbage
## Problem

`brigade handoff lint` treats a well-structured note that uses non-Brigade headings exactly the same as unstructured garbage: same errors, same exit code, empty `hints` and `warnings`. There is no signal that the note is salvageable and no pointer at the expected format.

Worse, `brigade ingest --dry-run` already does better on the identical file. The two tools disagree about the same input.

## Reproduction

Verified 2026-08-03 on `main` (ad5ae46f) against an isolated `mktemp -d` target.

Two files in the inbox. `homegrown.md` is a real, well-formed engineering note using its own headings (`## What happened`, `## Why it matters`, `## Decisions`, `## Next steps`). `garbage.md` is `asdf qwer zxcv 12345 lorem ipsum dolor`.

`brigade handoff lint` on each:

```
[fail] .../homegrown.md
  - missing required section: Type
  - missing required section: Title
  - missing required section: Summary
  - missing required section: Recommended memory action

[fail] .../garbage.md
  - missing required section: Type
  - missing required section: Title
  - missing required section: Summary
  - missing required section: Recommended memory action
```

Byte-identical, both exit 1. The JSON for `homegrown.md` confirms nothing distinguishes it:

```json
{"action": null, "errors": ["missing required section: Type", "..."],
 "hints": [], "warnings": [], "valid": false}
```

`hints` and `warnings` are both empty, even though the schema already carries them.

`brigade ingest --dry-run` on the same inbox:

```
[dry-run] inbox -> ...-garbage.md    (garbage.md; missing `Recommended memory action`)
[dry-run] inbox -> ...-homegrown.md  (homegrown.md; unknown sections present (parser may have split content):
                                      ['what happened', 'why it matters', 'decisions', 'next steps'])
```

Ingest enumerates the note's real headings and says the parser may have split content. Lint, which is the tool whose entire job is to explain what is wrong with a handoff, says less than the tool that is only previewing an ingest.

## Why it matters

This is the first-contact path for anyone bringing existing notes into Brigade. A migrating user gets a wall of "missing required section" with no indication that their file was 90% of the way there, no list of the headings Brigade did find, and no hint about the expected section names. The natural read is "my note is garbage", when in fact only the heading vocabulary differs.

## Proposal

Lint already has the machinery; it just is not using it on this path.

- When a file parses into sections but none of them match the required set, emit a `hint` rather than only errors. Include the unknown headings that were found, the way ingest already does.
- Distinguish "structured note, wrong heading vocabulary" from "no recoverable structure" in the printed output, so the two cases stop looking identical.
- Add a migration pointer to the expected sections (`Type`, `Title`, `Summary`, `Recommended memory action`) in the hint.
- Keep `valid: false` and the non-zero exit. This is about diagnostics, not about loosening the contract.
- Reuse the section-detection ingest already performs so the two tools cannot drift apart again.

## Acceptance criteria

- [ ] A well-formed note with non-Brigade headings produces at least one `hint` naming the unknown sections found.
- [ ] Its lint output is visibly different from that of an unstructured file.
- [ ] The required-section errors, `valid: false`, and the non-zero exit code are unchanged.
- [ ] Lint and `ingest --dry-run` report the same unknown-section set for the same file.
- [ ] Tests cover the structured-but-foreign-headings case and the no-structure case.
- [ ] `./scripts/verify` passes.

## Provenance

Cold-start test ledger report `20260610-112516-task-handoff-lint-demands-exact-brigade-section-headi-c38786`, re-verified live 2026-08-03. Three of the four reports in that batch turned out to be already fixed; this one is real.

solomonneas:
Lint now uses the same structural boundary as ingest (has_salvageable_structure), so a real engineering note with non-Brigade headings gets hints and a salvageable signal instead of being indistinguishable from garbage. Diagnostics-only: hard-failure exit codes unchanged.

Verified: tests/test_handoff_cmd.py green locally (verify receipt 20260809-015400-work-verify-1b3ebc).

Origin: Codex Cloud task applied locally and reviewed.

Fixes #680

### DIFF
--- a/src/brigade/handoff_cmd/linting.py
+++ b/src/brigade/handoff_cmd/linting.py
@@ -270,6 +270,7 @@ def lint_file(path: Path) -> HandoffLintResult:
     warnings: list[str] = []
     hints: list[str] = []
     action: str | None = None
+    salvageable = False
     try:
         text = path.read_text(errors="replace")
     except OSError as exc:
@@ -282,6 +283,11 @@ def lint_file(path: Path) -> HandoffLintResult:
         )
 
     raw_sections = _parse_markdown_sections(text)
+    # Use the same structural boundary as ingest: a note with several populated
+    # Markdown sections can be preserved for review, unlike unstructured text.
+    from ..ingest import has_salvageable_structure
+
+    salvageable = has_salvageable_structure(raw_sections)
     sections, collision_errors, noncanonical_headings = _canonicalize_sections(raw_sections)
     errors.extend(collision_errors)
     heading_warnings, heading_hints = _lint_noncanonical_heading_messages(noncanonical_headings)
@@ -301,6 +307,12 @@ def lint_file(path: Path) -> HandoffLintResult:
 
             if not scan_untrusted(text).flagged:
                 hints.append("this looks like a freeform note; try `brigade handoff migrate --target .`")
+        elif salvageable:
+            from ..untrusted import scan_untrusted
+
+            if not scan_untrusted(text).flagged:
+                expected = ", ".join(f"## {name}" for name in CANONICAL_HANDOFF_SECTION_HEADINGS)
+                hints.append(f"structured note detected; use Brigade handoff sections: {expected}")
 
     action_value = _section_value(sections, "Recommended memory action")
     if action_value:
@@ -324,6 +336,7 @@ def lint_file(path: Path) -> HandoffLintResult:
         warnings=tuple(warnings),
         hints=tuple(hints),
         readability=readability,
+        salvageable=salvageable,
     )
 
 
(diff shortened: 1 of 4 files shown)


solomonneas:
Independent review approved the current head with no Critical or Important findings. I attempted a follow-up for four nonblocking cleanup items, but the active pre-push guard blocked on existing intentional secret fixtures outside this PR. The new commit's history scan was clean. I did not bypass the hook or add unrelated allow-tags.

Merging the reviewed, green head. The cleanup can return in a guard-scoped follow-up.

PR #793, opened 2026-08-09
fix(security): coalesce overlapping scan findings

### LINKED ISSUE #705: Security scan: coalescing is not suppression-aware (follow-up to #696)
## Symptom

After the #696 fix lands, a line carrying two real credentials emits one finding. If a user had already suppressed the higher-ranked title on that line, the coalescing keeps that suppressed title and drops the unsuppressed one, so the second real credential goes silent.

## Evidence

The #696 patch is not on main. `SECRETS_TITLE_RANK` does not exist at HEAD:

```text
$ grep -rn "SECRETS_TITLE_RANK" src/
(no output)
```

The behavior described below refers to the proposed patch for #696 (325 lines across `src/brigade/security_cmd/models.py`, `src/brigade/security_cmd/scan_engine.py`, and `tests/test_security_cmd.py`), not to files on main.

Review text on the patch:

```text
if a user had previously suppressed ONLY the `Possible hardcoded credential` finding (accepted that risk), the baseline still showed the `Possible sensitive secret material` finding for the second credential; the patched coalescing keeps the higher-ranked title (rank 20 > 10), which is exactly the suppressed one - so the second, possibly real, credential goes silent.
```

The alias machinery to reuse does exist at HEAD:

```python
def _safe_fingerprint_aliases(finding: dict[str, Any]) -> tuple[str, ...]:
```

`src/brigade/security_cmd/scan_engine.py:939`

## Proposed fix

In the #696 coalescing path in `src/brigade/security_cmd/scan_engine.py`, when coalescing drops lower-ranked titles, record their would-be fingerprints as aliases on the surviving finding via `_safe_fingerprint_aliases`. Alternatively, prefer an unsuppressed candidate over a suppressed higher-ranked one when choosing which finding survives.

Consult the suppression baseline lookup so a suppressed higher-ranked title cannot hide an unsuppressed lower-ranked finding on the same line.

## Suggested milestone

0.27.0

solomonneas:
## Summary

- coalesce overlapping security findings before suppression matching
- choose the representative finding by severity and stable title rank
- prevent internal coalescing metadata from reaching JSON, SARIF, or evidence bundles

Fixes #705

## Verification

- `pytest -q tests/test_security_cmd.py`
- `ruff check src/brigade/security_cmd/models.py src/brigade/security_cmd/scan_engine.py tests/test_security_cmd.py`

### DIFF
--- a/src/brigade/security_cmd/models.py
+++ b/src/brigade/security_cmd/models.py
@@ -250,6 +250,45 @@ def _secrets_title_rank(title: str) -> int:
     return SECRETS_TITLE_RANK.get(title, 0)
 
 
+_INTERNAL_FINDING_KEYS = frozenset({"_coalesce_group", "_fingerprint_content"})
+
+
+def _strip_internal_finding_keys(finding: dict[str, Any]) -> dict[str, Any]:
+    for key in _INTERNAL_FINDING_KEYS:
+        finding.pop(key, None)
+    return finding
+
+
+def _finding_coalesce_rank(finding: dict[str, Any]) -> tuple[int, int]:
+    """Rank coalesce-group members: severity first, then secrets title tie-break.
+
+    Scanner-defined coalesce groups currently emit uniform severity; title rank
+    breaks ties within the same severity so the most specific secret label wins.
+    """
+    severity = SEVERITY_ORDER.get(str(finding.get("severity") or ""), 0)
+    return severity, _secrets_title_rank(str(finding.get("title") or ""))
+
+
+def _coalesce_findings(findings: list[dict[str, Any]]) -> list[dict[str, Any]]:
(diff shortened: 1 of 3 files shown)


solomonneas:
Rebased onto main and resolved the test_security_cmd.py conflicts with the merged #792 title-rank/fingerprint pins (both test sets kept). Not merging: test_phase257_production_family_files_stay_under_2000_lines still fails deterministically - src/brigade/security_cmd/scan_engine.py lands at 2004 lines against the 2000-line budget. Needs the file trimmed or split (a real refactor, out of scope for the landing pass). Everything else about the coalescing change looked consistent with the #792 pins (severity-then-title rank, one surviving finding per group).

Issue #845, opened 2026-08-10
memory eval: measure stale recall, projection drift, and scope leakage


solomonneas:
## Summary

Extend the shipped #722 retrieval harness with integration tests for the memory-card projection. Measure projection drift, stale or harmful recall, scope leakage, provenance completeness, duplicate live rows, instruction-like content handling, and index/query cost.

This issue reuses #722's adapters and reporting. It does not add Hindsight or change the semantic-retrieval gate.

## Problem

#722 measures keyword, grep, and optional semantic retrieval over a fixed card corpus. It does not test failure modes introduced by a derived evidence projection: partial scans, removed or superseded cards, cross-scope results, missing provenance, duplicate live versions, or quarantined content entering a default brief.

Cross-source memory search should not ship without tests that compare it to canonical Markdown and verified source evidence.

## Scope

Reuse #722's fixture layout, adapter pattern, JSON output, and table output. Add fixtures and measures for:

| Dimension | Required check |
| --- | --- |
| Projection fidelity | Canonical Markdown identity and hash compared with the live projection after create, edit, rename, and removal |
| Stale and partial state | Stale snapshot, interrupted scan, and removed-after-complete-scan cards excluded or visibly marked in default results |
| Harmful recall | Superseded or contradicted card cannot outrank current canonical state without a conflict signal |
| Scope leakage | Repository, task, operator, branch, and worktree boundaries enforced by closed scope fields |
| Provenance | Every returned item carries source identity, trust state, and available receipt linkage |
| Duplicates | One live result per external ID even when older content-hash versions remain stored |
| Instruction-like content | Quarantined or injection-pending cards cannot enter the default trusted path |
| Cost | Index build time, query p50 and p95, and index size for the checked-in corpus |

Use these arms where available:

1. current keyword memory search
2. #722 grep baseline
3. MiseLedger FTS after #843 and #844
4. context-pack selection where the same fixture applies
5. #722 optional semantic adapter without changing its gate

## Acceptance criteria

- [ ] The integration suite runs offline without a model download for all non-semantic arms.
- [ ] One command produces diffable JSON and table output by arm and failure category.
- [ ] Fixtures and expected results are checked in and reproducible.
- [ ] Removed, stale, partial, superseded, and duplicate states have separate assertions.
- [ ] Scope-leakage fixtures cover repository, task, operator, branch, and worktree boundaries.
- [ ] Every projected result is checked for source identity, trust, and provenance completeness.
[...]

Issue #844, opened 2026-08-10
memory: add evidence projection facade, health, and read-only identity audit


solomonneas:
## Summary

Add the Brigade-facing route for the native memory-card projection from #843. Expose manual crawl, version gating, receipts, health, generic explanation, projection-only rebuild, and a read-only identity audit.

File-backed Markdown memory remains usable when MiseLedger is missing, outdated, or unhealthy.

## Problem

The engine slice owns import and reconciliation. Brigade still needs an operator-facing boundary that checks engine capability, records scan outcomes, reports stale or partial state, and audits legacy identity before any reviewed backfill.

The facade must not make the derived archive a prerequisite for reading, editing, or searching canonical cards.

## Scope

- Register `memory` as a native evidence source in Brigade.
- Add `brigade evidence crawl memory <workspace>` and delegate to the engine source.
- Require a minimum compatible MiseLedger version and capability. Refuse unsupported versions before import starts.
- Surface scan ID and created, updated, unchanged, removed, skipped, and failed counts in the Brigade receipt.
- Show engine capability, last completed scan, canonical and live counts, hash divergence, unresolved relations, malformed or skipped cards, and stale or partial state through doctor or status.
- Support projected cards through the generic evidence show or explain surface. Add memory-specific rendering only if the generic contract cannot show the required safe metadata.
- Add a projection-only rebuild that leaves unrelated evidence sources untouched.
- Add a read-only identity coverage and collision audit for existing cards.
- Document path-derived fallback identity and its rename behavior.
- Share frontmatter, rename, failure, and compatibility fixtures with the engine slice.
- Preserve `brigade memory search` as the deterministic fallback.

## Acceptance criteria

- [ ] `brigade evidence crawl memory <fixture>` delegates only when the engine satisfies the minimum version and capability contract.
- [ ] An unsupported engine fails before any partial import.
- [ ] The Brigade receipt includes the scan ID and all six result counts.
- [ ] Doctor or status exposes the required projection health fields without printing card bodies.
- [ ] Generic evidence inspection shows projected-card identity, source, hashes, trust, and declared relations.
- [ ] Projection-only rebuild reproduces the engine fixture's live IDs and hashes without touching unrelated evidence.
- [ ] The identity audit reports explicit-ID coverage, fallback coverage, and collisions without changing cards.
- [ ] Missing or unhealthy MiseLedger does not block file-backed memory operations.
- [ ] A failed or partial crawl cannot report healthy or current state.

## Non-goals

[...]

Issue #843, opened 2026-08-10
evidence-ledger: project canonical memory cards with stable identity and safe reconciliation


solomonneas:
## Summary

Implement the engine-only slice approved in #833. Add a native MiseLedger memory source that projects canonical Markdown cards into `miseledger.adapter.v1` with stable external identity, the existing provenance envelope, qualified cross-source relations, completed-scan reconciliation, projection health, and source-local rebuild.

Markdown remains canonical. This issue does not add a Brigade facade, mutate cards, or change memory search.

## Problem

Generic document crawling loses card identity, frontmatter semantics, explicit receipt relations, and removal state. MiseLedger also cannot currently resolve a card relation to a receipt imported under another source, and its source-scan records do not provide a completed-scan reconciliation path for removed cards.

The resulting contract must distinguish a completed removal from an interrupted scan. A partial scan cannot tombstone live cards or report a healthy current projection.

The operator decisions recorded on #833 control this slice:

- new or reviewed cards use opaque `card-<uuid4>` IDs
- legacy cards remain readable through a normalized workspace-relative-path fallback with `identity_source`
- relation targets are qualified by source, collection, and external ID
- the engine lands before any Brigade facade
- Markdown remains the only memory authority

### Native memory source

- Add `miseledger crawl memory <path>`.
- Define a versioned memory-card source record and a bounded flat-frontmatter subset shared with Brigade fixtures.
- Set `item.external_id` from the explicit card ID or the documented path fallback.
- Record `identity_source` so callers can distinguish explicit and fallback identity.
- Preserve unknown frontmatter in bounded metadata. Malformed input produces a per-file failure, not a partial valid record.

### Provenance and identity

- Reuse the existing provenance envelope for captured and ingested times, raw and content hashes, source identity, trust, and injection state.
- Reuse #724's normalized content fingerprint for duplicate detection only. It is never card identity.
- Define live identity as `(source, collection, external_id)` with content-hash versions. Search and health treat only the latest valid version as live.
- Do not introduce a memory-specific timestamp, hash, trust, or provenance envelope.

### Relations and reconciliation

- Add a backward-compatible qualified relation target containing source, collection, and external ID.
- Resolve qualified cross-source references to Brigade receipts when the target exists.
- Keep unresolved references queryable instead of dropping them.
- Write a completed-scan manifest for the memory source.
- Permit source-local soft tombstones only after a completed scan proves absence.
[...]

Issue #498, opened 2026-07-24
evidence ingestion: origin-scoped redaction policy with counted, non-stored redactions


solomonneas:
> **Contributor summary:** Redact sensitive values at evidence-ingestion time by source origin, persist only redaction counts (not the removed values), and keep the audit trail honest.

## What
Add an evidence-ingestion policy that classifies sources, applies origin-scoped redaction before persistence, and records redaction counts without storing the removed values.

## Why (evidence)
Pattern from AgentPantry, which uses allowlisted origins, explicit consent, and source-to-sink controls for sensitive browser state. Brigade evidence currently persists whatever the capture path saw; ingestion is the correct (and only reliable) boundary for redaction, and counting redactions keeps the audit trail honest about what was removed.

Related literature: provenance-aware boundary validation for agent evidence (Syed et al., ICCA 2025, doi:10.1109/icca66035.2025.11430791).

Source: adjacent-project research sweep (local sibling analysis), brigade run 20260724-032852-93c729b8.

## Effort
M

solomonneas:
Landing queue update: Codex Cloud task task_e_6a78a875128083278ffe51ec532c44ab reached READY with no diff or tests. Metadata and diff were checked directly, so this item is skipped rather than assigning implementation claims to an empty artifact. Requeue it with a nonzero, issue-scoped diff.

solomonneas:
## Scope amendment: policy versions apply to future writes

Clarify the persistence boundary when an origin-scoped redaction policy is enabled or tightened:

- Every detector or policy decision records an explicit policy version.
- A new policy applies to future persistence only.
- Existing evidence rows are not silently rewritten when policy changes.
- Redaction counts and scanner state are retained without storing removed values.
- Scanner failure, timeout, or unavailability cannot produce a clean verdict.

Any historical scan is a separate operation. Its contract must be:

- explicit and operator-initiated
- resumable from a cursor or checkpoint
- idempotent across retries
- non-destructive toward canonical source material
- separately receipted with scanned, redacted, skipped, and failed counts plus the policy version
- visibly partial or failed in health output when it does not complete

This amendment does not implement a historical rescan worker in 0.27.0. The ingest boundary remains the primary control. A later rescan cannot promote trust or erase source history merely because a newer detector did not flag an item.

#498 remains a prerequisite before any future automatic transcript or external-content retention. It does not block the manual canonical-card projection derived from #833.

PR #850, opened 2026-08-10
feat(evidence-ledger): memory projection E1 namespace and failure-preserving rebuild

solomonneas:
Refs #843

## Summary

Engine-only E1 memory projection for MiseLedger (`engines/evidence-ledger`): namespace-scoped crawl/rebuild, failure-preserving detach/remap, health dual-read, and latest-version / known-item contracts.

## Maintainer repairs

Consolidated review scope through comment `5247021742`:

1. Latest-version ordering uses database-monotonic `items.ingest_seq` (schema v3), not wall-clock `updated_at` or content-hash ID order. Equal and backward clock regressions are included.
2. v2-to-v3 migration backfill deterministically assigns `ingest_seq` from prior `updated_at` order with stable ID fallback, so multi-version cards keep the prior live winner before any crawl. The upgrade regression covers live projection, relations, and health.
3. The same-text frontmatter known-item path advances `ingest_seq` and reconciles metadata, tags, provenance, artifacts, and outbound relations without minting duplicate items/events. Direct adapter already-known re-resolves inbound edges.

Prior E1 fixes retained: empty-namespace health aggregation, detach/remap rebuild preservation, UUID v4 and variant, truthful legacy `memory:cards` docs, inbound repoint on edit, UTF-8/idempotency/tombstone/namespace contracts.

## Out of scope

E2, #844, #845, and #498 are not entered.

### DIFF
--- a/engines/evidence-ledger/CHANGELOG.md
+++ b/engines/evidence-ledger/CHANGELOG.md
@@ -24,8 +24,9 @@ Releases before this changelog was started are on the [releases page](https://gi
   artifacts, post-resolution `unresolved_relations`, #724 content fingerprint
   for duplicate detection only, duplicate explicit-id fail-closed, empty-
   namespace status/doctor dual-read across all `brigade-memory` collections,
-  and legacy `memory:cards` scoped-rebuild (never tombstoned by namespaced
-  crawl).
+  legacy `memory:cards` scoped-rebuild (never tombstoned by namespaced
+  crawl), and latest-version selection for relation resolution plus live
+  health after content-addressed card edits.
 
 ## [0.6.0] - 2026-07-18
 
--- a/engines/evidence-ledger/CHANGELOG.md
+++ b/engines/evidence-ledger/CHANGELOG.md
@@ -17,12 +17,15 @@ Releases before this changelog was started are on the [releases page](https://gi
 
 ### Changed
 
-- Memory projection E1 hardening (#843): operator-declared `memory-<uuid4>`
-  namespace from `memory/NAMESPACE`, namespace-scoped collection/scan/rebuild,
-  failure-preserving rebuild (walk/stage before replace), post-resolution
-  `unresolved_relations`, #724 content fingerprint for duplicate detection only,
-  duplicate explicit-id fail-closed, and legacy `memory:cards` scoped-rebuild
-  dual-read rule.
+- Memory projection E1 hardening (#843): operator-declared RFC 4122
+  `memory-<uuid4>` (version 4 + variant bits) from `memory/NAMESPACE`,
+  namespace-scoped collection/scan/rebuild, detach/remap failure-preserving
+  rebuild that restores prior live IDs/hashes/relations/metadata/events/
+  artifacts, post-resolution `unresolved_relations`, #724 content fingerprint
+  for duplicate detection only, duplicate explicit-id fail-closed, empty-
+  namespace status/doctor dual-read across all `brigade-memory` collections,
+  and legacy `memory:cards` scoped-rebuild (never tombstoned by namespaced
+  crawl).
 
 ## [0.6.0] - 2026-07-18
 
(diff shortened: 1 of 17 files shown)


solomonneas:
Bounded repair request from independent Terra review at `bd1be7e`: a normal non-`--rebuild` edited card receives a new content-addressed item ID (`internal/ingest/importer.go:493`), but relation resolution updates only rows with `target_item_id IS NULL` (`:318-320`). The prior version stays untombstoned because the external ID remains present (`internal/ingest/memory.go:318-321`), leaving stale inbound links and potentially stale unresolved-relation health. Please make one focused repair: add a non-rebuild edit lifecycle regression test, and make relation/live-health selection use the latest version for `(source, collection, external_id)`. Keep scope engine-only, do not broaden E1, then report the new head and fresh focused/full verification.

solomonneas:
Fresh bounded repair request from Terra review at `9bda3a57`:

1. The new `ORDER BY i.created_at DESC, i.id DESC` is not chronological because memory `CreatedAt` is persisted empty (`internal/sources/memory/memory.go:415-421`, `internal/ingest/importer.go:583,590`). Tied rows choose lexicographically greatest content hash, not the edited version. Introduce a monotonic ingestion/version timestamp or explicit latest marker, and assert the edited content/hash is selected.
2. The added health test uses only an inbound receipt relation, which memory health never counts. Add v1 with an unresolved outbound card relation, edit v2 to remove or resolve it, and assert health counts only latest-version outgoing unresolved relations.

Keep this engine-only and limited to the lifecycle fix. Provider owner: Cursor bc-a7a66bc4-fe5b-4dc1-9c96-60a5e4b1cf8b.

solomonneas:
Exact-head Terra review found one remaining blocker at `ef0ade2e`:

The new `updated_at` ordering is correct for distinct content, but a re-ingest of an already known content-addressed item returns before writing the ingest stamp. Sequence: ingest v1, edit to v2, then restore byte-identical v1. The final ingest hits the known-item return and retains v1’s original stamp, so v2 remains selected as live. That leaves inbound relations projected onto v2 and excludes v1’s current outgoing unresolved relations from health.

Bounded repair: when a matching item is re-ingested, update its monotonic ingest/version stamp before returning, without minting a duplicate item or event. Add a `v1 -> v2 -> v1` regression that asserts v1 is live for projection and inbound relation resolution, and that unresolved-health counts use v1’s outgoing relations. The narrow targeted suite passed on this head, but it does not cover this sequence.

solomonneas:
Exact-head Terra review found one remaining direct-import blocker at `120adaa5`:

The known-item path now correctly restamps v1 on `v1 -> v2 -> identical v1`, but the matching completed source hash takes the `AlreadyKnown` return before `resolveRelations`. Live projection selects v1 by the refreshed stamp while existing inbound `relations.target_item_id` remains v2. The crawl regression passes because `CompleteMemoryScan` resolves relations separately; it does not cover direct adapter re-ingest.

Bounded repair: invoke `resolveRelations(tx)` before that known-source return, without creating an item or event, and add a direct-import inbound-repoint regression alongside the existing crawl test. Targeted tests pass on this head but miss this path.

solomonneas:
Exact-head Luna and Terra review found one consolidated versioning blocker set at `bc92cfc2`:

1. Live-version ordering uses raw wall-clock `updated_at` with an ID/hash tie-break. Equal or backward clock values can select the wrong version. Replace this ordering signal with a database-monotonic ingest/version sequence or an explicit latest marker, and add equal/backward-clock regressions.
2. The content item hash covers Text+Summary only. A frontmatter-only update returns through the known-item path, restamps the existing row, but never reconciles metadata, tags, provenance, artifacts, or relations. A receipt-A to receipt-B edit can therefore remain projected as receipt-A.

Bounded repair: define atomic canonical-field versioning or perform complete known-item reconciliation under the same monotonic version rule, then add a regression for a same-text frontmatter/provenance/relation change that proves the latest projection, relation target, and health data reflect receipt-B. Keep the earlier no-duplicate-event behavior and prior repaired cases intact. Do not split this into serial partial fixes.

solomonneas:
Exact-head Luna and Terra agree on one blocking migration defect at `f606d7fc`:

The `ingest_seq` migration gives every pre-v3 row a default of `0` and does not backfill it. Existing multiple memory versions therefore tie and live selectors fall back to content-hash ID ordering, which can revive an older version and misproject relations or health after upgrade.

Bounded repair: deterministically backfill `ingest_seq` during migration using the prior `updated_at` order with a stable fallback, and add a multi-version v2-to-v3 upgrade regression that verifies the latest version, relation projection, and health before any crawl/import occurs. All other reviewed behavior and Actions are clean.

PR #861, opened 2026-08-10
fix(release): authenticate published-artifact GitHub release lookup (#819)

### LINKED ISSUE #819: release: published-artifact-acceptance API call needs auth and backoff (darwin-arm64 403s)
The v0.26.1 publish workflow (run 31326623469) concluded failure solely because published-artifact-acceptance (darwin-arm64) hit HTTP 403 from an unauthenticated urllib call to api.github.com releases/tags - reproduced identically across 3 reruns while the three sibling platform jobs passed at the same moment and a direct curl returned 200. GitHub API rate/abuse detection on the macOS runner pool, not a release defect (PyPI live, check.brigade.tools reported 0.26.1 ok, 27 assets gated). Fix: pass GITHUB_TOKEN auth on the API call in scripts/published-artifact-acceptance.py and add retry with backoff, so future releases do not need manual triage of a red workflow that actually succeeded. Release-agent handoff: .claude/memory-handoffs/20260809-0261-shipped.md

solomonneas:
## What and why

The v0.26.1 publish workflow failed only on `published-artifact-acceptance (darwin-arm64)` because `brigade setup` hit the unauthenticated GitHub `releases/tags` API and received HTTP 403 from runner rate/abuse detection. This change adds an authenticated release-tag preflight with bounded retry on transient 403/429 responses, drives managed setup through the already verified release manifest, wires `GITHUB_TOKEN: ${{ github.token }}` into the Unix acceptance workflow step, and blocks cross-origin urllib redirects so `Authorization` cannot leak to another host.

Closes #819

## Type of change

- [x] Bug fix
- [ ] Docs

## Checklist

- [x] `pytest -q` passes locally (`./scripts/verify`: 6707 passed, 5 skipped; 83.10% coverage)
- [ ] Updated the `Unreleased` section of `CHANGELOG.md` for any user-visible effect (entries describe effects, not commit subjects)
- [x] Conventional commit messages; in-house Cursor co-author trailers on substantial coding commits

## Verification

- `pytest -q tests/test_published_artifact_acceptance.py tests/test_publish_workflow.py::test_published_artifact_acceptance_unix_step_passes_github_token_from_github_token` (47 passed)
- `./scripts/verify` (full gate)
- `brigade work verify run --target . --command "pytest -q tests/test_published_artifact_acceptance.py tests/test_publish_workflow.py::test_published_artifact_acceptance_unix_step_passes_github_token_from_github_token" --capture brigade-work`
- `brigade handoff lint .claude/memory-handoffs/20260810-2100-issue-819-published-artifact-github-auth.md`

## Notes

- Scoped to `scripts/published-artifact-acceptance.py`, `tests/test_published_artifact_acceptance.py`, and the Unix published-artifact acceptance step in `.github/workflows/publish.yml`; PR #750 behavior preserved.
- `GITHUB_TOKEN` is attached as `Authorization: Bearer` when present, never logged, and not forwarded on cross-origin redirects.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -453,6 +453,7 @@ jobs:
         if: matrix.kind == 'unix'
         env:
           BRIGADE_VERSION: ${{ github.ref_name }}
+          GITHUB_TOKEN: ${{ github.token }}
         run: python scripts/published-artifact-acceptance.py --brigade-version "${BRIGADE_VERSION#v}" ${{ matrix.extra }}
       - name: Run Windows native acceptance (published CLI)
         if: matrix.kind == 'windows'
--- a/scripts/published-artifact-acceptance.py
+++ b/scripts/published-artifact-acceptance.py
@@ -14,6 +14,7 @@ import subprocess
 import sys
 import tempfile
 import time
+import urllib.error
 import urllib.request
 from pathlib import Path
 from typing import Any, Callable, Mapping, Sequence
@@ -25,6 +26,10 @@ REPOSITORY = "escoffier-labs/brigade"
 PYPI_PROJECT_URL = "https://pypi.org/pypi/brigade-cli/json"
 PYPI_AVAILABILITY_TIMEOUT_SECONDS = 6 * 60
 PYPI_POLL_INTERVAL_SECONDS = 5
+USER_AGENT = "brigade-published-artifact-acceptance/1.0"
+GITHUB_RELEASE_TAG_MAX_ATTEMPTS = 6
+GITHUB_RELEASE_TAG_INITIAL_BACKOFF_SECONDS = 2.0
+GITHUB_RELEASE_TAG_MAX_BACKOFF_SECONDS = 30.0
 # agent-notify ldflags inject main.version, main.commit, and main.buildDate.
 # A bare `go build` leaves "dev" / "unknown" / "unknown"; published release
 # assets must report the exact Brigade release version, a hex git SHA (the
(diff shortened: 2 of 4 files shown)


solomonneas:
Exact-head Terra review found one important workflow-contract blocker at `78c69286`:

`published-artifact-acceptance.py` adds `Authorization` only when `GITHUB_TOKEN` is present, but the target Darwin publish workflow step provides only `BRIGADE_VERSION`. `github.token` is available through the Actions context, not as a documented default runner environment variable, so the preflight remains unauthenticated and can reproduce #819.

Bounded repair: pass `GITHUB_TOKEN: ${{ github.token }}` to that exact workflow step and add a workflow-contract regression that prevents the token plumbing from being removed. Keep the implementation and scope otherwise unchanged.

solomonneas:
Consolidated Luna and Terra review on exact head 47f1bbfd found one security blocker. Python urllib follows redirects and preserves the Authorization header, so a redirected GitHub API request can send GITHUB_TOKEN to another origin. Use a no-redirect opener or allow redirects only when the final request remains HTTPS on api.github.com, and add a cross-origin redirect regression proving the token is not forwarded.

The workflow token wiring and bounded 403/429 backoff are otherwise correct. Clean the PR body by removing Cursor markers/footer and correcting its false statement that no publish workflow changed. The in-house Cursor co-author trailer is permitted; remove any PR-body claim that no such trailer exists rather than rewriting commit history. Preserve #750 behavior and re-run the focused workflow and acceptance tests.

PR #863, opened 2026-08-10
fix(onboarding): track managed Claude handoff template

solomonneas:
## What and why

#858 added repo-local recovery when a global Git exclude hides the managed Claude handoff template. The remaining quickstart guidance still described `.claude/` as entirely local, which could leave `.claude/memory-handoffs/TEMPLATE.md` untracked while Claude onboarding expects it in the repository.

This PR updates the documentation and managed installer rules together. When Claude Code is selected, the generated Git ignore block keeps session handoffs and settings local while un-ignoring the parent directories needed to track `.claude/memory-handoffs/TEMPLATE.md`. A normal `brigade init` rerun refreshes that managed block without overwriting existing `CLAUDE.md`, `AGENTS.md`, or user rules outside the markers. The operator-health wording and quickstart recording now describe the same exception.

The change remains repo-local. It does not edit global Git configuration or touch #750.

Closes #860

## Type of change

- [x] Bug fix
- [x] Documentation

## Verification

- Focused installer and operator tests: 21 passed
- Full `./scripts/verify`: 6737 passed, 5 skipped, 83.12% coverage
- Brigade verification receipt: `20260811-035232-work-verify-6c5495`
- Memory handoff content-guard lint: clean
- GitHub Actions: all required checks passed on `bf00f578`

## Checklist

- [x] Existing `CLAUDE.md` and `AGENTS.md` files survive a normal upgrade rerun
- [x] Parent un-ignore rules allow the managed template to be staged without `git add -f`
- [x] Session handoffs and settings remain ignored
- [x] The quickstart cast and generated SVG match the operator-health wording
- [x] `CHANGELOG.md` records the user-visible change
- [x] No new runtime dependencies
- [x] Existing commits truthfully retain their Cursor Agent and Solomon Neas co-author trailers

PR #862, opened 2026-08-10
docs(worker-events): document Codex notification envelope matrix (Refs #592)

solomonneas:
## What and why

Issue #592 first slice (fail-closed Codex app-server worker-event classification/scrubbing) already landed on main via #823, but the issue stayed open. This closeout completes the documented field-classification matrix for every recorded item type, adds adversarial golden fixtures for all 14 matrix item types, and maps acceptance criteria to enforcement points so the issue can close cleanly.

Terra review on `aa354fa4` (comment 5246656635) blocked one docs/contract issue: the docs presented a closed recorded-method list while `codex_appserver` forwards every non-delta notification and the writer records it raw. Head `c54da6ea` repairs that without narrowing the recorder.

Refs #592

## Type of change

- [x] Docs
- [x] Tests / fixtures

## Terra repair (bounded)

- Document raw Codex notification capture as open-ended for non-delta methods.
- Document that unknown methods remain local/`unclassified` and are rejected from scrub/export.
- Add `test_recorder_to_matrix_contract_keeps_raw_capture_open_ended` so docs cannot imply a closed recorder allowlist.
- Preserve the 14 item-type matrix, adversarial fixtures, fail-closed scrubber, media/artifact classes, legacy behavior, and docs-only/runtime boundary. Recorder path unchanged.

## Acceptance mapping (#592 first slice)

- Documented scrub matrix covers known methods/item types (`docs/worker-event-streams.md` + `classification_matrix()`).
- Raw capture is open-ended for non-delta methods; matrix is scrub/export-only (docs + recorder-to-matrix contract test).
- Unknown methods/fields fail closed with bounded diagnostics.
- Credentials, headers, cookies, env values, raw prompts, private retrieved content, absolute home paths, and provider error bodies absent from scrubbed projections.
- Scrubbed projections preserve order, stable IDs, safe operation names, schema versions, and source digests.
- Distinct raw/scrubbed media types and artifact classes.
- Audit/replay reject raw unless `policy=local-only`.
- Legacy streams remain inspectable and unclassified until scrubbed.

## Out of scope

- Legacy backfill of historical streams
- Export/archive scrubbed-sidecar wiring
- Narrowing the working recorder allowlist
- Memory/provenance surfaces (#850/#851) and publish workflow (#750)

## Verification

Focused:

```console
$ brigade work verify run --target /workspace \
    --command "ruff check src/brigade/worker_events.py tests/test_worker_events.py" \
    --command "ruff format --check ..." \
[...]

PR #864, opened 2026-08-10
feat(run): enforce wall-clock and dispatch run budgets (#593)

solomonneas:
## What and why

First enforceable slice of #593: the coordinator reserves **wall-clock** and **worker-dispatch** ceilings before new worker work starts, appends durable idempotent `run_budget.*` lifecycle events, and rebuilds remaining allocation from the journal on restart.

**Taxonomy decision (locked):** budget exhaustion and operator cancellation are terminal **policy lifecycle states** (`budget-exhausted` / `operator-cancelled`), not #576 worker `FailureClass` values and not infrastructure-neutral under #580. Provider/seat `capacity-exhausted` remains infrastructure.

Child frame / durable child-run allocation stays out of scope (#594).

**Repairs:**
- Parallel reservations + stable `dispatch:{seat}:{attempt}` identity + live Ctrl-C → `operator-cancelled` (comment 5247794372)
- Aggregate `token_budget` preserved, not aliased to `input_tokens` (comment 5248440501)
- Normal-run wiring, atomic same-seat attempt+reservation, truthful optional input/output schema/bridge (comment 5248588004)
- Declared-only hard ceilings documented; ordinary undeclared `brigade run` / dogfood / model-trial paths remain unbounded for backward compatibility—no invented flags, defaults, or timeout→budget mapping (comment 5249038705)
- Restart reclaim evaluates wall-clock before authorizing a durable pending dispatch so an expired run cannot relaunch attempt 1 (comment 5249411385)
- `brigade runs resume` gates on persisted declaration + original `started_at` + authoritative budget lifecycle projection before `AppServer.start()`, refusing expired/exhausted declared runs without provider/synthesis calls (comment 5249824510)
- Persisted declared resumes fail closed on untrusted lifecycle journals (read/chain/partial-tail/compatibility) and malformed verification-budget ceiling fields (comment 5250113662)
- Four remaining recovery bypasses closed: missing journal for declared runs, incompatible/empty budget shapes, projected dispatch usage exhaustion without an exhausted event, and unknown verification-budget dimensions/schema versions (comment 5251293454)
- Post-parking hardening: strict persisted `run_budget` envelope validation (closed field set, non-bool integer schema versions), fail-closed initial verification-contract budget parsing, and projector preservation of `verification_contract` / `run_budget` across rewrites

Rebased onto main `ce8e7f1e` (#871 import-provenance merge); no file overlap with the branch, focused and full suites re-verified post-rebase.

## Type of change

- [x] New feature (run budget enforcement)
[...]

PR #866, opened 2026-08-11
feat(runs): fail-closed scrubbed worker-event export (#592)

### LINKED ISSUE #592: Worker event streams: field classification and fail-closed scrubbed projections
## Problem

Brigade's app-server transport preserves raw worker streams under `events/<worker>.jsonl`. Other current worker transports do not record this stream. The recorded files can contain provider notifications, model text, tool metadata, paths, and transport diagnostics.

Replay and audit work would make these streams more durable and easier to consume. They need an explicit classification and scrubbing contract before any feature treats them as fixtures or portable evidence.

## Proposal

Define and enforce a privacy boundary for worker event streams:

- classify each event field as public metadata, private content, secret, or prohibited.
- keep raw streams local and access-controlled by default.
- produce a bounded scrubbed projection for inspection, replay evidence, export, or attachment.
- store digests and classified artifact references in lifecycle events instead of copying raw payloads.
- fail closed when an unknown event type or field cannot be classified.

The scrubber must understand transport-specific envelopes without promising that every opaque provider payload is safe to retain.

## Acceptance criteria

- [ ] A documented field-classification matrix covers every event type currently recorded by the app-server transport.
- [ ] Unknown event types or fields fail scrub/export with bounded diagnostics.
- [ ] Credentials, authorization headers, cookies, environment values, raw prompts, retrieved private content, absolute home paths, and provider error bodies are absent from scrubbed projections.
- [ ] Scrubbed projections preserve event order, stable IDs, safe operation names, schema versions, and source digests.
- [ ] Raw streams and scrubbed projections have distinct media types and artifact classifications.
- [ ] Replay and audit consumers reject raw streams unless an explicit local-only policy allows them.
- [ ] Golden fixtures include adversarial secrets and private-content payloads for each supported transport.
- [ ] Legacy streams remain inspectable locally but are marked unclassified until successfully scrubbed.

## First slice

- Cover only the Codex app-server notification envelope, which is the only transport that currently records worker streams.
- Add the classification matrix and a fail-closed scrubber with adversarial golden fixtures.
- Keep legacy backfill and export integration out of the first slice.

## Related work

- #489 covers app-server lifecycle and resume semantics.
- #498 covers explicit run-artifact redaction policy.
- #568 owns the durable lifecycle journal and replayable projection.
- #595 depends on this boundary before worker streams become fixtures.

## Non-goals

- Make raw model transcripts public.
- Guarantee safe export of an unknown provider envelope.
- Replace the ProvenanceEnvelope trust and content classification contract.
- Treat redaction as a substitute for access control.

solomonneas:
## What and why

Issue #592 remained open after #823 and #862 because `brigade runs export` could still copy raw worker streams and classify them as public support data. This change closes that boundary for top-level and nested `events/**` paths. It also authenticates the reserved `events/lifecycle.jsonl` journal so raw JSON-RPC worker envelopes cannot pass as lifecycle records.

Closes #592

## Behavior

- Raw worker JSONL stays local and unclassified in the source run.
- Export replaces discovered streams with scrubbed sidecars.
- Nested event paths receive the same classification and privacy checks as top-level streams.
- `events/lifecycle.jsonl` must be empty or a valid `brigade.run_event.v1` hash chain.
- Unknown methods, malformed projections, unclassifiable nested streams, and lifecycle worker-envelope smuggling fail closed.
- Archive validation and import repeat the privacy checks before accepting copied data.
- The source run remains unchanged.

## Verification

- Focused archive and worker-event suite: 56 passed
- Exact-head Luna review: ready
- Exact-head Terra review: ready
- GitHub Actions on `8ebce963`: 23 successful checks, no failures or pending checks
- Rebased on current main `3e9278f9`, including #863

## Out of scope

- Historical stream backfill
- Protected #750 publish work
- Memory projection work

## Checklist

- [x] Added direct and nested export, validate, and import regressions
- [x] Added malicious and valid lifecycle-journal cases
- [x] Updated `CHANGELOG.md` and receipt documentation
- [x] Existing commits retain their Cursor co-author trailer

### DIFF
--- a/.grok/memory-handoffs/2026-08-11-lifecycle-journal-smuggle-866.md
+++ b/.grok/memory-handoffs/2026-08-11-lifecycle-journal-smuggle-866.md
@@ -1,0 +1,42 @@
+# Memory Handoff
+
+## Type
+
+bugfix
+
+## Title
+
+Lifecycle journal reserved-path worker-envelope smuggling (#866)
+
+## Summary
+
+Maintainer review on PR #866 found that reserved top-level `events/lifecycle.jsonl` was excluded from worker-stream discovery without authenticating its record shape, so a digest-valid archive could smuggle raw JSON-RPC worker notifications as public `role=journal`. Export/validate/import now authenticate `brigade.run_event.v1` journals and reject worker envelopes in that path.
+
+## Durable facts
+
+- Nested worker-stream recursive discovery remains in place
+- `assert_lifecycle_journal_authentic` rejects JSON-RPC worker envelopes in `events/lifecycle.jsonl`
+- Valid empty or `brigade.run_event.v1` lifecycle journals still export/validate/import
+- `_assert_export_privacy_tree` runs checkpoint, worker-stream, and lifecycle gates together
+
+## Evidence
+
+- files changed: `src/brigade/work_run_archive.py`, `tests/test_work_run_archive.py`, `docs/worker-event-streams.md`, `docs/receipt-schemas.md`, `CHANGELOG.md`
+- commands run: `pytest -q tests/test_work_run_archive.py tests/test_worker_events.py`; `brigade work verify run`; `./scripts/verify`
+- error strings: `events/lifecycle.jsonl carries a worker notification envelope; reserved journal path rejects raw JSON-RPC worker streams`
+
+## Recommended memory action
+
+no-card
+
+## Target document
+
+.learnings/ERRORS.md
+
+## Suggested document content
+
+### Lifecycle journal reserved-path smuggle
+
+Excluding `events/lifecycle.jsonl` from worker discovery is not enough. Archive
+export/validate/import must authenticate the lifecycle record shape or reject
+JSON-RPC worker envelopes before granting the public journal classification.
--- a/.grok/memory-handoffs/2026-08-11-nested-worker-event-export-866.md
+++ b/.grok/memory-handoffs/2026-08-11-nested-worker-event-export-866.md
@@ -1,0 +1,43 @@
+# Memory Handoff
+
+## Type
+
+bugfix
+
+## Title
+
+Nested worker-event JSONL export boundary repair (#866 / #592)
+
+## Summary
+
+Terra sendback on PR #866 showed `events/nested/coder.jsonl` bypassed fail-closed export and was classified public support. Discovery is now recursive under `events/`, nested streams scrub to sidecars or refuse export, and validate/import reject nested raw-stream smuggling.
+
+## Durable facts
+
+- `list_worker_event_streams` uses `rglob("*.jsonl")` and excludes only top-level `events/lifecycle.jsonl`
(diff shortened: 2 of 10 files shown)


solomonneas:
## Consolidated Terra sendback on `f48eb7f8`

A nested worker-event stream can bypass the fail-closed export boundary.

`events/nested/coder.jsonl` is not discovered by the direct stream scan. The archive classifier then treats the remaining JSONL file as public support. An exact-head reproduction exported an unknown worker method and a bearer-token marker without redaction.

Required correction:

1. Treat worker-like JSONL anywhere beneath `events/` as a worker-event stream candidate.
2. Scrub supported records and reject unknown methods, fields, malformed projections, or unclassifiable streams before archive creation.
3. Make archive validation and import reject a nested raw-stream smuggling attempt.
4. Add export and validate/import regressions for nested paths with secret, private-path, and unknown-method markers.
5. Remove the Cursor body markers and tracking footer.

Preserve source-run immutability, the existing scrubbed sidecar contract, safe event identity, and supported delta handling. Keep this within #592. PR #864 overlaps only documentation and changelog files, so retain current-main wording during any later rebase.

The completed Actions are green, but the existing smuggling regression covers only a direct `events/coder.jsonl` path.

solomonneas:
Fresh exact-head review found one remaining archive privacy bypass at `41ce36fc239c34a15ed805de31219eeb39db2aa7`.

The repair correctly catches nested worker streams, but the reserved top-level `events/lifecycle.jsonl` filename is excluded from worker-stream discovery and from both pre-copy and post-copy raw-stream assertions. A valid-digest archive can therefore manifest a raw JSON-RPC notification containing an Authorization bearer value and prompt body as public `role=journal`; validation accepts it and import copies it. The reproduction completed with `LIFECYCLE_SMUGGLE_IMPORTED` under Brigade receipt `20260811-030723-work-verify-4f96e1`.

Keep the lifecycle journal contract, but authenticate and parse its allowed record shape before granting the exemption, or reject raw JSON-RPC worker envelopes in that reserved file. Add direct `validate_archive` and `import_archive` regressions for the reserved filename, including secret-bearing JSON-RPC content and a valid lifecycle record. Preserve the nested-stream repair, rebase only if main moved, run focused archive/worker-event tests and the full gate through Brigade, push to this same branch with lease, and keep the PR draft. No duplicate PR.

solomonneas:
Both independent exact-head code reviews are READY at `e48b4af6`. One landing repair remains.

Please rebase this branch onto current `main` `3532f07c` and preserve the reviewed #592 behavior unchanged. Resolve only the expected mechanical overlap in `CHANGELOG.md` and `docs/receipt-schemas.md`; stop if another semantic conflict appears. Run the focused worker-event/archive tests and the normal Brigade verification gate, then push with an exact lease.

Also replace the Cursor wrapper/footer in the PR body with a truthful summary of the final scope and tests. The new head requires fresh Actions, Luna, and Terra review before merge.

PR #873, opened 2026-08-11
fix(evidence): use covering index for status counts

solomonneas:
## Summary

- Count joined item `source_id` values instead of `id` values in the evidence status aggregate.
- Add regression coverage for the covering-index plan and zero-item source counts.

## Why

`brigade evidence status` requests per-source item totals from the evidence engine. The existing query joined on `items.source_id` but counted `items.id`, forcing table-row reads after SQLite found matches through `idx_items_source_external`. Since `items.source_id` is non-null, counting it returns the same totals and lets SQLite answer the join from the covering index.

## Verification

- `go -C engines/evidence-ledger test ./internal/app -run TestCollectStatusSourceCountsUseCoveringIndex -count=1`
- `go -C engines/evidence-ledger vet ./...`
- `go -C engines/evidence-ledger test ./...`
- `./scripts/verify`

On the live archive, direct SQL completed the source-count aggregate in 3.49 seconds. The other status aggregates completed in 0.04, 0.29, and 0.06 seconds.

Two 30-second CLI probes timed out before status collection because the current `main` engine first ran the schema 1-to-4 migration and provenance backfill. The direct SQL timings isolate the status queries from that migration cost.

## Risk

Low. This changes no schema and preserves zero-item rows from the `LEFT JOIN`. The regression test checks both populated and empty sources and pins the current covering-index plan.

### DIFF
--- a/engines/evidence-ledger/internal/app/app.go
+++ b/engines/evidence-ledger/internal/app/app.go
@@ -193,6 +193,10 @@ type Status struct {
 	Capability     map[string]any         `json:"capability,omitempty"`
 }
 
+// statusSourceCountSQL counts joined non-null items.source_id so SQLite can
+// satisfy the LEFT JOIN with COVERING INDEX idx_items_source_external.
+const statusSourceCountSQL = `select s.kind, count(i.source_id) from sources s left join items i on i.source_id = s.id group by s.kind order by s.kind`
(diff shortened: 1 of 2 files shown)


solomonneas:
@coderabbitai review

PR #878, opened 2026-08-12
fix(security): sanitize finding output

### LINKED ISSUE #673: security scan: sanitize text, Markdown, and SARIF output at the finding boundary
## Problem

`brigade security scan` writes unsanitized finding fields to its generated Markdown and SARIF files. Text-mode stdout also prints raw `path`, `title`, `fingerprint`, `evidence`, `suggestion`, and response options.

The CLI describes `--output-dir` as a destination for redacted artifacts, but a finding containing a private URL, token-shaped value, absolute private path, ANSI escape, or control byte reaches those outputs unchanged.

The same scan-time report is sent to `_render_markdown_report()` and `_sarif_report()` in `write_evidence_bundle()`. Text output prints directly from the raw finding records in `scan()`.

## Reproduction

Create a finding whose evidence contains a private URL or token-shaped value, then run the scan in text mode with an output directory. The raw marker appears in stdout, `security-report.md`, and `security-report.sarif`.

This behavior is present on `main`. It was found while reviewing #670.

## Required change

- Build one closed, sanitized finding projection for scan outputs.
- Use it for text stdout, Markdown, and SARIF generation.
- Preserve machine identifiers and repo-relative paths exactly after redaction.
- Do not mutate the stored source report while rendering.

## Acceptance

- Tests inject private URLs, token-shaped values, absolute private paths, ANSI escapes, and control bytes into a scan finding.
- Raw markers are absent from text stdout, Markdown, and SARIF.
- Expected redaction placeholders are present.
- Fingerprints, rule IDs, and safe repo-relative paths remain byte-for-byte stable.
- Existing scanner and security command tests pass.

solomonneas:
## Summary

Sanitize finding-controlled text at the terminal, review, Markdown, and SARIF output boundaries.

## Verification

- `pytest -q tests/test_security_cmd.py`
- `ruff check src/brigade/security_cmd/commands.py src/brigade/security_cmd/models.py src/brigade/security_cmd/reports.py src/brigade/security_cmd/scan_engine.py tests/test_security_cmd.py`

Fixes #673

### DIFF
--- a/src/brigade/security_cmd/commands.py
+++ b/src/brigade/security_cmd/commands.py
@@ -527,6 +527,7 @@ def scan(
         if requested_output_dir is not None:
             print(f"artifacts: {report['artifacts']}")
         for finding in report["findings"]:
+            finding = _sanitize_finding_for_output(finding)
             print(
                 f"- [{finding['severity']}] {finding['category']} "
                 f"{finding['path']}:{finding['line']} {finding['title']}"
--- a/src/brigade/security_cmd/models.py
+++ b/src/brigade/security_cmd/models.py
@@ -9,6 +9,7 @@ import json
 import os
 import re
 import sys
+import unicodedata
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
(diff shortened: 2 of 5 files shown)


solomonneas:
Merge shepherd review block after clean rebase onto `main` (`1a95dc70`).

#673 requires one sanitized finding projection for text, Markdown, and SARIF, with private URLs, token-shaped values, and absolute private paths removed from generated artifacts. The new `_sanitize_finding_for_output()` only removes control and format characters. `write_evidence_bundle()` shallow-copies the report and writes `security-report.json` from the raw finding, so those values remain in the bundle.

Evidence:

- `pytest -q tests/test_security_cmd.py -k writes_redacted_evidence_bundle` passed.
- A focused bundle repro failed through Brigade: verify receipt `20260812-165903-work-verify-fa7370` (`exit=1`). It injects only fake `.invalid` URL/token/path markers and asserts the raw marker is absent from JSON, Markdown, and SARIF.

Leaving this PR draft until the output projection redacts those values and the acceptance case covers all three artifact types.

PR #891, opened 2026-08-12
feat(memory): mint stable card IDs

solomonneas:
## Summary

Implements #867.

- Mints `card-<uuid4>` IDs for new canonical cards and reviewed metadata backfills.
- Preserves valid IDs on replacement and reinforcement.
- Carries legacy path, stem, and topic aliases through care queues, search logs, refresh imports, and retrieval evaluation.
- Refuses reviewed backfill when aliases collide, before any canonical write.
- Writes relative-path-only ID mappings in backfill receipts.

## Compatibility

Legacy cards retain their current topic or filename identity. Explicit-ID cards use the stable ID as primary and keep old keys as aliases, so renames preserve their identity.

## Integration note

#844 is limited to the projection crawl facade on current main and does not yet expose its planned read-only identity audit. This PR does not change evidence-projection files or engine contracts.

## Verification

- Focused final-head suite: 220 passed, receipt `20260812-165842-work-verify-2ea08d`.
- Lint, format, mypy, version sync, managed snapshot, template profile snapshot, and coverage passed in full receipt `20260812-164451-work-verify-f5785f`.
- The full suite then reported 7,098 passed, 3 skipped, and two unrelated host failures: `crontab -n` could not create a tempfile under `/var/spool/cron` in `tests/test_care_cmd.py`.
- Memory Handoff lint: receipt `20260812-165818-work-verify-081a59`.

Refs #867

PR #898, opened 2026-08-12
feat(care): namespace scheduler registrations by target identity

### LINKED ISSUE #759: care: opt-in scheduler install so maintenance runs without a daemon (install/status/uninstall)
## Summary

Brigade's execution model is explicit-invocation only: no daemons, no background processes. That invariant should stay. But it currently leaves a gap: the memory and workspace maintenance loop (inbox drain, staleness checks, receipt pruning) only runs if the operator hand-writes cron or systemd entries. The maintainer workspace does exactly that today with an external scheduler calling `brigade runbook run`, which proves the pattern but ships nothing to users.

Proposal: a `brigade care` command family that scaffolds the user's own scheduler instead of running one.

## Design

- `brigade care install --target .` writes scheduler entries owned by the user: a crontab block on POSIX (systemd user timers as an option), invoking existing brigade commands and runbooks. Entries are managed blocks with a content hash, following the pattern in #732: an honest `care status` check and surgical `care uninstall` removal, never a regex guess.
- `brigade care status` lists installed entries plus the receipt of each entry's last run.
- Every scheduled run writes a normal receipt, so provenance is identical to a hand-typed invocation.
- Scheduled entries are restricted to non-model commands by default. A runbook that dispatches a worker seat can only be scheduled when the config carries an explicit budget gate (see the runbook pack issue). Unattended model spend is a new risk class and stays opt-in twice: once for care, once for budget.
- Windows: #752 and #754 show real Windows users on 0.26.0. First slice may honestly degrade (print the Task Scheduler equivalent instead of writing it), but `care install` must not silently no-op there.

## Acceptance

- install/status/uninstall round-trip leaves the crontab byte-identical to pre-install.
- A tampered or hand-edited entry is reported by `care status`, not clobbered.
- Command inventory and roadmap checks pass (`brigade roadmap commands --check`).

## Out of scope

Any Brigade-owned supervisor or long-lived process. The scheduler is the user's.

solomonneas:
## Summary
- Namespace systemd user units and launchd agents by a hash of the absolute target path.
- Prove uninstall of target A leaves target B registrations intact (systemd + launchd).
- Update care docs/changelog for the target-scoped install/status/uninstall model.

Fixes #759

## Scrutiny
Previous attempt was rejected for cross-target uninstall risk. This diff includes and passes:
- `test_care_registration_identity_is_stable_and_target_specific`
- `test_care_systemd_uninstall_only_invoking_target`
- `test_care_launchd_registrations_are_target_namespaced`

## Test plan
- [x] Cross-target identity + uninstall tests via `brigade work verify run`
- [x] Broader `tests/test_care_cmd.py` excluding host-permission `crontab -n` probes

Note: `test_care_crontab_block_passes_crontab_syntax_check` and `test_care_crontab_percent_path_passes_crontab_syntax_check` fail on this host with `/var/spool/cron/: mkstemp: Permission denied` (environment), not cross-target uninstall logic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -251,10 +251,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   required set fails closed with typed `no-admissible-tool` and one bounded
   replan instead of letting the model improvise tools.
 - `brigade care install|status|uninstall` (#759): opt-in scaffold that writes
-  hash-stamped managed entries into the operator's crontab (or systemd user
-  units), matching `docs/scheduled-care.md`. Memory-care recipes invoke the
-  shipped runbooks so scheduled fires write normal receipts; Windows prints
-  Task Scheduler equivalents instead of silently no-oping; tampered blocks are
+  target-namespaced systemd user timers on Linux or launchd agents on
+  macOS, matching `docs/scheduled-care.md`. Memory-care recipes invoke the
+  shipped runbooks so scheduled fires write normal receipts; other platforms report explicit unsupported scheduling; tampered blocks are
   reported and not clobbered without `--adopt`.
 - Worker event stream field classification and fail-closed scrubbed projections
   for the Codex app-server transport (#592): closed scrub/export classification
--- a/docs/execution-model.md
+++ b/docs/execution-model.md
@@ -22,8 +22,11 @@ A scheduled invocation is still an explicit invocation: the same command runs wh
 
 ## Opt-in care scaffold
 
-`brigade care install --target .` writes managed entries into the operator's
-own scheduler (crontab by default; `--backend systemd` writes user units).
+`brigade care install --target .` explicitly opts the target into the operator's
+own scheduler: namespaced systemd user timers on Linux or namespaced launchd
+agents on macOS. Other platforms report that managed scheduling is unsupported.
+Each registration includes a hash of the absolute target path, so status and
+uninstall can never select another target's entries.
 `brigade care status` audits those entries and shows the latest runbook
 receipt for each recipe. `brigade care uninstall` removes only the
 hash-stamped Brigade block. The schedule still belongs to the operator:
--- a/docs/memory-care.md
+++ b/docs/memory-care.md
@@ -80,7 +80,7 @@ Queue entries include card identity, issue type, severity, priority, safe summar
 
 ## Boundary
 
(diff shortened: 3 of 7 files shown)


solomonneas:
No drift found against #891/#864 — checked: (a) care_cmd.py and cli/care.py contain no card-identity logic at all (no card_id/alias/stem keying; registrations are keyed solely by _target_identity = sha256[:16] of the resolved absolute target path, care_cmd.py:125-128, which is target-path identity, not card identity, so #891's stable card IDs do not intersect this change); (b) the memory_cmd APIs consumed here (_memory_care_runbooks_dir, MEMORY_CARE_RUNBOOK_NAMES, _write_memory_care_runbooks) are unchanged in main after #891; (c) scheduled entries launch the plain CLI argv 'brigade runbook run --approved <runbook> --target .' (care_cmd.py:178, care_cmd.py:412-421) — no wrapping or bypass of run dispatch, so #864 budget enforcement applies on the normal path; (d) care_cmd.py does not import run_budget/run_events/run_lifecycle/run_projector/run_resume/run_transport/aboyeur, and the runbook_cmd internals it reads (_read_runbook, _pin_payload_from_runbook, _run_receipts) are untouched by #864; (e) git merge-tree against origin/main is clean. One unrelated concrete defect found while reading:

1. launchd weekday off-by-one: care_cmd.py:433 maps cron day-of-week to launchd as int(fields[4]) + 1, but launchd and cron use the same numbering (0 and 7 = Sunday, 1 = Monday). The weekly-outcome-ratchet entry ("0 7 * * 1", care_cmd.py:92-93) is Monday 07:00 in cron/systemd but renders Weekday 2 (Tuesday) in the launchd plist — the adjacent comment at care_cmd.py:426 even says "weekly additionally pins Monday".

solomonneas:
## Merge shepherd: reviewed + rebased

Target-identity namespaced care scheduler registrations (systemd + launchd + auto backend). Focused suite: local crontab syntax probes fail on this host (`crontab -n` → `/var/spool/cron/: mkstemp: Permission denied`); remaining care tests pass. Rebased onto current `origin/main` (clean). Marking ready. No independent contract-drift review comments present at check time.

solomonneas:
## Merge shepherd sendback

Verified the independent review finding: launchd `Weekday` used `int(fields[4]) + 1`, so cron Monday (`1`) became launchd Tuesday (`2`) despite the Monday comment. Fixed to pass cron weekday through unchanged, with a regression test asserting weekly-outcome-ratchet → `Weekday: 1`.

PR #917, opened 2026-08-13
fix(work): bound graph snapshot copies to the graphtrail timeout

solomonneas:
## Summary
- Graph snapshot copies during `brigade work verify run` now use the same `graphtrail_delta_timeout_seconds` cap as `graphtrail sync` (10s default, overridable in `.brigade/config.json`).
- If that copy exceeds the cap, the receipt records `code_graph_delta` as `sync_timed_out` with a reason and leaves the verification command result unchanged. A slow graph cannot turn a passing check into a rejected run.

## Test plan
- [x] New unit test: `_backup_sqlite` raises `TimeoutError` and deletes the partial destination
- [x] New verify test: snapshot timeout keeps receipt `completed` with `code_graph_delta.status == sync_timed_out`
- [x] Existing graphtrail timeout tests still pass
- [x] `ruff check`, `ruff format --check`, `mypy`, version/snapshot checks
- [x] Full pytest: 7349 passed, coverage 83.37%

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- Graph snapshot copies during `brigade work verify run` now honor the same
+  `graphtrail_delta_timeout_seconds` cap as `graphtrail sync`. A snapshot that
+  exceeds that cap marks `code_graph_delta` `sync_timed_out` with a reason and
+  leaves the verification command result unchanged, so a slow graph cannot
+  reject a passing check. The 10s default and per-target `.brigade/config.json`
+  key are unchanged.
 - Source and pipx-from-checkout installs now declare `0.27.0`, which PEP 440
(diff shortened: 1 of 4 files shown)


solomonneas:
Focused shepherd review of the timeout contract.

A snapshot copy that exceeds `graphtrail_delta_timeout_seconds` is fail-open: `_backup_sqlite` raises `TimeoutError`, the stale-baseline path records `code_graph_delta.status=sync_timed_out` with a reason, and `_finalize_verify_receipt` still keys receipt status off the command result. Passing checks stay `completed`. The config default remains 10s (`DEFAULT_GRAPHTRAIL_DELTA_TIMEOUT_SECONDS`); no new key.

Independent review asked to map success-path and after-snapshot copy timeouts to `sync_timed_out` as well. Those paths already annotate `capture_failed` and do not reject the verify. Relabeling a successful sync as `sync_timed_out` would be the wrong status. Not a merge blocker.

Waiting on required CI and on #915/#916 landing before rebase and no-drift confirm.

PR #927, opened 2026-08-14
fix(care): discover installed entry registrations

### LINKED ISSUE #914: care: bare status reports 'missing' when only per-entry registrations are installed
Found during the production care-schedule install on 2026-08-13. With three healthy per-entry registrations installed and enabled (care-scan, handoff-ingest, memory-closeout - all with completed first-run receipts), bare `brigade care status --target .` still evaluates against the default atomic five-recipe set and reports `status: missing`. Querying with explicit `--entry` flags reports `status: current` with target_match true and receipts present.

Expected: bare status should enumerate installed per-entry registrations (the #901 model) and report their aggregate state, reserving 'missing' for a target with no registrations at all. As shipped, a healthy per-entry deployment looks broken to any monitoring that calls bare status - which is exactly what dashboards and fleet parity checks will do.

Also worth folding in: `brigade care install` writes unit files but leaves enablement to a printed next_commands step; status could distinguish installed-but-not-enabled from enabled to catch operators who miss that step.

solomonneas:
## Summary

- Bare `brigade care status` now discovers target-scoped per-entry registrations before aggregating their state.
- Systemd output reports `enabled: false` after unit installation until every discovered timer is enabled through `timers.target.wants`.
- Added systemd and launchd regression coverage for empty targets and per-entry registrations.

## Root cause

The no-`--entry` status path always evaluated the five-entry atomic default. A target with only installed memory-care entries therefore included absent default units and reported `missing`.

Fixes #914.

## Validation

- `brigade work verify run --target . --command "./.venv/bin/pytest -q tests/test_care_cmd.py::test_care_bare_status_enumerates_per_entry_registrations_and_systemd_enablement" --capture brigade-work`
- `brigade work verify run --target . --command "./scripts/verify" --capture brigade-work` - 7,403 passed, 3 skipped, coverage 83.44%.


solomonneas:
No drift found against the transaction stack — checked:

- Base freshness: merge-base of fix/914-care-status equals the current main tip (2dbf5a7d), so #922, #921, #923, #924, and #925 are all ancestors of head b5f62eb3. Merge is trivially clean.
- No kernel bypass: the PR only touches the care status path, which is read-only (file reads, is_symlink/resolve checks in the new _installed_launchd_entries / _installed_systemd_entries / _systemd_timer_is_enabled helpers in src/brigade/care_cmd.py). It adds no multi-file writes, so nothing belongs in the projection/kernel.py transaction kernel.
- Care writes were not migrated on main (no transaction usage in main's care_cmd.py); the transactional migrations landed in harness profiles (#921), mcp_cmd.py (#923), and skills_cmd.py (#924) — disjoint from this PR's files.
- No fixture-isolation conflict with #925: that PR changed only tests/test_pre_push_hook.py; the new care test uses the same tmp_path + monkeypatch isolation pattern as neighboring care tests and reuses the existing _unit_path helper.
- Signatures match main exactly (_systemd_backend_status, _launchd_status); all helpers the PR calls (_catalog_by_id, _target_identity, _launchd_dir, _systemd_user_dir, _launchd_plists) exist unchanged on main. The four new helpers are additive with no name collisions.
- Payload consumers unaffected: the only downstream consumer, memory_operations._safe_care_status, reads the top-level enabled key, which the PR does not touch; _launchd_status has exactly one caller (status()), so the empty-entries current→missing change is confined to the intended path. Care's current/missing/tampered vocabulary and exit-code mapping are unchanged.

Issue #762, opened 2026-08-08
dogfood: migrate the maintainer memory jobs onto care-managed entries


solomonneas:
## Summary

The maintainer workspace runs its memory maintenance through hand-written scheduler jobs calling brigade commands. Once #759 and the runbook pack land, those jobs should be reinstalled as `brigade care`-managed entries running the shipped runbooks - same functions, now on the packaged path.

## Why

The feature ships already proven at real scale (the draft-review job alone worked through a four-digit inbox backlog), and any gap between the hand-rolled loop and the packaged loop surfaces here first instead of in a user report.

## Acceptance

- Each existing memory job maps to a care-managed entry or a documented won't-port decision.
- One week of receipts from the migrated entries with no regression in inbox depth or card staleness.
- A short writeup of what the migration surfaced, filed back as issues.

## Closure guard

The receipt window that started at `2026-08-21T17:28:56Z` is invalid. It contained failed `daily-care-pass` receipts, the `memory-closeout` timer was disabled during the interval, and the stale-card count increased. The clock was reset on 2026-08-27 and has not restarted.

Do not close this issue based on implementation presence, milestone counts, or the latest green receipt. Before closing:

- Post a new start timestamp in UTC with baseline values for pending handoffs, total cards, total care issues, and stale cards.
- Confirm valid scheduled status for `handoff-ingest`, `care-scan`, `memory-refresh`, `evidence-crawl`, and `memory-closeout`. Resolve or explicitly defer the nonempty closeout queue under the documented policy.
- Observe seven full 24-hour periods after the new timestamp.
- Audit receipts across the complete interval. Reconcile every failure or deferral against this acceptance contract.
- Record end values showing no increase in pending handoffs or stale cards.
- Post the closing writeup with the receipt interval and baseline comparison.

Code-presence review does not satisfy the timed dogfood requirement.

Depends on #759 and the memory-care runbook pack.

solomonneas:
[...]

Issue #935, opened 2026-08-15
fix(proc): _SeatProcessRegistry is missing terminate, masking seat timeouts as AttributeError


solomonneas:
## Problem

`_SeatProcessRegistry` does not implement `terminate`, so any seat that times out
(or raises) crashes the run with a confusing `AttributeError` instead of reporting
the actual timeout.

Observed on brigade-cli 0.27.0 running `brigade run --worker grok`:

```
error: worker failed: '_SeatProcessRegistry' object has no attribute 'terminate'
```

## Cause

`run_transport.py:470` wraps the registry per seat:

```python
seat_process_registry = process_registry.for_seat(selected_agent.name)
```

`ProcessRegistry.for_seat()` returns a `_SeatProcessRegistry` (`proc.py:58-59`), which
implements only `register`, `unregister`, and `cancel` (`proc.py:94-118`). It does **not**
implement `terminate`, while `ProcessRegistry` does (`proc.py:77`).

Both failure paths in `proc.py` call it on whatever registry they were handed:

- `proc.py:387` — `subprocess.TimeoutExpired` -> `process_registry.terminate(process)`
- `proc.py:412` — `except BaseException` -> `process_registry.terminate(process)`

So the seat-scoped wrapper reaches those lines and raises `AttributeError`. The
underlying process is also left un-terminated, since the call that would have killed
it is what blew up.

## Impact

- A seat timeout is reported as an unrelated `AttributeError`, hiding the real cause.
- The timed-out child process is not terminated, so it can outlive the run.
- Affects any `brigade run` dispatch that goes through a seat-scoped registry, which
  is the normal path for `--worker`.

## Fix

Delegate to the parent, mirroring the existing `register`/`unregister` pattern:

```python
def terminate(self, process: subprocess.Popen[bytes]) -> None:
    self._parent.terminate(process)
```

## Acceptance criteria

- [ ] `_SeatProcessRegistry.terminate` delegates to the parent registry.
- [ ] A regression test asserts the seat-scoped registry exposes the same call surface
      the `proc.py` failure paths depend on (`register`, `unregister`, `cancel`,
      `terminate`), so a future method added to `ProcessRegistry` and used in those
      paths cannot silently go missing from the wrapper again.
- [ ] A test covers the timeout path end to end: a seat that exceeds its timeout
      surfaces the timeout result (code 124), not an `AttributeError`.

PR #937, opened 2026-08-15
fix(care): report per-entry systemd registrations in status_payload

solomonneas:
Refs #762

Worked by the Brigade `grok` seat via `brigade-issue-pr`.

`status_payload` derived its systemd entries from the atomic `CARE_ENTRIES` set, so a target carrying per-entry registrations read as a missing install instead of the entries it actually has. Discovery now matches bare `brigade care status`.

## Verification

`tests/test_care_cmd.py` passes 45/45, captured through Brigade (receipt `20260815-204427-work-verify-dbddfd`). The full suite was **not** completed locally: it was still running after 12 minutes and was holding the worktree, so CI is the authoritative full-suite gate here. Do not merge on a red CI.

## Provenance note

This branch exists because of #935. The seat finished its work, then `brigade run` crashed on the missing `_SeatProcessRegistry.terminate`, so the dispatch reported failure while the child process survived and kept running unsupervised for 24 minutes. The output was recovered from the orphaned worktree and re-applied to a clean branch off the same base commit (463461e9). It is worth reviewing with that history in mind, though the change itself is coherent and self-consistent.

### DIFF
--- a/src/brigade/care_cmd.py
+++ b/src/brigade/care_cmd.py
@@ -1325,9 +1325,13 @@ def status_payload(*, target: Path, home: Path | None = None) -> dict[str, Any]:
     Topology and other read-only consumers should use this instead of scraping
     crontab text or private markers. ``enabled`` is true when either backend has
     a Brigade care block installed for *target* (current, stale, or tampered).
+
+    Systemd discovery matches bare ``brigade care status``: when a target has
+    per-entry registrations (including the #762 memory-job inventory), report
+    those entries instead of treating the absent atomic CARE_ENTRIES set as a
+    missing install.
     """
     target = target.expanduser().resolve()
-    entries = [_entry_status(target, entry) for entry in CARE_ENTRIES]
     if _is_windows():
         return {
             "enabled": False,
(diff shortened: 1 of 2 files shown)


solomonneas:
Landing this. Two review notes, neither blocking:

**1. Changed `Closes #762` to `Refs #762` before merge.** #762 is a milestone-0.27.0 dogfood issue whose acceptance is "each memory job maps to a care-managed entry or a documented won't-port decision", "one week of receipts with no regression in inbox depth or card staleness", and "a short writeup filed back as issues". This PR is the 51-line `status_payload` discovery fix that the migration surfaced, not the migration itself. Auto-closing #762 on this merge would have dropped the remaining dogfood tracking. #762 stays open.

**2. Non-blocking nit:** `status_payload` now computes the same thing twice:

```python
systemd_entries = installed_systemd if installed_systemd else CARE_ENTRIES
...
if installed_systemd:
    selected_entries = installed_systemd
else:
    selected_entries = CARE_ENTRIES
```

`selected_entries` is identical to `systemd_entries`. Worth collapsing to the single binding on a follow-up touch.

Otherwise the fix is correct and mirrors the established pattern in `status()` (`if not entry_ids: entries = _installed_systemd_entries(...)`), so `status_payload` and bare `brigade care status` now agree. Test covers the real regression: a per-entry install reading as disabled.

PR #857, opened 2026-08-10
fix(work): bind task claim to CAS claim records (#856)

solomonneas:
## What and why

`brigade work task claim` moved tasks to `in_progress` without creating a nested CAS `claim` record, so guarded `work reassign` returned `not_claimed` and `work release` could not address the row. This PR recovers orphan commit `81085625` (only that commit's semantics; not the full `cursor/brigade-work-graph-hardening-5764` branch) adapted to current `main`.

Closes #856

## Type of change

- [x] Bug fix

## Changes

- `task_claim` now calls `claim_mod.apply_claim_to_task` for pending tasks, rejects foreign holders with exit 13, preserves `claim_id` on same-actor footprint refine, and includes `claim` / `claim_id` in JSON and text output.
- Six red-to-green regression tests in `tests/test_work_cmd_claim.py`.
- `CHANGELOG.md` Unreleased entry.

### Red (pre-fix, on `main`)

```bash
pytest tests/test_work_cmd_claim.py::test_task_claim_returns_and_persists_claim_id \
  tests/test_work_cmd_claim.py::test_task_claim_enables_guarded_reassign_and_release \
  tests/test_work_cmd_claim.py::test_task_claim_foreign_holder_rejects_without_stranding \
  tests/test_work_cmd_claim.py::test_task_claim_same_actor_refine_keeps_claim_id \
  tests/test_work_cmd_claim.py::test_task_claim_blocked_task_rolls_back \
  tests/test_work_cmd_claim.py::test_task_claim_text_output_includes_claim_id -q
```

Result: **6 failed** — `KeyError: 'claim'` on JSON payloads; foreign holder `task_claim` returned `0` instead of `13`.

### Green (this branch)

```bash
pytest tests/test_work_cmd_claim.py::test_task_claim_returns_and_persists_claim_id \
  tests/test_work_cmd_claim.py::test_task_claim_enables_guarded_reassign_and_release \
  tests/test_work_cmd_claim.py::test_task_claim_foreign_holder_rejects_without_stranding \
  tests/test_work_cmd_claim.py::test_task_claim_same_actor_refine_keeps_claim_id \
  tests/test_work_cmd_claim.py::test_task_claim_blocked_task_rolls_back \
  tests/test_work_cmd_claim.py::test_task_claim_text_output_includes_claim_id -q
```

Result: **6 passed in 0.25s**

```bash
pytest tests/test_work_cmd_claim.py tests/test_work_cmd_footprint.py -q
```

Result: **29 passed in 1.20s**

[...]

Issue #856, opened 2026-08-10
bug(work): task claim omits claim ID and blocks guarded reassignment


solomonneas:
## Summary

`brigade work task claim` moves a task to `in_progress` and records its assignee, but it does not create or return an atomic `claim_id`. The guarded reassignment path then cannot address that claim.

Tested with Brigade `0.27.0.dev20260810` while tracking issue #852.

## Reproduction

1. Add a task from a GitHub issue.
2. Plan the task.
3. Run `brigade work task claim <task-id> --actor <actor> --file <path> --json`.
4. Inspect the task and attempt guarded reassignment.

## Actual result

- The task becomes `in_progress` and records `assignee` plus `claimed_at`.
- No `claim_id` is stored or returned.
- `brigade work reassign` returns `not_claimed` for the task.
- Retrying through the legacy `brigade work claim` path returns `wrong_status` because the task is already in progress.

## Expected result

The task claim operation should atomically create a claim identity that the guarded reassign and release paths recognize. A caller should not need a second claim command after the task has already moved to `in_progress`.

## Acceptance

- `brigade work task claim --json` returns a stable `claim_id`.
- The claim ID is persisted with the task or its canonical claim record in the same operation.
- Guarded reassign and release accept that identity.
- A failed claim operation cannot leave the task in `in_progress` without a usable claim.
- Existing claim and task records remain readable.
- Add red-to-green tests for claim, guarded reassignment, release, and failure rollback.

## Evidence

Task: `20260810-194449-feat-update-install-daily-0-27-beta-wheels-throu-df9a73`.

Brigade verification receipt: `20260810-194643-work-verify-7aff03`.

PR #948, opened 2026-08-17
fix(work): fail-closed work-graph hardening (recovered branch)

solomonneas:
## What and why

Recover the unfinished work-graph hardening campaign from `cursor/brigade-work-graph-hardening-5764` onto current main. Corrupt or future-version `tasks.json` no longer parses as an empty ledger (which let the next mutation wipe recoverable work). Hand-edited unknown edge types survive rewrite, import promote fails closed on cycle-forming proposed edges, and footprint lists/sidecar reads are capped and confined to the workspace target.

The previous campaign ended without a PR. This branch replays only the work main still lacked.

## Recovered from `cursor/brigade-work-graph-hardening-5764`

- Fail-closed task-ledger reads: `TaskLedgerError` with reasons `corrupt_ledger` and `unsupported_ledger_version`
- Preserve hand-edited unknown edge types across `normalize_edges` rewrite
- Catch `EdgeError` / `TaskLedgerError` on `import_promote` (exit 2, import stays pending)
- Footprint size cap (`MAX_FOOTPRINT_ENTRIES = 512`), drop traversal/absolute paths, confine reconcile `sidecar_path` reads to `--target`
- `tests/test_work_cmd_work_graph_hardening.py` (769-line adversarial suite)

## Dropped as already-landed

- `fix(work): bind task claim to CAS claim records` (commit `81085625` on the old branch) already merged via [PR #857](https://github.com/escoffier-labs/brigade/pull/857) / #856, including the follow-up that adopts legacy `in_progress` rows without `--actor`. Not replayed.
- The recovered suite still asserts the dual-surface CAS behavior against current main (`task_claim` cannot steal a live holder; pending claims write a releasable record; same-actor refine keeps `claim_id`).

## Reconciliation with main (~90 commits ahead of the old branch point)

- Replayed onto current main rather than a raw rebase: `ledger.py`, `task_ops.py`, `footprint.py`, and `imports.py` all moved substantially.
- `TaskLedgerError` subclasses `ValueError`. Current `task_add` already has a `ValueError` handler, so the ledger handler is registered first; otherwise JSON `as_dict()` payloads would never emit.
- `#857` already owns `task_claim` CAS binding (and omitted the old branch's `TaskLedgerError` helpers). This PR adds those helpers only, and passes `target=` into `reconcile_footprint` from `task_done` / run completion.
- `ensure_ledger_edges` still stamps `version = 2` after a successful read, so legacy v1 migrates in memory; future versions are rejected before that call.
- Campaign member ledgers already fail closed via their own reader; left untouched.

## Policy inversion: future ledger versions now fail closed

This is a deliberate contract change, not a broken fixture to patch around. CI on `test (3.11)` failed in `tests/test_work_store_measurement.py` because that harness still encoded the old tolerate-future-versions policy. Targeted `test_work_cmd_*` suites missed it (K3 review F4).

**Old contract (measurement protocol 3):** `scripts/measure_work_store.py` wrote a v102 ledger (`TASK_LEDGER_VERSION + 100`) and asserted `schema_version_policy.future_version_rejected is False`. `_read_task_ledger` plus `ensure_ledger_edges` coerced the in-memory version to 2. That was the documented tolerate/coerce policy.

**Why fail-closed replaces it:** Silent-empty or coerced reads of corrupt / future ledgers are how recoverable work gets wiped. The next mutation rewrites `tasks.json` from the empty or downgraded parse. K3 review F1 is the same rationale: hostile ledger state must fail gracefully (`error: ...` / `reason=unsupported_ledger_version` / rc=2) with on-disk bytes untouched, not parse as empty or silently downgrade.

**What moved (the v102 probe was not deleted):** Protocol 4 still writes version 102 and still measures the store's reaction. The cell now requires clean rejection and lossless reporting:

- `future_version_rejected` is `True` (was `False` — assertion inverted because the measured policy inverted)
- `protocol_version` is `4` (was `3` — pass criteria and report fields for this cell changed)
- `observed_policy` is `fail_closed_unsupported_ledger_version` (was `ensure_ledger_edges_coerces_version_to_current` / the sqlite coerce string — those names encoded the old expectation)
- New required proofs: `future_version_reason=unsupported_ledger_version`, `future_version_exit_code=2`, `future_version_bytes_unchanged=true`, `future_version_clean_error=true` (JSON also records `future_version_cli_exit_code=2` and `future_version_after_read=null`)
- Legacy v1 is unchanged: `downgrade_rejected` stays `False`; v1 still migrates in memory

The sqlite characterization adapter now raises `TaskLedgerError` on a future `meta.version` instead of coercing via `ensure_ledger_edges`. Losslessness there is a logical row digest (WAL header churn is not a write).

## Type of change

- [x] Bug fix
- [ ] Docs

## Checklist

- [x] `pytest -q` passes locally (7726 passed, 5 skipped, 68 subtests passed in 38:18)

## Targeted tests

```
python -m pytest tests/test_work_store_measurement.py -q
# 20 passed in 33.62s

python -m pytest tests/test_work_cmd_work_graph_hardening.py \
  tests/test_work_cmd_ledger.py tests/test_work_cmd_edges.py \
  tests/test_work_cmd_footprint.py tests/test_work_cmd_claim.py \
  tests/test_work_cmd_imports.py tests/test_work_cmd_facade.py -q
# 241 passed in 16.42s

python -m pytest -q
# 7726 passed, 5 skipped, 68 subtests passed in 2298.53s (0:38:18)
```

261 targeted tests passed, including the four CI failures. Full suite green locally. Temp-dir smoke: truncated `tasks.json` returns `reason=corrupt_ledger` and is not rewritten; a v99/v102 ledger returns `reason=unsupported_ledger_version` and keeps extra fields.

<div><a href="https://cursor.com/agents/bc-538f21d3-f9ef-4e8b-99d6-5c2433822785?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-538f21d3-f9ef-4e8b-99d6-5c2433822785&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Corrupt or future-version `.brigade/work/tasks.json` ledgers now fail
   closed on every `brigade work` surface (and other commands that read the
   ledger) with `error: ...` and exit 2, instead of parsing as empty or
-  dumping a traceback. Unknown hand-edited edge types are preserved on
+  dumping a traceback. The work-store measurement harness
+  (`scripts/measure_work_store.py`, protocol 4) now reports
+  `schema_version_policy.future_version_rejected: true` with a clean
+  exit-2 error and untouched ledger bytes; it no longer treats future
+  versions as coerced. Unknown hand-edited edge types are preserved on
   rewrite and surfaced via `work ready` (`unknown_type_count`) and a
   `work doctor` WARN. `work import promote --all` returns 2 when any
   import fails. (#948)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   independent `reviewer` or `reviewer_codex` seat. (#920)
 
 ### Fixed
+- Corrupt or future-version `.brigade/work/tasks.json` ledgers now fail
+  closed on every `brigade work` surface (and other commands that read the
+  ledger) with `error: ...` and exit 2, instead of parsing as empty or
+  dumping a traceback. Unknown hand-edited edge types are preserved on
+  rewrite and surfaced via `work ready` (`unknown_type_count`) and a
+  `work doctor` WARN. `work import promote --all` returns 2 when any
+  import fails. (#948)
 - Bare `brigade care status` now discovers target-scoped per-entry
   registrations instead of scoring the atomic five-entry default. A target
   with no discovered units reports `missing`. Systemd status also reports
--- a/scripts/measure_work_store.py
+++ b/scripts/measure_work_store.py
@@ -17,11 +17,14 @@ Protocol defaults match the #846 scoping comment: 10 warmups and 100 measured
 trials, nearest-rank p50/p95, two-claimer barrier race requiring one success
 and one exit 13, and pass/fail gates for #737/#738 anchors where exercised.
 R3 retains the residual no-listener cells: same-actor rejection, guards/empty-filter,
-restart recovery, schema coercion observation, branch/config backup restore,
+restart recovery, fail-closed future-version rejection, branch/config backup restore,
 install footprint, cold-start/backup timing, resource samples, and observed
 metrics-state plus deleted-secret/history checks. Latency remains descriptive
 (no absolute service level). Numbers are never fabricated: cells are
 ``measured``, ``blocked``, or ``unavailable``.
+Protocol 4 inverts ``schema_version_policy``: a future ledger version must be
+rejected cleanly (exit 2, bytes untouched) instead of coerced to the current
+schema.
 
 Standard library plus the in-repo ``brigade`` package only.
 """
@@ -59,10 +62,11 @@ from brigade.work_cmd import ledger as ledger_mod
 ISSUE = 846
 SLICE = "R3"
 KIND = "work-store-characterization"
-PROTOCOL_VERSION = 3
+PROTOCOL_VERSION = 4
 TIMEBOX_NOTE = (
-    "R3 bounded repair: correct the SQLite release characterization while "
-    "retaining the accepted no-listener scope and residual cells. Dolt shapes "
+    "R3 bounded repair plus protocol 4: schema_version_policy records "
+    "fail-closed rejection of future ledger versions (clean error, exit 2, "
+    "bytes untouched) instead of ensure_ledger_edges coercion. Dolt shapes "
     "stay optional characterization candidates "
     "and must not enter Brigade runtime dependencies. Server listeners, TLS, "
     "and permission cells that need a listener remain blocked pending separate "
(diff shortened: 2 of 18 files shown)


solomonneas:
VERDICT: NEEDS WORK

# Review: PR #948 — fix(work): fail-closed work-graph hardening (recovered branch)

Reviewed diff of `063c3f10` against `origin/main` (`df54f458`). Checked out the PR
head in a scratch worktree and ran the claimed test sets plus manual smoke tests.

## Verification of PR-body claims (all confirmed)

- **Reconciliation claims hold.** `TaskLedgerError` subclasses `ValueError`
  (`src/brigade/work_cmd/ledger.py:58`) and the new `except TaskLedgerError`
  handlers are correctly ordered *before* the existing `ValueError` handlers in
  `task_add` (`src/brigade/work_cmd/session/task_ops.py:127,238`) — JSON
  `as_dict()` payloads do emit (verified manually: `--json` claim against a v99
  ledger returns `reason=unsupported_ledger_version`, rc=2, ledger bytes
  untouched, `future_field` preserved).
- **No CAS duplication.** The diff touches no `claiming.py` and contains no
  task-claim CAS implementation — only dual-surface test assertions against the
  behavior landed via #857. The "dropped as already-landed" claim is accurate.
- **Targeted tests pass as claimed.** `tests/test_work_cmd_work_graph_hardening.py`:
  23 passed. `test_work_cmd_ledger/edges/footprint/claim/imports/facade`: 209
  passed. `ruff check` and `ruff format --check` clean on all touched paths.
- **Sidecar confinement works.** `resolve()` + `relative_to(root)` in
  `_delta_payload` (`src/brigade/work_cmd/footprint.py:389-404`) also defeats
  in-target symlinks pointing outside, and both `reconcile_footprint` call sites
  (`session/run_status.py:245`, `session/task_ops.py:712`) now pass `target=`.
  These are the only two callers.
- **No secrets/leaks** in the 769-line test file — placeholder actors/paths only.

## Findings

### F1 (blocking): uncaught `TaskLedgerError` tracebacks on read surfaces

The PR adds clean `_emit_ledger_error` handling to `task_add`, `task_edge_add`,
`claim`, and `import_promote`, but every other command that calls
`_read_task_ledger` now crashes with an uncaught traceback (rc=1) on exactly the
corrupt-ledger state this PR targets. Reproduced on the PR head:

```
$ brigade work tasks --target <corrupt>   # traceback, rc=1
$ brigade work ready --target <corrupt>   # traceback, rc=1
$ brigade work release --actor a --target <corrupt>  # traceback, rc=1
```

Raise site: `src/brigade/work_cmd/ledger.py:96`. Uncaught callers include
`tasks()` (`src/brigade/work_cmd/session/task_ops.py:30`), and callers in
`context_cmd.py:249,672`, `services.py:624`, `verification.py:2185`,
`reviews.py:226,449,733,794`, `handoff_cmd/issue_ops.py:572,619`,
`operator_cmd/migration.py:311`, `repos_cmd/actions_dispatch.py:777`.

Data safety is fine — the raise happens before any write, so nothing is wiped —
but before this PR these read commands rendered an (empty) ledger; after, they
dump a stack trace. For a hardening PR whose purpose is graceful failure on
hostile ledger state, the primary daily read commands should emit the same clean
`error: task ledger is corrupt ...` / rc=2 the mutation paths now produce.
Suggest either a catch in the `work` dispatch layer
(`src/brigade/cli/work/dispatching.py:633`) or per-command handlers.

### F2 (minor): unknown edge types now persist silently forever

`src/brigade/work_cmd/edges.py:122-126` — preserving hand-edited unknown types
is the stated intent and unknowns are correctly excluded from readiness/cycle
logic (`READINESS_EDGE_TYPES`, `edges.py:242,306`), but a typo like
`"dependson"` now survives every rewrite with no surface (no doctor check, no
`ready` warning) telling the user their edge is inert. Consider a WARN-level
doctor check or an `unknown_type_count` in readiness output. Non-blocking.

### F3 (minor): `import promote --all` partial failure returns rc=0

`src/brigade/work_cmd/imports.py:966-968` — rc=2 only when *everything* failed;
a mix of promoted + failed returns 0 with `failed: N` on stdout. Scripts gating
on rc will miss partial failures. Acceptable design, but worth a line in the
command help or changelog. Non-blocking.

### F4 (process): full suite and changelog skipped

The PR body acknowledges `pytest -q` full suite was not run and the
`CHANGELOG.md` Unreleased entry was deliberately left out. The corrupt-ledger
behavior change (F1 tracebacks + fail-closed mutations) is user-visible; per
CONTRIBUTING it should get an entry. CI will cover the full suite, so this is a
process note, but the changelog gap should be closed before merge.

## Test quality

The adversarial suite is genuinely good: real subprocess N-way claim races with a
barrier, kill-mid-write via `os.replace` monkeypatch, truncation mid-object,
v1-migration and v99/string-version rejection, traversal payloads in five forms,
and sidecar escape tested both directions (outside rejected, inside still
joins). Assertions check on-disk bytes, not just return codes. No flakiness
observed in a local run (23 passed in ~2s).

## Summary

Scope matches the issue and the recovery story is honest — the reconciliation
claims in the body all check out against the diff, and the already-landed CAS
commit is not duplicated. Core implementation is correct and well tested. The
one substantive gap is F1: the fail-closed error is only wired into four of the
many `_read_task_ledger` call paths, leaving read-only commands tracebacking on
the exact state this PR hardens against. Fix F1 (and ideally the changelog,
F4); F2/F3 are follow-up material.

solomonneas:
Landing blocker from CI. **Not merging.** This is a real regression, not a flake.

`test (3.11)` failed at Pytest on run `32050149658`; `test (3.10)` and `test (3.12)` were cancelled by matrix fail-fast, so the single 3.11 failure is the whole story.

## Four tests fail in a file this PR does not touch

```
tests/test_work_store_measurement.py::test_cli_writes_measurement_and_fixture_manifest
tests/test_work_store_measurement.py::test_json_residual_cells_reject_hardcoded_metrics_and_secrets
tests/test_work_store_measurement.py::test_measure_matrix_json_and_sqlite_report_shape
tests/test_work_store_measurement.py::test_no_leftover_measurement_dirs
```

All four die the same way:

```
brigade.work_cmd.ledger.TaskLedgerError: task ledger version 102 is newer than supported 2:
  .../brigade-work-store-measure-.../ws-json-.../.brigade/work/tasks.json
src/brigade/work_cmd/ledger.py:136: TaskLedgerError
```

## Why this is a contract change, not a broken fixture

`tests/test_work_store_measurement.py:110` asserts:

```python
assert result["schema_version_policy"]["future_version_rejected"] is False
```

The measurement harness deliberately writes a future ledger version (102) and measures how the store reacts. It asserts the current documented policy: a future version is **tolerated**, not rejected. This PR inverts that policy, so the probe now raises where it used to return, and the assertion is semantically stale on top of that.

So this is not an incidental fixture break. `future_version_rejected` is a measured, asserted behavioral contract, and changing it is arguably a legitimate part of "fail closed on corrupt or future task ledgers" — but the measurement surface has to move with it.

## What is needed

Either update `test_work_store_measurement.py` so the harness expects `future_version_rejected: True` and its probes tolerate `TaskLedgerError`, or make the measurement probe catch the new error and record the rejection rather than propagating it. Whichever way, the `schema_version_policy` report shape should reflect the new fail-closed policy, and the CHANGELOG entry should mention that the store measurement policy now reports future versions as rejected.

## Note on process

This is exactly what F4 in the K3 review predicted. The PR body acknowledged the full `pytest -q` suite was not run locally and deferred to CI; the targeted suites (`test_work_cmd_work_graph_hardening`, ledger/edges/footprint/claim/imports/facade) all passed and missed this because the affected file is outside that set. The fix round for F1-F4 is otherwise sound — I verified the `guard_task_ledger_dispatch` wiring, the `_read_task_ledger` raise, the `unknown_type_count` surface, the promote rc=2, and the changelog entry, and all of those hold. This is the only thing blocking.

Issue #945, opened 2026-08-17
memory: vault-propose write-back path so a vault workflow needs no second tool


solomonneas:
## Problem

Companion to #943. Together they let an operator run the whole vault workflow from Brigade alone, instead of installing and configuring a second tool alongside it.

#943 gives agents a read path: search an operator's vault and cite what they find. This issue is the other half. An agent that learns something durable during a task has no Brigade path to contribute a note back into the operator's vault.

What exists today:

- `memory project-vault` writes into a `Brigade Memory/` subfolder it fully owns. That content is generated and regenerated; it is not a place to put a new note.
- the `note` skill template writes a file to `~/notes/` and then defers the actual move into a vault to an operator-supplied script that Brigade does not ship

So the last hop is either manual or a bespoke script per operator.

## Proposal

```
brigade memory vault-propose --vault <path> --title <title> [--dry-run] [--json]
```

Body read from stdin. Contract:

- **Staged outside the vault.** The proposal is written to an owner-only staging directory first, never composed in place inside the vault.
- **Allowlisted inbox only.** It lands in an operator-named inbox subfolder. It never writes into the `Brigade Memory/` projection, which stays generated, and never overwrites an existing note.
- **Transactional.** Delivery goes through the projection kernel, so a failed write restores rather than leaving a partial note.
- **`--dry-run` is real.** It reports the exact destination and the rendered bytes without touching the vault.

## Safety requirements

These are the failure modes this class of feature actually hits, and they should be requirements rather than review findings:

1. **No swap between validation and write.** Hold the staged source open through delivery so the bytes that were validated are the bytes that land. Re-opening by path between check and write is a time-of-check/time-of-use hole.
2. **Reject symlinked ancestors.** A destination whose parent chain contains a symlink can be redirected outside the vault after validation. Resolve and reject rather than following.
3. **Refuse a staging directory that resolves inside the vault.** Otherwise the staging area becomes vault content and the boundary is meaningless.
4. **Owner-only permissions** on the staging directory and any derived state.
5. **Fail closed on platforms** where the containment checks cannot be performed, rather than falling back to a weaker path check.

## Non-goals

- Editing or reorganizing existing operator-authored notes. Proposals are additive.
- Vault sync.
- Replacing `project-vault`. That stays the writer for generated content.

## Related

- #943: the read path, and the reason this exists as a pair
- #888, #930: the one-way projection both build on
- #931: stable-ID identity, which proposals should adopt rather than re-derive

Issue #943, opened 2026-08-17
memory: vault-search read path to close the projection round trip


solomonneas:
## Problem

`brigade memory project-vault` (#888, merged in #930) writes canonical memory into an operator's Obsidian vault. There is no read direction. An agent working in a Brigade repo cannot search that vault and cite what it finds, so the round trip is half open.

Everything Brigade ships toward Obsidian today is write-only:

- the `note` skill template, which writes a note to `~/notes/` and syncs it to a vault inbox
- `memory project-vault`, which projects canonical memory into a vault subfolder

`memory search` and `memory recall` read only the five canonical stores in `memory_operations.CANONICAL_DESTINATIONS` (cards, rules, `TOOLS.md`, `USER.md`, `.learnings`). Nothing reads an operator vault.

Paired with #945 (write-back), the goal is that an operator configures a vault once and gets the whole workflow from `brigade` alone, with no second tool to install alongside it.

## Why this belongs in Brigade rather than an external tool

Brigade already owns the vault boundary and its safety framing:

- `--vault` is an operator-supplied path, never inferred
- the projection ledger lives under `.brigade/memory-vault/`
- `obsidian_vault.status_payload` reports `path: "redacted:operator-vault"`, so the vault location stays out of topology output
- vault content is already treated as untrusted data, preserved rather than merged, with conflict copies on drift
- `guard.redact_text` already runs over projected content

Reading from a vault is the same boundary in the opposite direction.

## Command surface

```
brigade memory vault-index   [--json]
brigade memory vault-search  <query> [--scope <name>] [--limit <n>] [--include-archived] [--json]
brigade memory vault-show    <note-id> [--json]
brigade memory vault-doctor  [--json]
```

`--scope` names a configured root, not a filesystem path. This is a containment property rather than ergonomics: the allowlist lives in config, so the command cannot be aimed at an arbitrary subfolder of the vault. An unknown scope is an error, not an empty result.

## Result contract

One object per hit:

| Field | Notes |
| --- | --- |
| `id` | stable note identifier |
| `title` | |
| `scope` | which configured root it came from |
| `relative_path` | vault-relative, for citation |
| `content_hash` | `sha256:...`, so a citation can be checked later |
| `updated_at` | |
| `tags` | |
| `snippet` | bounded excerpt |
| `trust` | per-result label, e.g. `untrusted_vault_content` |
| `score` | ranking score, so a caller can see why a hit ranked |

[...]

Issue #931, opened 2026-08-14
vault projection: key wikilinks on stable card IDs with alias dual-read


solomonneas:
Follow-up from the #930 merge (Obsidian vault projection). Independent review verified kernel lifecycle, untrusted-content handling, and redaction clean, but found identity drift the merge deferred as out-of-#888-scope (review comment on #930 has file:line refs):

1. Wikilinks are keyed on title-derived path stems (obsidian_vault.py around lines 454 and 265), not the #891 stable card IDs - which the code computes (line ~250) and then ignores.
2. Projected notes carry no `aliases` frontmatter, so Obsidian cannot resolve links across a rename.
3. Reference collection (_references ~475, by_source ~286) drops any ref expressed as a stable card ID or legacy alias.

This is the third appearance of the same drift class (previously #892 and #896, both fixed pre-merge): new code deriving its own card identity instead of using card_identity()/dual-read. Fix: generate wikilinks and reference edges from stable IDs with alias frontmatter for backward resolution, mirroring the dual-read contract everywhere else.

PR #930, opened 2026-08-14
feat(memory): project canonical memory into an Obsidian vault

solomonneas:
Fixes #888

## What changed

- Adds `brigade memory project-vault` for a one-way projection of canonical memory into an existing Obsidian vault.
- Generated notes include care-state frontmatter and tags, plus wikilinks between related memory.
- Adds category and harness MOC notes, and a JSON Canvas built from the Brigade memory topology.
- Uses the projection kernel for transactional vault and ledger writes, detects vault drift, preserves manual edits through conflict copies, and redacts projected content.

## Verification

- `./scripts/verify` passed through Brigade after rebasing onto current `origin/main`: receipt `20260814-150608-work-verify-2a1012`.
- Focused regression receipts: `20260814-143021-work-verify-261f9c`, `20260814-143150-work-verify-5427bd`, `20260814-143218-work-verify-5cfed1`, `20260814-143305-work-verify-98f262`, and `20260814-145057-work-verify-71b580`.

PR #953, opened 2026-08-17
fix(memory): make the Obsidian projection browsable on a real corpus

solomonneas:
## What

`brigade memory project-vault` (#888, #930) was correct but not browsable at real scale. Running it against a 1,254-card corpus produced:

- an 11.3 MB canvas with **83,586 edges**
- **687 notes named by filename slug**, with the card's real title sitting in its body heading right below
- a `Maps/Harnesses/unknown.md` listing every note, because cards carry no `source_harness`
- canvas nodes in a single row roughly 345,000 pixels wide

Same corpus after this change: **10,373 edges, 1.6 MB**, human-titled notes, no degenerate map.

## Changes

| Fix | Where |
| --- | --- |
| Title a note from its leading heading when frontmatter has no `title`, and strip that heading from the body | `_record`, `_leading_heading` |
| Rank related notes and cap at `MAX_RELATED = 12` | `_relationships` |
| Omit a map whose members are every note | `_render_maps` |
| Grid canvas layout, topology below the note grid | `_render_canvas` |

On titles: the inventory falls back to the filename slug when a card has no frontmatter `title`, which is fine for an index and bad for a vault someone reads. An explicit frontmatter title still wins, and a body heading that differs from it stays as content. Four cards in the test corpus depend on that.

On ranking: every pair sharing a category used to become related, unconditionally. One 221-card category contributed roughly 24,000 pairs on its own. Explicit refs outrank shared tags, which outrank a shared category, and refs always survive the cap.

## Tests

Five tests, each written to fail first:

- `test_project_vault_titles_a_note_from_its_leading_heading`
- `test_project_vault_keeps_an_explicit_frontmatter_title_over_the_body_heading`
- `test_project_vault_caps_related_links_and_canvas_edges`
- `test_project_vault_omits_a_map_that_covers_every_note`
- `test_project_vault_lays_out_canvas_notes_in_a_grid`

`./scripts/verify`: 7,689 passed, 3 skipped, coverage 83.55% against the 78% floor.

## Docs

Adds an `## Obsidian vault projection` section to `docs/memory-operations.md` with a mermaid flowchart of where memory actually flows. The diagram shows the read path from an operator vault as a dashed, absent edge, pointing at #943 and #945.

## Rebased onto #934

Originally this PR carried a follow-up warning: it renames notes whose titles previously came from the filename slug, and projected notes had no `aliases` frontmatter, so Obsidian could not resolve links across the rename. #934 landed that fix, and this branch is now rebased on top of it. The rename hazard is handled upstream and the warning no longer applies.

Both conflicts were in functions the two changes share, and both were resolved by layering rather than choosing:

- `_record`: #934's `identity` / `stable_id` / `aliases` block is kept whole, with the title-and-body derivation appended below it.
- `_relationships`: #934's dual-read `claim()` index builder is kept verbatim, including the ambiguity marking, with the ranking loop layered on top.

The interaction worth naming: #934 marks a dual-read key claimed by two records as `None` rather than last-writer-wins. The ranking loop reads refs through `by_source.get(reference)` and guards on `is not None`, so an ambiguous key yields no link and no score. `_REFERENCE_WEIGHT` only ever applies to refs that resolved unambiguously, so the cap cannot evict a real ref in favour of a tag match, and cannot resurrect a neutralized one.

`_references` and `_render_note` are untouched here, so #934's ID normalization and `aliases:` emission carry through unmodified. All four of its regression tests pass unchanged.

`./scripts/verify` on the rebased branch: 7,705 passed, 3 skipped, exit 0.

Refs #888. Related: #931, #934, #943, #945.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   recover <operation-id>`. Refs #911.
 
 ### Changed
+- `brigade memory project-vault` produces a browsable vault on a real corpus.
+  Notes take their title from the card's leading heading when frontmatter has
+  no `title`, instead of the filename slug, and that heading is no longer
+  repeated in the body. Related links are ranked and capped at 12 per note, so
+  a large shared category no longer forms a complete graph. Maps that would
+  cover every note are omitted, and canvas nodes lay out in a grid rather than
+  one row. On a 1,254-card corpus this took the canvas from 83,586 edges and
+  11.3 MB to a bounded graph. (#888)
 - Center Memory Operations Cards now shows corpus size, a chip cloud, compact
   expandable rows, and 30-row paging. Handoffs is a sibling tab on that page;
   `/view/handoffs` redirects there. Snapshot polling reloads the page so
--- a/docs/memory-operations.md
+++ b/docs/memory-operations.md
@@ -40,6 +40,42 @@ The identity audit is read-only. It reports explicit card ids versus path fallba
 
 The crawl command projects memory into the evidence log only when the installed engine advertises `memory-projection.v1`. It accepts `--dry-run`, `--json`, `--rebuild`, and a positive `--limit`. It rejects the broader `--full` scan. Completed scans may tombstone missing projection rows. Failed or interrupted scans preserve the last completed projection and mark it stale instead of deleting it. Evidence projection status also appears inside Memory Operations inventory rows when evidence artifacts are present or missing.
 
+## Obsidian vault projection
+
+```bash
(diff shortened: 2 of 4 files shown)


solomonneas:
Resolving the one open thread to land: the Copilot observation is accurate - with more than MAX_RELATED explicit refs, ranked[:MAX_RELATED] drops the overflow, so the code comment overstates. Ranking weights put explicit refs ahead of inferred ones, so refs survive in every corpus case up to the cap; the >12-explicit-refs edge case is real but benign (the cap exists to keep maps browsable, which is this PR's whole point). Folding the comment correction into the already-noted follow-up that makes MAX_RELATED configurable, rather than blocking a green PR on a one-line docstring. Resolution reasoning on record per landing protocol.

PR #980, opened 2026-08-18
perf(evidence-ledger): batch search integrity item lookups

### LINKED ISSUE #964: perf(evidence-ledger): applySearchIntegrity issues N+1 queries on search read paths
Follow-up to #586 / PR #960. Latency, not correctness.

## Problem

`applySearchIntegrity` (`engines/evidence-ledger/internal/app/integrity.go:254`) runs one `select ... from items where id = ?` per search result. With the search limit at up to 200, a single read path can issue 200 sequential queries.

## Suggested fix

Batch into one `select ... where id IN (...)` keyed by the result ids, or fold the needed integrity columns into the initial search query so no second pass is required.

## Provenance

Raised by the Copilot reviewer on #960 alongside two security findings (uppercase-digest verification bypass and forensic fail-open on unknown injection status). Both of those were blocking and were fixed in `841e4a03`. This one was judged non-blocking during landing: it is a performance characteristic on a read path, with no integrity or correctness impact.

solomonneas:
## What and why

`applySearchIntegrity` issued one `select ... from items where id = ?` per search hit. Search can return up to 200 rows, so a single read path ran 200 sequential queries. This loads the needed integrity columns in one `IN (...)` query keyed by unique result ids. Integrity behavior is unchanged.

Fixes #964

## Type of change

- [x] Bug fix
- [ ] Docs

## Checklist

- [x] Targeted `go test ./internal/app` integrity suite passes locally (`gofmt` clean, `go vet ./internal/app` clean)
- [x] Conventional commit messages, no AI co-authorship trailers

## Verification

```
cd engines/evidence-ledger
gofmt -l internal/app/integrity.go internal/app/integrity_test.go
go vet ./internal/app
go test ./internal/app -count=1 -timeout 15m -run 'TestSearchSuppresses|TestShowHides|TestForensic|TestHTTPAndMCPHaveNo|TestEvidenceBundle|TestUppercase|TestIntegrity|TestLoadItems|TestApplySearch|TestSearchIntegrity'
```

Result: PASS (`ok github.com/escoffier-labs/miseledger/internal/app`). `TestLoadItemsForSearchIntegrityUsesOneQuery` asserts the item lookup issues a single `from items` select for mixed, duplicate, and missing ids.

<div><a href="https://cursor.com/agents/bc-6c78cf65-5a61-48f2-b6fd-7aa67214ca2b?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-6c78cf65-5a61-48f2-b6fd-7aa67214ca2b&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   recover <operation-id>`. Refs #911.
 
 ### Changed
+- Search integrity verification loads item text and provenance in one
+  `IN (...)` query keyed by the result ids, instead of one select per
+  hit (search limit is up to 200). (#964)
 - `brigade memory project-vault` produces a browsable vault on a real corpus.
   Notes take their title from the card's leading heading when frontmatter has
   no `title`, instead of the filename slug, and that heading is no longer
--- a/engines/evidence-ledger/CHANGELOG.md
+++ b/engines/evidence-ledger/CHANGELOG.md
@@ -10,6 +10,9 @@ Releases before this changelog was started are on the [releases page](https://gi
 
 ### Changed
 
+- Search integrity verification loads item text and provenance in one
+  `IN (...)` query keyed by the result ids, instead of one select per
+  hit (search limit is up to 200). (#964)
 - Memory projection E2 read-path hardening (#843): soft-tombstone superseded
   content-hash versions (and drop their FTS rows) so each
   `(source, collection, external_id)` has one default-live version; generic
--- a/engines/evidence-ledger/internal/app/integrity.go
+++ b/engines/evidence-ledger/internal/app/integrity.go
@@ -247,16 +247,72 @@ func attachIntegrityFields(out map[string]any, view integrityView) {
 	}
 }
 
-func applySearchIntegrity(db *sql.DB, results []SearchResult) error {
-	for i := range results {
-		var text, metadataJSON, rawJSON string
-		if err := db.QueryRow(`select coalesce(text,''), metadata_json, coalesce(raw_json,'') from items where id = ?`, results[i].ID).Scan(&text, &metadataJSON, &rawJSON); err != nil {
-			if err == sql.ErrNoRows {
-				continue
-			}
-			return err
+type itemQuerier interface {
+	Query(query string, args ...any) (*sql.Rows, error)
+}
+
+type searchIntegrityItem struct {
+	Text         string
+	MetadataJSON string
+	RawJSON      string
+}
+
+func searchResultIDs(results []SearchResult) []string {
+	ids := make([]string, 0, len(results))
+	seen := make(map[string]struct{}, len(results))
+	for _, result := range results {
+		if result.ID == "" {
+			continue
 		}
-		view := inspectItemIntegrity(text, rawJSON, metadataJSON, nil, false)
+		if _, ok := seen[result.ID]; ok {
+			continue
+		}
+		seen[result.ID] = struct{}{}
+		ids = append(ids, result.ID)
+	}
+	return ids
+}
+
+func searchIntegrityLookupSQL(n int) string {
+	return `select id, coalesce(text,''), metadata_json, coalesce(raw_json,'') from items where id in (` + placeholders(n) + `)`
+}
+
+func loadItemsForSearchIntegrity(db itemQuerier, ids []string) (map[string]searchIntegrityItem, error) {
+	if len(ids) == 0 {
+		return map[string]searchIntegrityItem{}, nil
+	}
+	args := make([]any, len(ids))
+	for i, id := range ids {
+		args[i] = id
+	}
+	rows, err := db.Query(searchIntegrityLookupSQL(len(args)), args...)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+	out := make(map[string]searchIntegrityItem, len(ids))
+	for rows.Next() {
+		var id string
+		var item searchIntegrityItem
+		if err := rows.Scan(&id, &item.Text, &item.MetadataJSON, &item.RawJSON); err != nil {
+			return nil, err
+		}
+		out[id] = item
+	}
+	return out, rows.Err()
+}
+
+func applySearchIntegrity(db *sql.DB, results []SearchResult) error {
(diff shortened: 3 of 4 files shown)


solomonneas:
VERDICT: READY

Independent review. I did not merge, and I pushed no changes. Everything below was executed in a scratch clone at `pr980`, Go 1.22.2.

This is an integrity read path, so I did not take "the refactor looks equivalent" on faith. I built a differential oracle: the pre-#980 N+1 loop, copied verbatim into a test file, run against the same corpus as the batched version, comparing `Origin`, `Modality`, `TrustLabel`, `Snippet`, and `IntegrityMismatch` field by field.

## What I executed

**1. Full engine suite, as requested.**

```
$ cd engines/evidence-ledger && go test ./... -count=1 -timeout 45m
EXIT=0
```

20 packages ok, zero failures. Targeted integrity run also green:

```
$ go test ./internal/app -count=1 -run 'Integrity|Search|ApplySearch|LoadItems|Forensic|Uppercase'
ok  github.com/escoffier-labs/miseledger/internal/app  2.989s
```

**2. Differential oracle over a hostile id corpus (16 ids).** Exact match, UPPER variant, lower variant, trailing space, leading space, empty string, garbage/missing, duplicates of a *mismatched* item, literal `%`/`_` SQL wildcard chars in the id, a near-miss id under LIKE semantics, plus a legacy (`metadata_json = '{}'`) item and forged-content-hash items.

```
=== RUN   TestProbe980DifferentialOracle
    all 16 ids in the variant table produced identical integrity decisions
    suppressed=7 of 16 results
--- PASS
```

The `suppressed=7` assertion is there on purpose: a differential oracle where nothing is ever suppressed proves nothing. 7 of 16 hit the mismatch path.

On the case-variant question specifically: `items.id` is `id text primary key` with no `COLLATE NOCASE` (`internal/archive/db.go:261`), so SQL `IN` matching is binary-exact and the Go map lookup `items[results[i].ID]` cannot drift from it. If that column ever gains a case-insensitive collation, the SQL would match a row the Go map lookup would miss, and this would fail open. Not a defect today; worth a comment on the map lookup.

**3. Column alignment.** The batched select prepends `id` as column 0, which is the classic place to silently swap `text`/`metadata_json`/`raw_json`.

```
=== RUN   TestProbe980ColumnAlignment
    text="DISTINCT_TEXT_MARKER" metadata~1006 bytes raw="{\"fixture\":\"0be7eb31a1dc\"}"
--- PASS
```

Correct. And `inspectItemIntegrity(item.Text, item.RawJSON, item.MetadataJSON, ...)` preserves the original argument order despite the struct field order differing — I checked that by hand as well.

**4. Mutation check.** I swapped `&item.MetadataJSON` and `&item.RawJSON` in the `rows.Scan` call:

```
--- FAIL: TestSearchSuppressesMismatchedSnippetAndFlagsIntegrity
--- FAIL: TestUppercaseStoredDigestWithTamperedContentRaisesMismatch
--- FAIL: TestApplySearchIntegrityBatchesMultipleResults
--- FAIL: TestSearchIntegrityHandlesMultipleHitsInOnePass
```

Four tests catch it, two of them new in this PR. The tests are real.

**5. Scale, at and past the 200 clamp.** 250 real rows:

```
n=200: 1 query, 200 items loaded
n=250: 1 query, 250 items loaded
```

One query, all items, at and above the search limit.

**6. Leak scan** on the full diff: clean.

## Findings

**1. (low, latent, not reachable today) `searchResultIDs` drops `result.ID == ""`, so an items row with an empty id can never be matched by the batched path.** The legacy `where id = ?` would match it. I demonstrated this diverges, and that the divergence is a fail-open in the suppression direction — not just a cosmetic label difference:

```
=== RUN   TestProbe980EmptyIDSnippetLeak
    legacy : origin="workspace" modality="tool-output" trust="untrusted" snippet="" mismatch=true
    batched: origin=""          modality=""            trust=""          snippet="LEAK-ME" mismatch=false
    SNIPPET LEAK: legacy suppressed the snippet, batched kept "LEAK-ME"
```

Reachability, which is why this is low and not blocking: I traced item id generation to `stableID` (`internal/ingest/importer.go:950`), which is `hex.EncodeToString(h.Sum(nil))[:24]` and therefore always 24 characters. `provenance.NewEvidenceEnvelope` independently rejects an empty item_id (`item_id must be a non-empty string`) — I hit that while writing the probe. To build the fixture above I had to insert the row directly with `db.Exec`, at which point an attacker already owns the ledger. So this is a hardening gap, not a live bypass.

Cheapest fix is to drop the `if result.ID == "" { continue }` skip and let SQL decide, which restores exact legacy parity at no cost — the dedup already handles the repeated-empty-string case. Alternatively assert the invariant. Either way it should not be left implicit in an integrity path.

**2. (low) `loadItemsForSearchIntegrity` is generic but the IN expansion has a ceiling.** Distinct absent ids, this SQLite build:

```
n=200   -> ok
n=999   -> ok
n=1000  -> ok
n=5000  -> ok
n=33000 -> ERROR: SQL logic error: too many SQL variables (1)
n=100000 -> ERROR: SQL logic error: too many SQL variables (1)
```

Both call sites clamp at 200 (`app.go:1942`, `app.go:2429`), so this is unreachable through search today. But the helper is exported within the package with no guard, and the failure mode is the whole search erroring rather than degrading. A chunking loop or a documented precondition on the helper would close it.

**3. (note) Memory tradeoff is real but acceptable.** The old loop held one item's `text` + `raw_json` at a time; the new one holds up to 200 simultaneously. For large raw payloads that is a 200x resident-set change on a read path. Fine at the current clamp, worth remembering if the search limit ever rises.

Correctness holds. The batched query returns byte-identical integrity decisions to the N+1 version across the whole variant table, the column order is right, and the mutation check confirms the tests would catch a regression. The two low findings are hardening, not blockers.

Issue #883, opened 2026-08-12
fix(work): release scanner run authority on proof publication failure


solomonneas:
## Problem

Scanner-run directory descriptors can remain open when import proof publication raises during `--ingest-output` processing.

`_scanners_run_payload` stores retained run authorities, then calls `_append_import_records`. If proof publication or its rollback raises, control leaves the function before the final `_release_scanner_run_directory_authority` loop. Each affected run can retain the run-root and run-leaf descriptors plus its stale `_SCANNER_RUN_DIRECTORY_AUTHORITIES` entry until process exit.

The stale entry does not grant another run authority because `_scanner_run_one` installs a new authority before use. The current impact is descriptor exhaustion on repeated failure paths.

## Expected behavior

Release every retained scanner-run authority from a `finally` block that covers ingestion, receipt rewrites, and proof publication failures. Add a regression that forces `_append_import_records` to raise and checks that both descriptors and the authority-map entry are gone.

Evidence:

- Brigade review run `20260812-123320-final-opus`
- Brigade evidence `d67b0b53c5ca6ded273c7569`
- Observed against PR #871 head `93723f29035c88c58972f91f78e2e3d43146d029`

Follow-up to #871.

PR #982, opened 2026-08-18
fix(work): release scanner run authority inside the scan loop

### LINKED ISSUE #952: fix(work): scanner run authority can still leak inside the scan loop before runs.append
Follow-up to #883 / PR #946. Narrower than the failure path #946 fixes, and **not** made worse by it.

## Problem

PR #946 wraps `_scanners_run_payload` in `try/finally`, and the `finally` releases every run in `runs`. That closes the `_append_import_records` failure path #883 describes, plus the receipt rewrites and the `ingest_errors` early return.

The remaining window is inside the scan loop itself. `runs.append(run)` happens *after* `_scanner_stamp_new_imports` and the loop-local `_write_scanner_run_receipt(run)`, while the authority is installed earlier inside `_scanner_run_one`. So if either of those raises, the run is not yet in `runs` and the `finally` cannot release it. The same applies if `_scanner_run_one` raises after registering the authority.

Impact matches #883: retained run-root and run-leaf descriptors plus a stale `_SCANNER_RUN_DIRECTORY_AUTHORITIES` entry until process exit. The stale entry grants no authority, since `_scanner_run_one` installs a fresh one before use, so this is descriptor exhaustion on repeated failure paths rather than an authority bug.

## Suggested fix

Either append to `runs` immediately after `_scanner_run_one` returns (so the `finally` covers the run from the moment its authority exists), or push the release into a `try/finally` inside `_scanner_run_one` itself. The second is more localized and keeps the invariant next to the code that installs the authority.

A regression should force `_scanner_stamp_new_imports` (and separately `_write_scanner_run_receipt`) to raise, then assert the authority map is back to its prior key set and both descriptors return `EBADF` — the same shape as `test_scanners_run_releases_directory_authority_when_import_publication_raises` added in #946.

## Provenance

Raised in the K3 reviewer comment on #946 and confirmed against the diff during landing. Judged non-blocking for #946, which strictly reduces the leak surface.

solomonneas:
## What and why

PR #946 released scanner-run directory descriptors from `_scanners_run_payload`'s `finally`, but only for runs already appended to `runs`. Stamp and the loop-local receipt rewrite still happened before `runs.append`, and `_scanner_run_one` could raise after installing the authority. Those windows leaked the run-root and run-leaf descriptors until process exit.

This change appends each run immediately after `_scanner_run_one` returns so the payload `finally` covers stamp and receipt-rewrite failures, and releases inside `_scanner_run_one` when it raises after install. The success path still leaves the authority for the caller.

Fixes #952

Follow-up to #883 / #946.

## Type of change

- [x] Bug fix
- [ ] Docs

## Checklist

- [x] Targeted `python -m pytest tests/test_work_cmd_scanners.py tests/test_work_import_provenance_stamp.py -q` passes locally (262 passed)
- [ ] Updated the `Unreleased` section of `CHANGELOG.md` for any user-visible effect (entries describe effects, not commit subjects)

Internal descriptor-release fix; no user-visible changelog entry, same as #946. In-house commit includes the allowed Cursor agent trailer.

## Test plan

- `test_scanners_run_releases_directory_authority_when_stamp_raises` forces `_scanner_stamp_new_imports` to raise and asserts the authority-map key set is restored and both descriptors return `EBADF`.
- `test_scanners_run_releases_directory_authority_when_receipt_rewrite_raises` lets `_scanner_run_one` write receipts, then fails the loop-local rewrite once `provenance_imports_stamped` is set, with the same EBADF assertions.
- Existing `test_scanners_run_releases_directory_authority_when_import_publication_raises` still passes.
- Full scanner + provenance-stamp modules: `python -m pytest tests/test_work_cmd_scanners.py tests/test_work_import_provenance_stamp.py -q` → 262 passed.

[Targeted scanner authority-release tests](https://cursor.com/agents/bc-fac7aef4-59f0-44f0-9b48-433fad1f8027/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fscanner_authority_release_tests.txt)

<sub>To show artifacts inline, <a href="https://cursor.com/dashboard/cloud-agents#my-pull-requests">enable</a> in settings.</sub>

<div><a href="https://cursor.com/agents/bc-fac7aef4-59f0-44f0-9b48-433fad1f8027?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-fac7aef4-59f0-44f0-9b48-433fad1f8027&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/src/brigade/work_cmd/scanners.py
+++ b/src/brigade/work_cmd/scanners.py
@@ -1563,96 +1563,102 @@ def _scanner_run_one(
         "forced": force,
     }
     _SCANNER_RUN_DIRECTORY_AUTHORITIES[id(receipt)] = authority
-    _write_scanner_run_receipt(receipt)
-    if blocker is not None:
-        completed = helpers._now()
-        receipt.update(
-            {
-                "status": "failed",
-                "completed_at": completed.isoformat(),
-                "duration_seconds": (completed - started).total_seconds(),
-                "exit_code": None,
-                "timed_out": False,
-                "error": blocker,
-                "stdout_summary": "",
-                "stderr_summary": blocker,
-                "output_after": _scanner_output_snapshot(output_path),
-            }
-        )
-        _write_scanner_run_file(authority, "stdout.log", b"")
-        _write_scanner_run_file(authority, "stderr.log", (blocker + "\n").encode("utf-8"))
-        _write_scanner_run_receipt(receipt)
-        return receipt
-    if not cwd.is_dir():
-        completed = helpers._now()
-        error = f"scanner cwd does not exist: {cwd}"
-        receipt.update(
-            {
-                "status": "failed",
-                "completed_at": completed.isoformat(),
-                "duration_seconds": (completed - started).total_seconds(),
-                "exit_code": None,
-                "timed_out": False,
-                "error": error,
-                "stdout_summary": "",
-                "stderr_summary": error,
-                "output_after": _scanner_output_snapshot(output_path),
-            }
-        )
-        _write_scanner_run_file(authority, "stdout.log", b"")
-        _write_scanner_run_file(authority, "stderr.log", (error + "\n").encode("utf-8"))
-        _write_scanner_run_receipt(receipt)
-        return receipt
-    _prebind_child_visible_directories(target)
     try:
-        with _scanner_child_environment_sandbox(target) as child_env:
-            completed_process = subprocess.run(
-                argv,
-                cwd=cwd,
-                text=True,
-                capture_output=True,
-                timeout=float(scanner.get("timeout") or 300),
-                shell=False,
-                env=child_env,
+        _write_scanner_run_receipt(receipt)
+        if blocker is not None:
+            completed = helpers._now()
+            receipt.update(
+                {
+                    "status": "failed",
+                    "completed_at": completed.isoformat(),
+                    "duration_seconds": (completed - started).total_seconds(),
+                    "exit_code": None,
+                    "timed_out": False,
+                    "error": blocker,
+                    "stdout_summary": "",
+                    "stderr_summary": blocker,
+                    "output_after": _scanner_output_snapshot(output_path),
+                }
             )
-        stdout = completed_process.stdout or ""
-        stderr = completed_process.stderr or ""
-        _write_scanner_run_file(authority, "stdout.log", stdout.encode("utf-8"))
-        _write_scanner_run_file(authority, "stderr.log", stderr.encode("utf-8"))
-        completed = helpers._now()
-        receipt.update(
-            {
-                "status": "completed" if completed_process.returncode == 0 else "failed",
-                "completed_at": completed.isoformat(),
-                "duration_seconds": (completed - started).total_seconds(),
-                "exit_code": completed_process.returncode,
-                "timed_out": False,
-                "stdout_summary": _scanner_run_summary(stdout),
-                "stderr_summary": _scanner_run_summary(stderr),
-                "output_after": _scanner_output_snapshot(output_path),
-            }
-        )
-    except subprocess.TimeoutExpired as exc:
-        stdout = exc.stdout if isinstance(exc.stdout, str) else ""
-        stderr = exc.stderr if isinstance(exc.stderr, str) else ""
-        _write_scanner_run_file(authority, "stdout.log", stdout.encode("utf-8"))
-        _write_scanner_run_file(authority, "stderr.log", stderr.encode("utf-8"))
-        completed = helpers._now()
-        receipt.update(
-            {
-                "status": "failed",
-                "completed_at": completed.isoformat(),
-                "duration_seconds": (completed - started).total_seconds(),
-                "exit_code": None,
-                "timed_out": True,
-                "error": f"scanner timed out after {scanner.get('timeout')} seconds",
-                "stdout_summary": _scanner_run_summary(stdout),
-                "stderr_summary": _scanner_run_summary(stderr),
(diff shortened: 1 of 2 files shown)


solomonneas:
VERDICT: READY

Independent review. I did not merge, and I pushed no changes. Everything below was executed in a scratch clone at `pr982` (0b86c5da, based on current `main` ee393050), Python 3.12.3.

## What I executed

**1. The shipped tests, and the suites the PR body claims.**

```
$ python -m pytest tests/test_work_cmd_scanners.py tests/test_work_import_provenance_stamp.py -q
262 passed in 8.69s
```

Full suite on the branch, since this touches a descriptor-lifetime invariant:

```
$ python -m pytest tests -q
7957 passed, 3 skipped, 68 subtests passed in 774.69s (0:12:54)
```

**2. Mutation check M1 — are the two new tests real?** I moved `runs.append(run)` back to its pre-PR position (after `_write_scanner_run_receipt(run)`) and left everything else alone:

```
FAILED tests/test_work_cmd_scanners.py::test_scanners_run_releases_directory_authority_when_stamp_raises
FAILED tests/test_work_cmd_scanners.py::test_scanners_run_releases_directory_authority_when_receipt_rewrite_raises
2 failed, 6 passed
```

Both new tests fail when the fix is reverted. They are not vacuous.

**3. Scope check — is the 88-line reindent hiding anything?** I extracted the `_scanner_run_one` body from both revisions, stripped leading whitespace, and diffed:

```
@@ -39,0 +40 @@
+try:
@@ -129,0 +131,5 @@
+except BaseException:
+# The payload finally only releases runs that reached `runs.append`.
+# Release here so a failure after install cannot retain the descriptors.
+_release_scanner_run_directory_authority(receipt)
+raise
```

The reindent is byte-identical modulo indentation. No drive-by logic change. Scope matches #952 exactly (both suggested fixes, no expansion).

**4. Adversarial probes the diff's tests do not cover.** #952 names three failure windows; the PR tests two of them (stamp, loop-local receipt rewrite). The third — "the same applies if `_scanner_run_one` raises after registering the authority" — is not covered by any shipped test, so I wrote four probes for it. All assert the authority map returns to its prior key set and both descriptors return `EBADF`.

- **A** — `_prebind_child_visible_directories` raises inside `_scanner_run_one`, after authority install.
- **B** — the *first* `_write_scanner_run_receipt(receipt)` inside `_scanner_run_one` raises (nothing written yet).
- **C** — `KeyboardInterrupt` rather than `Exception`, to confirm the handler width is really `BaseException`.
- **D** — two scanners: the first completes and is appended, the second raises inside `_scanner_run_one`. Asserts 2 live authorities at raise time and both released after.

On `pr982`:

```
$ python -m pytest tests/test_probe982.py -q
4 passed in 0.20s
```

On `main`, same file:

```
FAILED tests/test_probe982.py::test_probe_a_raise_inside_scanner_run_one - AssertionError: authority map not restored: extra={139797902673664}
FAILED tests/test_probe982.py::test_probe_b_first_receipt_write_raises
FAILED tests/test_probe982.py::test_probe_c_base_exception_inside_scanner_run_one
FAILED tests/test_probe982.py::test_probe_d_multi_scanner_second_raises
4 failed in 0.48s
```

The fix is real on all four, including the `BaseException` width and the multi-scanner composite.

**5. Leak scan** on the full diff (secrets, tokens, hostnames, absolute paths, account names): clean.

## Findings

**1. (medium, non-blocking) The `except BaseException` release hunk has zero test coverage.** Mutation M2: I replaced the handler body with a bare `raise`, deleting the release entirely, and reran both suites the PR body cites:

```
$ python -m pytest tests/test_work_cmd_scanners.py tests/test_work_import_provenance_stamp.py -q
262 passed in 7.55s
```

262 green with half the fix deleted. My probes A-D catch it (4 failed), nothing shipped does. The code is correct — I proved that above — but a future refactor can silently remove it and CI will not notice, which is how #883 and #952 got here in the first place. Suggest adding at least probe A or B; both are ~20 lines and reuse the PR's own `_capture_retained_scanner_run_authorities` / `_assert_scanner_run_authorities_released` helpers. Happy to have a fixer lane drop them in.

**2. (low, pre-existing, not this PR) `_release_scanner_run_directory_authority` pops `_SCANNER_RUN_PUBLICATION_SNAPSHOTS` and `_SCANNER_RUN_DIRECTORY_AUTHORITIES` but not `_SCANNER_RUN_PROOFS[id(run)]`.** Not reachable from the new raise path (the proof is registered by the caller after `_scanner_run_one` returns), so this PR does not make it worse. Noting it because it is the same `id(dict)`-keyed-global shape and the same "grows until process exit" failure mode. Worth its own issue rather than scope creep here.

**3. (note) CHANGELOG omission is defensible.** The unchecked box is deliberate and matches #946, which also landed with no entry; I grepped `CHANGELOG.md` and found no descriptor-release entry for #946 or #883. No user-visible surface changes here.

Nothing blocking. The fix is correct, the scope is clean, and the two shipped tests hold up under mutation.

PR #984, opened 2026-08-18
feat(memory): mint stable card IDs with alias dual-read

### LINKED ISSUE #867: memory: mint stable card IDs with alias-compatible dual-read
## Summary

Mint stable card IDs on new and explicitly reviewed memory writes, then add alias-compatible reads across the existing file-backed memory consumers. Legacy cards remain readable during the staged migration.

## Problem

Card identity currently depends on filenames, path stems, topics, or other consumer-specific keys. Renames can split identity across care queues, recall logs, refresh imports, and evaluation fixtures. The memory projection can audit this state, but the canonical write paths do not mint a durable ID.

### Mint-on-write

- Mint one opaque `card-<uuid4>` ID when a new canonical card is created.
- Mint an ID on an existing card only through a named, reviewed rewrite path.
- Never replace or regenerate a valid existing ID.
- Keep normalized workspace-relative-path fallback reads for legacy cards.

### Alias-compatible reads

- Add dual-read identity handling for care queues, recall and search logs, refresh imports, and memory evaluation fixtures.
- Prefer the explicit card ID when present.
- Preserve the legacy key as an alias during the transition.
- Detect alias collisions and fail without rewriting canonical cards.

### Audit and migration receipts

- Consume #844's read-only identity coverage and collision audit.
- Produce a deterministic dry-run old-to-new mapping receipt before any batch migration.
- Report explicit-ID coverage, fallback coverage, malformed IDs, duplicate IDs, and alias collisions.
- Keep hard missing-ID validation disabled until audited coverage reaches 100 percent.

## Acceptance criteria

- [ ] New cards receive one valid `card-<uuid4>` ID.
- [ ] Repeating a write or reinforcement never changes a valid ID.
- [ ] Legacy cards without IDs remain readable through the documented fallback.
- [ ] Care, recall, search logs, refresh imports, and evaluation fixtures resolve both explicit IDs and legacy aliases.
- [ ] Rename behavior preserves identity for explicit-ID cards.
- [ ] Collision handling is deterministic and makes no canonical edits.
- [ ] Dry-run mapping receipts contain no card bodies or private absolute paths.
- [ ] Missing IDs remain a warning until audited coverage reaches 100 percent.

## Non-goals

- automatic fleet-wide backfill in the first slice
- mass canonical-card rewrites
- scheduler or care migration work
- semantic search replacement
- changing Markdown as the canonical memory authority

## Dependencies

- Design record and operator decision: #833
- Facade and read-only identity audit: #844
- Engine projection contract: #843
- Retrieval evaluation fixtures: #722 and #845

solomonneas:
## What and why

Mint one opaque `card-<uuid4>` ID on new canonical memory writes, and only mint on existing cards through the reviewed `memory care backfill` path. File-backed consumers dual-read explicit IDs and legacy path/stem/topic aliases so renames no longer split identity. Alias collisions fail without rewriting cards. Dry-run backfill writes a mapping receipt with coverage (explicit IDs, path fallbacks, malformed IDs, duplicates, alias collisions) and old-to-new IDs; the receipt has relative paths only, no card bodies.

Fixes #867

## Type of change

- [x] Surface change (new harness, depth, include, or breaking template/ingester change) — opened an issue first per CONTRIBUTING.md

This is the remaining mint-on-write and dual-read slice for #867. #891 landed the first write path; #844/#892 added the read-only identity audit; #934 added `claim()` for vault dual-read. This PR extracts `claim()` into `card_identity.IdentityIndex`, uses it as the mint entry point, and wires the remaining consumers.

## Checklist

- [x] Targeted pytest passed locally (`181` in the identity/care/recall/eval/ingest/vault modules; `2` refresh-import cases)

## Behavior

- New cards receive one valid `card-<uuid4>` ID; repeating a write or reinforcement never changes a valid ID.
- Legacy cards stay readable through the documented workspace-relative path fallback.
- Care queues, recall JSON, search logs, refresh imports, and retrieval-eval fixtures resolve both explicit IDs and legacy aliases.
- `IdentityIndex.claim` marks a colliding key unresolvable; apply/import/eval fail closed and do not edit canonical cards.
- Dry-run mapping receipt: `.brigade/memory-care/backfills/dry-run-mapping.json` (`brigade.memory-identity-mapping.v1`).
- Missing IDs remain a `memory care` doctor **warning** until audited coverage is 100 percent.

## Verification

```text
pytest -q tests/test_card_identity.py tests/test_memory_cmd.py tests/test_memory_hooks.py \
  tests/test_memory_retrieval_eval.py tests/test_memory_search_mcp.py \
  tests/test_memory_identity_audit.py tests/test_memory_vault_projection.py \
  tests/test_card_fingerprint.py tests/test_ingest.py
# 181 passed

pytest -q tests/test_work_cmd_imports.py -k memory_refresh
# 2 passed
```

<div><a href="https://cursor.com/agents/bc-74644c6e-c7ca-4c41-a78c-5beb0c8e7c52?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-74644c6e-c7ca-4c41-a78c-5beb0c8e7c52&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


solomonneas:
VERDICT: BLOCKED - the dry-run mapping receipt does not predict what `--apply` writes, coverage is structurally capped below 100 percent so the missing-ID gate can never lift, and the "never replace a valid existing ID" guard has no test behind it (mutation survives).

Independent review, executed in a scratch clone against real card workspaces built in `/tmp/probe984m` and `/tmp/probe984x`. The core is good: the `IdentityIndex` extraction is a clean lift of #934's collision contract, collision handling is genuinely correct, and idempotence and rename both hold up under probing. The blockers are three acceptance criteria from #867 that the implementation does not reach.

## Blocking

### 1. The dry-run mapping receipt is non-deterministic and does not predict `--apply`

#867: "Produce a deterministic dry-run old-to-new mapping receipt before any batch migration."

`mint_claimed_card_id` is called inside `_backfill_candidates`, which runs on every invocation including each dry run. Every call produces a fresh `uuid4`.

```
$ brigade memory care backfill --target . --json     # dry run 1
new_id = card-1a850eca-7d9b-410c-94c6-9be379c46b58

$ brigade memory care backfill --target . --json     # dry run 2, nothing changed
new_id = card-bda8a574-c6b2-4a0e-af46-26dee2483839

$ brigade memory care backfill --target . --apply --json
new_id = card-b4c8bef2-e872-4592-a193-67f7593ea811

$ grep '^id:' memory/cards/incomplete-legacy.md
id: card-b4c8bef2-e872-4592-a193-67f7593ea811
```

Three runs, three different IDs, and the one actually written matches neither preview. An operator who reviews `dry-run-mapping.json` and then applies gets a different mapping than the one they approved, so the receipt cannot serve its stated purpose as the pre-migration review artifact. The old-to-new pairing is also unverifiable after the fact, since the persisted receipt is overwritten by the apply run.

Options: derive the ID deterministically from the stable legacy key (a uuid5 over `path:<relative_path>` would keep the `card-<uuid>` shape while making dry run and apply agree), or have `--apply` consume the IDs from the existing dry-run receipt rather than minting fresh ones.

### 2. Coverage can never reach 100 percent, so the missing-ID gate can never lift

#867: "Keep hard missing-ID validation disabled until audited coverage reaches 100 percent."

Minting on an existing card only happens here:

```python
if candidate["fields"] and not identity.explicit:
    candidate["id"] = mint_claimed_card_id(claimed_ids, rel)
```

`candidate["fields"]` is only populated when `last_reviewed`, `fresh_until`, or `fingerprint` is missing, and cards with all three are skipped by the earlier `continue`. So a legacy card that has complete care frontmatter but no `id` is unreachable by the only mint-on-existing path.

Two cards, one complete-but-ID-less and one incomplete:

```
$ brigade memory care backfill --target . --apply --json
candidate_count = 1                  # only incomplete-legacy.md
coverage = {"explicit_ids": 1, "path_fallbacks": 1, ...}

# run again, nothing left to do
$ brigade memory care backfill --target . --json
candidate_count = 0
coverage = {"alias_collisions": 0, "duplicate_ids": 0, "explicit_ids": 1, "malformed_ids": 0, "path_fallbacks": 1}
missing_ids_validation = warning

$ grep -c '^id:' memory/cards/complete-legacy.md
0
```

The backfill has terminated with nothing left to propose, and `path_fallbacks` is still 1. `missing_ids_validation` stays `warning` permanently. On a real corpus the cards most likely to already have complete care frontmatter are the well-maintained ones, so this is not an edge case.

The CHANGELOG line "Missing IDs stay a doctor warning until coverage is 100 percent" reads as a staged migration with an end state. As shipped there is no path to that end state.

### 3. Mutation survived: the "never replace a valid existing ID" guard is untested

#867: "Never replace or regenerate a valid existing ID." That is enforced by exactly one clause, `and not identity.explicit`. I removed it:

```python
-        if candidate["fields"] and not identity.explicit:
+        if candidate["fields"]:
```

```
$ python -m pytest -q tests/test_card_identity.py tests/test_memory_cmd.py \
    tests/test_memory_hooks.py tests/test_memory_retrieval_eval.py tests/test_work_cmd_imports.py
192 passed in 19.30s
```

192/192 still green. The mutation is not benign. On a card that already carries a valid ID and is missing one care field, the mutated build corrupts canonical memory:

```
$ brigade memory care backfill --target . --apply --json
written_count = 1
fields = ['fresh_until', 'fingerprint', 'id']

$ cat memory/cards/has-id.md
---
title: Has A Valid Id
id: card-77777777-7777-4777-8777-777777777777
last_reviewed: 2026-01-01
fresh_until: 2026-04-01
fingerprint: 230d8358dc8e...
id: card-32346a8d-c5e3-4bc5-9d70-8f59fca9ced4
---
```

Two conflicting `id:` keys in one frontmatter block. Unmutated behavior is correct (`fields = ['fresh_until', 'fingerprint']`, existing ID untouched), so the code is right and only the test is missing. Please add a case: card with a valid explicit `id` plus a missing care field, assert `id` is not in `candidate["fields"]` and the frontmatter still has exactly one `id:` after `--apply`.

## Non-blocking, ranked

**4. `_backfill_write` appends fields with no duplicate-key check.** It does `additions = [f"{field}: {candidate[field]}" for field in candidate["fields"]]` and splices them before the closing fence, without checking whether the key already exists. That is why finding 3 corrupts silently instead of failing loudly. A guard here would make the whole write path fail-closed regardless of how `fields` was computed.

**5. `coverage.duplicate_ids` does not see duplicate explicit IDs.** With two cards carrying an identical `id`:

```json
"coverage": {"alias_collisions": 1, "duplicate_ids": 0, "explicit_ids": 3, "malformed_ids": 0, "path_fallbacks": 1}
```

The collision is caught (via `alias_collisions`, and correctly), but an operator reading `duplicate_ids: 0` next to it gets a mixed signal. Worth reconciling the two counters or documenting that `duplicate_ids` is namespace-scoped from #844's audit while `alias_collisions` is the cross-card check.

**6. `IdentityIndex.resolve()` returns `None` for both "missing" and "collided".** Callers have to remember `is_collision` to tell them apart. `work_cmd/services.py:71` and `memory_retrieval_eval/{quality,corpus}.py` all do. `obsidian_vault.py:350` calls `by_source.get(reference)` with no collision check, so a collided reference is silently dropped rather than reported. That is the pre-existing #934 behavior and is safe there (dropping a link beats mis-linking), so not a regression. But now that this is a shared API, consider making `resolve` raise on collision with a separate `get_or_none`, so the safe-drop is an explicit choice at each call site rather than the default.

**7. CLI JSON envelope carries absolute paths.** `"target": "/tmp/probe984m"` and `"receipt_path": "/tmp/probe984m/.brigade/memory-care/backfills/dry-run-mapping.json"`. The receipt *file* itself is clean, which is what #867 actually requires. Flagging only because the envelope is the thing an agent is most likely to paste.

## What passed

Executed, all clean:

- **Legacy dual-read.** A card with no minted ID resolves through the shared index by both its normalized relative path and its stem: `'memory/cards/complete-legacy.md' -> 'memory/cards/complete-legacy.md'` and `'complete-legacy' -> 'memory/cards/complete-legacy.md'`. Verified through `work_cmd.services._memory_refresh_identity_index`, and the same `IdentityIndex` backs care, recall/search logs, refresh imports, and the retrieval-eval corpus and quality paths.
- **Rename preserves identity.** After `git mv incomplete-legacy.md renamed-card.md`, `card_identity` returns the same `card-b4c8bef2-...` with `explicit=True`, aliases updated to the new path and stem, and a follow-up backfill proposes nothing (`candidate_count = 0`, no re-mint). Note the old path key stops resolving after the rename, which matches the criterion as written ("rename behavior preserves identity for explicit-ID cards") but does mean consumers holding a stale path key need the ID.
- **Idempotence.** Re-running `--apply` leaves the card byte-identical (md5 unchanged) and the ID unchanged. No diff churn.
- **No bulk rewrite.** `git status --porcelain` after apply shows only the single card the backfill needed. `git diff origin/main...HEAD` touches no existing card files.
- **Collision handling is correct and this is the strongest part of the PR.** Two cards claiming `card-99999999-9999-4999-8999-999999999999`:

```
$ brigade memory care backfill --target . --apply
error: identity alias collisions must be resolved before reviewed backfill
TRUE exit = 2
```

  `dry_run` stays `true` despite `--apply`, `written_count: 0`, both cards byte-unchanged, and the receipt names both paths deterministically. Exit 2, not a silent last-write-win. `IdentityIndex.claim` collisions are also sticky: re-claiming with the original record does not un-collide.
- **Receipt file leak check.** No absolute paths and no card bodies in `dry-run-mapping.json`; `old_id` uses the `path:<relative>` form.
- **Suite**: 192 passed across the five changed test files.
- **CHANGELOG**: accurate, scoped to Unreleased/Added, cites #867. The only overreach is the coverage claim covered in finding 2.
- **`obsidian_vault.py` change is a pure refactor.** The extracted `IdentityIndex` reproduces the previous inline `claim` closure exactly; no behavior change.

Scope tracks #867 closely: mint-on-write, the reviewed rewrite path, alias dual-read across all five named consumer families, collision detection without canonical edits, and the coverage receipt are all present. Blockers 1 and 2 are the receipt-determinism and coverage-ceiling criteria; blocker 3 is a missing test on a criterion the code already satisfies.

solomonneas:
Second probe pass on #984. Verdict is unchanged (BLOCKED); this adds one blocking finding I missed in my first comment, a root cause for one of my non-blocking notes, and two more test/hygiene issues. Everything below I re-ran myself. One claim from the pass did not reproduce and I am recording that too.

## New blocking finding: `mapping_old_id` records a legacy key no consumer ever used

For a legacy card keyed by `topic:`, `card_identity` uses the topic as `card_id`, and consumers record exactly that. `mapping_old_id` returns the path form instead:

```
card_identity.card_id : legacy-topic-key
explicit              : False
aliases               : ('memory/cards/legacy-no-id.md', 'legacy-no-id')
mapping_old_id        : path:memory/cards/legacy-no-id.md
```

Confirmed the consumer side records the topic, not the path:

```
$ brigade memory care scan --target . --json
  card_id= legacy-topic-key  aliases= ['memory/cards/legacy-no-id.md', 'legacy-no-id']
```

So the receipt's `old_id` is `path:memory/cards/legacy-no-id.md` while every care queue entry, search log line, and refresh import for that card is keyed on `legacy-topic-key`. Anyone replaying the mapping receipt to remap those keys finds nothing to remap. Combined with the non-determinism in my first comment, the receipt is wrong on both halves of the old-to-new pair for topic-keyed cards.

`card_identity.py:143-149` falls back to `f"path:{relative_path}"` whenever `identity.explicit` is false. It should return `identity.card_id` for the non-explicit case, or emit both keys. The `path:` convention it borrows comes from `memory_identity_audit`, which uses a different identity model.

This is also why my first comment missed it: my probe cards had no `topic:` field, so `card_id` fell back to the path and the mismatch was invisible. Cards carrying `topic:` are the common legacy shape.

## Root cause for my earlier note on `duplicate_ids`

I flagged `duplicate_ids: 0` sitting next to `alias_collisions: 1` as a confusing counter. The cause is that `memory_identity_audit` only populates `by_namespace` under `if namespace:`, so the counter is dead unless `memory/NAMESPACE` exists:

```
# two cards sharing card-99999999-9999-4999-8999-999999999999, no memory/NAMESPACE
{"alias_collisions": 1, "duplicate_ids": 0, "explicit_ids": 3, "malformed_ids": 0, "path_fallbacks": 1}

# identical corpus, after `echo probe-ns > memory/NAMESPACE`
{"alias_collisions": 1, "duplicate_ids": 1, "explicit_ids": 3, "malformed_ids": 0, "path_fallbacks": 1}
```

`blocking = bool(identity_collisions) or coverage["duplicate_ids"] > 0` still gates correctly because the `identity_collisions` half fires, so this is not a safety hole. But half the gate is inert for any workspace without a NAMESPACE file, and the receipt under-reports duplicates there. The audit's own `blocking_reasons` (including `missing_namespace`) is fetched and discarded.

## Vacuous assertion in a new test

`tests/test_work_cmd_imports.py:2461` reads the wrong path:

```python
inbox = tmp_path / ".brigade" / "work" / "imports.jsonl"
before = inbox.read_text() if inbox.is_file() else ""
...
assert (inbox.read_text() if inbox.is_file() else "") == before
```

The real path is `_work_root(target) / "imports" / "inbox.jsonl"` (`work_cmd/helpers.py:91`). `is_file()` is always False, so the assertion is `"" == ""` and the "inbox unchanged under collision" guarantee is not actually tested. The rest of that test is fine; only this assertion is inert.

## A dry run mutates the working tree

`brigade memory care backfill` without `--apply` creates `.brigade/` in a previously clean repo, even when there is nothing to propose:

```
$ brigade memory care backfill --target . --json
candidate_count = 0
$ git status --porcelain
?? .brigade/
```

I can see this is deliberate (the receipt is the point), but a planning command that writes on a zero-candidate run is surprising, and it means "dry run" no longer means "no side effects". Worth either a note in `docs/memory-care.md` or skipping the receipt write when there are no candidates.

## Smaller items

- **Case sensitivity in the new dual-read.** `valid_card_id` is deliberately case-insensitive, but index lookups use the raw string, so a queue entry with `"id": "CARD-1111..."` and no `file` fails with `requires file`. Same for eval gold keys.
- **`IdentityIndex.records()` is O(n²)** via `all(value is not item for item in seen)`, replacing an O(n) dict iteration in `quality.py`. Not a problem at today's corpus size; a dict keyed on `id(value)` keeps it linear.
- **`IdentityIndex.claim` compares records with `is`, not `==`.** Correct for current callers because each reuses one object per card, but `IdentityIndex[str]` holding plain `rel` strings makes the invariant depend on object identity. Worth a docstring line.
- **Malformed IDs still get a duplicate `id:` key on apply.** A card with `id: card-not-a-uuid` ends up with two `id:` lines, and the original key drops out of `aliases`. Identical on `origin/main`, so not a regression, but this PR added `malformed_ids` to the coverage receipt without blocking the path. Related to my earlier note that `_backfill_write` splices fields with no duplicate-key check.

## One claim I could not reproduce

The pass reported that refresh import creates duplicate work items for one card when it is queued under multiple keys, and called it a regression against `origin/main`. It does not reproduce for me. Dedup works in both shapes I tried:

```
# minted card queued by file AND by explicit id
created = 1  skipped_duplicates = 1

# legacy card queued by topic + stem + relpath, none with "file"
created = 1  skipped_duplicates = 2
```

One task per card in both cases. Recording it so nobody chases it.

## Still standing from my first comment

The three blockers are unchanged: the dry-run receipt does not predict `--apply`, coverage is structurally capped below 100 percent so the missing-ID gate can never lift, and the "never replace a valid existing ID" guard has no test (dropping `not identity.explicit` leaves 192/192 green while corrupting frontmatter). The second pass independently reproduced the non-determinism and the escaped mutation, and separately confirmed that mutating `IdentityIndex.claim` to last-write-wins *is* caught, by three tests across three files. The collision path really is the well-tested part of this PR.

solomonneas:
RE-VERIFIED: READY

All three blockers plus the follow-up `mapping_old_id` blocker are fixed, and each fix is mutation-verified. Re-probed at `10e802d8` with a fixture built specifically around the shape that hid the original bug: legacy cards keyed by `topic:` rather than by path.

## Blocker fixed: dry run now predicts apply

`mint_stable_card_id` derives the ID as UUID5 over a stable seed with the version nibble forced to 4, so the shape `valid_card_id` accepts is preserved. Three dry runs and the apply now agree:

```
DRY RUN 2:  complete-topic-key -> card-71d14912-6e3a-4dc3-94a9-699d7590f03e
            legacy-topic-key   -> card-38c1e5ce-9fab-4dcf-9361-6eadedc91196
DRY RUN 3:  complete-topic-key -> card-71d14912-6e3a-4dc3-94a9-699d7590f03e
            legacy-topic-key   -> card-38c1e5ce-9fab-4dcf-9361-6eadedc91196
APPLY:      complete-topic-key -> card-71d14912-6e3a-4dc3-94a9-699d7590f03e
            legacy-topic-key   -> card-38c1e5ce-9fab-4dcf-9361-6eadedc91196

$ grep -H '^id:' memory/cards/*.md
memory/cards/complete-topic.md:id: card-71d14912-6e3a-4dc3-94a9-699d7590f03e
memory/cards/legacy-topic.md:id: card-38c1e5ce-9fab-4dcf-9361-6eadedc91196
```

The written IDs match the receipt exactly. The receipt is now a real preview.

## Blocker fixed: `mapping_old_id` records the consumer key

`old_id` is now `identity.card_id` for non-explicit cards, so the topic-keyed case reads `complete-topic-key` and `legacy-topic-key` rather than the `path:` form no consumer ever used. That is the exact fixture that hid this from my first pass.

Mutation check, reverting to the explicit-only branch:

```
FAILED tests/test_card_identity.py::test_legacy_cards_use_path_fallback_and_keep_aliases
FAILED tests/test_memory_cmd.py::test_memory_care_backfill_dry_run_mapping_predicts_apply
FAILED tests/test_memory_cmd.py::test_memory_care_backfill_mints_id_on_complete_idless_card
3 failed, 49 passed
```

## Blocker fixed: coverage reaches 100 percent and the gate lifts

A legacy card with complete care frontmatter and no `id` is now a candidate (`candidate_count = 2` where it was 1). After apply:

```
candidate_count = 0
coverage        = {"alias_collisions": 0, "duplicate_ids": 0, "explicit_ids": 2, "malformed_ids": 0, "path_fallbacks": 0}
missing_ids_validation = enabled
```

`path_fallbacks` reaches 0 and `missing_ids_validation` flips from `warning` to `enabled`. The staged migration now has an end state, which is what the criterion asked for.

## Blocker fixed: the "never replace a valid existing ID" guard is tested

Re-applied my original mutation against the refactored guard at `memory_cmd.py:1465`:

```
FAILED tests/test_memory_cmd.py::test_memory_care_backfill_never_replaces_valid_existing_id
1 failed, 43 passed
```

That mutation previously left 192/192 green while writing a duplicate `id:` key into canonical frontmatter.

## Other findings from my follow-up, re-probed

```
duplicate_ids without memory/NAMESPACE : {"alias_collisions": 1, "duplicate_ids": 1, ...}   (was 0)
dry run, clean repo, zero candidates   : candidate_count = 0, git status -> []              (no .brigade/ created)
idempotent re-apply                    : byte-identical, no re-mint
collision + --apply                    : exit=2, "identity alias collisions must be resolved
                                         before reviewed backfill", canonical cards unchanged
```

Suite: 204 passed across the five changed test modules plus `test_memory_identity_audit.py`, up from 192.

## One note on the imports assertion at `tests/test_work_cmd_imports.py:2461`

The path fix is correct (`imports/inbox.jsonl`), and the assertion is real, but it took a two-part mutation to prove it, which is worth recording because single-guard mutations are misleading here:

- Removing only the `continue` in the collision branch: 97 passed. The collided card still never gets written, because `if errors:` returns 2 before `_append_import_records`.
- Removing only the `if errors:` write gate: 97 passed. `records` is empty, because the `continue` already dropped the card.
- Removing both: `FAILED test_work_import_memory_refresh_rejects_alias_collisions_without_writing`, 96 passed.

So the assertion does bite, and the reason a single mutation survives is genuine defense in depth rather than a weak test. Recording it so a future reviewer running a one-guard mutation does not misread the survivor as a vacuous assertion.

## Smaller items from my follow-up, still open

Non-blocking and not gating: case sensitivity in dual-read lookups (`"CARD-1111..."` with no `file` still fails), `IdentityIndex.records()` remains O(n squared), `claim` still compares with `is` rather than `==`, and malformed-ID cards still gain a duplicate `id:` key on apply (identical on main, so not a regression). Worth a follow-up issue rather than another round here.

Good turnaround. Every blocking finding is fixed at the root rather than papered over at the assertion, and the coverage gate actually lifting is the part I most wanted to see.

PR #987, opened 2026-08-18
fix(claude-hooks): stop shared-workspace writes from re-arming closeout

### LINKED ISSUE #959: claude-hooks: Stop gate blocks on another session's writes in a shared workspace
## Symptom

In a shared agent workspace, the Claude Stop hook blocks closeout with:

```
Recent write work in Brigade-wired repos has no verification receipt for this session.
```

Running the exact command it suggests produces a completed receipt, and the next Stop re-blocks with the same message. Observed 10 times in one session. The block is unsatisfiable from inside the session: there is no sequence of captures that ends it.

## What it is not

Two plausible explanations were tested and ruled out, so the obvious fixes will not work:

1. **Not the verify command marking itself as write work.** `runtime.py` short-circuits routed verifies in `PostToolUse` before the `wrote` branch, so `brigade work verify run` does not set `last_write_at`.
2. **Not a detection miss from shell wrapping.** `_is_routed_verify` was exercised directly against the four command shapes in play (bare, piped to `grep`, env-prefixed, both) and returns `True` for all of them.

## What the evidence points at

The target is an operator workspace with more than one agent session working in it concurrently. Session state for one session:

```
write_observed              = True
last_write_at               = 2026-08-17T17:58:59Z
last_verification_write_at  = 2026-08-17T17:58:59Z
started_at                  = 2026-08-17T15:58:07Z
```

The last capture from this session produced a receipt stamped `17:58:46Z`, thirteen seconds *before* `last_write_at`. Nothing from this session ran in that window. Receipts landing in the same target minutes later belong to a different agent entirely, running unrelated commands.

The Stop gate compares a per-session receipt against a threshold derived from writes to a **shared tree**. When two sessions work the same target, session A's writes advance session B's `last_write_at`, so B is told it has unverified write work it never performed. B captures, A writes again, and the gate re-arms. The rate of blocking is set by how busy the other agent is, which is why it presents as an unsatisfiable loop rather than an occasional false positive.

## Suggested direction

Not proposing an implementation, since the ownership model is the real question. Options that seem worth weighing:

- Attribute observed writes to the session that made them, rather than treating any worktree delta as the current session's write work.
- Exclude Brigade's own bookkeeping paths (`.brigade/work/verify-runs/`, `.brigade/work/claude-hooks/`) from the fingerprint that feeds the threshold.
- Detect a concurrently-active session in the target and downgrade the block to a warning, since a hard block that another process controls cannot be cleared by the blocked agent.

A workspace shared by several harnesses is a normal Brigade deployment, so this is likely to reproduce for anyone running more than one agent against one hub.

## Reproduce

1. Open two agent sessions against the same Brigade-wired workspace.
2. In session A, do ordinary work that writes files, on a loop.
3. In session B, make one write, then run the captured verification the Stop hook asks for.
4. Session B's Stop hook blocks again, citing write work session B did not do.

solomonneas:
## What and why

The Claude Stop hook treated the shared worktree as this session's write work. In a multi-agent hub, session A's writes advanced session B's `last_write_at` and invalidated B's receipt against the live tree, so B's closeout looped no matter how many times it captured.

Review of #987 found the loop is closed, but neighbor detection was over-broad: a read-only concurrent session counted as foreign write evidence, and the Stop downgrade could swallow this session's own post-receipt writes. This revision tightens foreign-write detection to write evidence only, and keeps a hard block when this session itself wrote after its last capture.

Fixes #959

## Type of change

- [x] Bug fix
- [ ] Docs

## Checklist

- [x] Targeted pytest for the touched modules (`tests/test_claude_hooks_runtime.py`, `tests/test_claude_hooks_doctor.py`: **117 passed**)

## Behavior

- Bash worktree deltas are not attributed to this session when another Claude write heartbeat (`write_observed` / `last_write_at` / `pending_write_at`) or other-harness receipt falls in the same window. A read-only neighbor that only touches its session file is not foreign write evidence.
- Write-tool PreToolUse records `pending_write_at` so a concurrent Write is visible before PostToolUse.
- A time-valid session receipt still counts when the live tree has drifted under a foreign writer.
- After this session has captured, a hard block that another process can re-arm is downgraded to a warning. Sessions that never captured, or that wrote again after their own receipt (Write/Edit/Bash PostToolUse), still hard-block even when neighbors are alive.

## Review fixes (blocked comment)

- Finding 1: drop bare `st_mtime` neighbor test in `_foreign_write_since` (`d951c34c`).
- Finding 2: Stop downgrade defers to this session's own unverified attributed writes (`e05802d3`, follow-up `26bb5300`).
- Finding 3: `test_stop_still_blocks_when_this_session_writes_after_its_receipt` now uses foreign-write-after-own-write order (`8e8816fc`).

Targeted test output: 117 passed in 35.64s.

<div><a href="https://cursor.com/agents/bc-3b16a527-f554-459d-9eed-f42a4d65864d?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-3b16a527-f554-459d-9eed-f42a4d65864d&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


solomonneas:
**VERDICT: BLOCKED** - the #959 loop is genuinely closed, but the same predicate that closes it also lets a purely read-only concurrent session (1) silently drop this session's real Bash write and (2) downgrade this session's own unverified write to an advisory. Both are regressions against `main`, and both fire in exactly the deployment #959 was filed for.

## Method

Scratch clone, `gh pr checkout 987`. Probes drive the real managed hook through the CLI (`brigade work hook-run --event <E> --package brigade-claude-work-loop@1.2.0`, JSON payload on stdin), two session ids against one wired target, with real receipts from `brigade work verify run` and `BRIGADE_CLAUDE_SESSION` set per session. The identical script was run against `main` for a baseline.

Fixture note, since it cost two false failures before I caught it: in a freshly wired repo the first `brigade work verify run` rewrites the managed `.gitignore` block and writes `memory/outcome/records.jsonl`, which drifts the audited tree *after* the receipt is stamped, so `_receipt_since` can never match. The probes gitignore `/memory/` and `/error.log` (as this repo does) and settle one warm-up capture first.

## Probe 1: the #959 loop is closed

| scenario | `main` | PR 987 |
|---|---|---|
| B captures, no foreign write, B Stop | pass | pass |
| A writes after B's receipt, B Stop | **BLOCK** | no block |
| A writes again (the loop), B Stop | **BLOCK** | no block |
| 13s pattern: B receipt at T-14s, `last_write_at` pinned to A's write at T-1s | **BLOCK** | no block |
| same state re-evaluated (the unsatisfiable repeat) | **BLOCK** | no block |

The fourth row is reconstructed from the issue: B's receipt is stamped 13 seconds before the `last_write_at` that A's write pinned, with A's session state carrying the matching `last_write_at`, and B's `started_at` two hours back. On `main` that state hard-blocks and stays blocked on repeat evaluation. On this branch it resolves to a warning and stays resolved. Direction confirmed.

## Probe 2: does the gate still fire? Four scenarios regressed

**Finding 1 (blocking): a read-only concurrent session makes this session's Bash write vanish.**

`_bash_write_detected` now bails through `_foreign_write_since`, which ends in `_other_session_state_touched_since` - an `st_mtime` test on other sessions' state files. Every tool call by every other session rewrites its state file, including read-only ones. Probe: B runs `python3 build.py`, which really writes `real_source.py`; inside B's Bash window session A does one `git log -1` PreToolUse and nothing else.

```
[PR-987] B write_observed=False last_write_at=None -> Stop: PASS   (no receipt asked for, no warning)
[MAIN]   B write_observed=True                    -> Stop: BLOCK
```

A session that only read cannot have caused a worktree delta, so this is not ambiguity being handled conservatively, it is attribution being discarded. `main`'s narrower test (other state has `write_observed is True` and `last_write_at >= started`) was the defensible version of the same idea, and the loop directly above `_other_session_state_touched_since` already performs it.

**Finding 2 (blocking): after one capture, this session's own unverified write downgrades to a warning whenever any neighbor is alive.**

`_shared_tree_block_is_foreign` asks only whether the session captured *at some point* since `started_at` (not whether a receipt covers the current threshold), then whether any foreign activity followed the threshold. Three probes, all with B holding genuinely unverified work it wrote itself after its own receipt:

```
B's unverified write, then A writes            -> PR: WARN   main: BLOCK
B's unverified write, then A only reads        -> PR: WARN   main: BLOCK
B's unverified write, then a different session's verify receipt only, zero writes
                                               -> PR: WARN   main: BLOCK
```

The PR body says "Sessions that never captured, or that wrote again after their own receipt, still hard-block." The first clause holds (probed, still blocks). The second holds only when no foreign activity follows the session's own write. `test_stop_still_blocks_when_this_session_writes_after_its_receipt` passes because it orders the foreign write *before* B's own write; swap those two calls and it warns instead. In a busy hub the swapped ordering is the common one, which is the premise of the issue.

The emitted text is also wrong in these cases. It opens with "Verification is recorded for this session" when B wrote after its receipt, and asserts "a concurrent session has written since this session's last capture" when the neighbor only read.

**Finding 3 (non-blocking): `_receipt_since` drift fallback uses `min(drifted)`.** With several drifted receipts, a foreign write after the *oldest* one satisfies the newest. `max(drifted)` matches the intent (a foreign write after the receipt being relied on) and costs nothing.

**Finding 4 (nit): `pending_write_at` leaks on a failed write.** It is set in PreToolUse for `_WRITE_TOOLS` and cleared only inside the `if wrote:` branch of PostToolUse. A denied or errored Write never reaches PostToolUse, so the timestamp persists and marks that session as a foreign writer at that instant for every later evaluation. Bounded (fixed timestamp), but a `PostToolUseFailure` cleanup would close it.

## Probe 3: bookkeeping paths

A write only under `.brigade/work/verify-runs/` leaves `write_observed` false and `last_write_at` None on both branches, and Stop is clean; the control (a real source write through the identical Bash shape) arms the gate on both. `.brigade` is already in `_SNAPSHOT_IGNORE_DIRS`, so direction 2 from the issue was satisfied before this PR and correctly was not re-implemented.

## Probe 4: mutation check

Three core hunks reverted one at a time, restored after each run. Each kills exactly one new test:

```
_bash_write_detected foreign check -> main's narrow loop
  FAILED tests/test_claude_hooks_runtime.py::test_stop_ignores_other_harness_receipt_as_this_session_write
  1 failed, 113 passed

Stop downgrade branch disabled
  FAILED tests/test_claude_hooks_runtime.py::test_stop_warns_when_foreign_write_outruns_this_session_receipt
  1 failed, 113 passed

_receipt_since tree-drift fallback disabled
  FAILED tests/test_claude_hooks_runtime.py::test_receipt_since_accepts_tree_drift_when_foreign_session_wrote
  1 failed, 113 passed
```

The tests are real and pinned to the behavior they claim. Unmutated, `tests/test_claude_hooks_runtime.py tests/test_claude_hooks_doctor.py` is 114 passed via `brigade work verify run`, matching the PR body.

## Scope and hygiene

5 files, all under `src/brigade/claude_hooks/` plus tests and CHANGELOG. No drive-by refactors: `_iter_verify_receipts` and `_receipt_target_matches` are extractions the new code actually reuses. One `Fixed` entry under `Unreleased`, written as effects. No absolute paths, hostnames, or credentials in the diff; `content-guard` and every other check is green. Copilot's review never ran (quota), so this is the first substantive pass.

## Which ownership direction, and does it close the loop without weakening the gate

The PR takes direction 1 (attribute writes to the session that made them) and direction 3 (downgrade to a warning when another session is active), and correctly skips direction 2 because `.brigade` was already excluded from the fingerprint. Direction 1 is implemented as a negative rather than a positive: instead of establishing which session made a worktree delta, the code establishes that somebody else *might* have, then declines to attribute. That is a fair reading of an unattributable signal, and together with the `pending_write_at` heartbeat and the `_receipt_since` drift tolerance it does close the loop, which the probes confirm against the exact state from the issue. What it does not do is keep two different questions apart. The same `_foreign_write_since` predicate answers both "could another session have caused this delta" (where a read-only neighbor is irrelevant, but currently counts) and "is this block outside the session's control" (where the session's own directly attributed writes should dominate, but currently do not). Because the predicate is satisfied by mere activity rather than by evidence of a write, the gate stops being enforceable for any session sharing a hub with a live neighbor, which is the same population the issue was filed for: it trades an unsatisfiable false positive for a routine false negative. Tightening `_foreign_write_since` to require write evidence (drop the bare mtime test, or gate it on the other state carrying `write_observed` / `last_write_at` / `pending_write_at` inside the window), and making the Stop downgrade defer to this session's own post-receipt attributed writes (a `Write`/`Edit` PostToolUse is direct attribution, not a shared-tree inference), keeps every row of the first table passing while restoring the four blocks that regressed.

solomonneas:
**RE-VERIFIED: READY**

Re-ran the full probe set from https://github.com/escoffier-labs/brigade/pull/987#issuecomment-5330449881 against `26bb5300`. Both blocking findings are fixed, and fixed in the direction I asked for rather than around it: `_other_session_state_touched_since` is deleted outright rather than tuned, and the Stop downgrade now defers to a recorded `last_attributed_write_at` rather than to a wider foreign-activity window.

## Loop closure still holds

Same two-session probes through the real `brigade work hook-run` CLI, `main` alongside for baseline:

| scenario | `main` | `26bb5300` |
|---|---|---|
| B captures, no foreign write, B Stop | pass | pass |
| A writes after B's receipt, B Stop | **BLOCK** | no block |
| A writes again (the loop), B Stop | **BLOCK** | no block |
| 13s pattern from #959 | **BLOCK** | no block |
| same state re-evaluated | **BLOCK** | no block |

Fixture correction on my side, worth recording since it briefly looked like a regression: my original 13-second reconstruction back-dated the receipt but left `last_attributed_write_at` at wall-clock now, which made B's own write postdate its own receipt. With `last_attributed_write_at` placed before the receipt (faithful to the issue's dump, where the receipt at 17:58:46Z follows B's real work and only `last_write_at` is foreign-pinned at 17:58:59Z), rows 4 and 5 pass. The block I saw on the first re-run was the new guard doing its job on a state I had built wrong.

## Both regressions closed

```
B's own unverified write + neighbor that only writes    -> BLOCK  (was WARN)
B's own unverified write + neighbor that only reads     -> BLOCK  (was WARN)
B's own unverified write + foreign receipt only, 0 writes-> BLOCK  (was WARN)

B runs `python3 build.py` writing real_source.py, A does one `git log -1` inside B's Bash window:
  write_observed=True, last_attributed_write_at=set -> Stop: BLOCK   (was False / PASS)
```

## Solo-session enforcement is unchanged from `main`

```
                                    main      26bb5300
unverified write               ->   BLOCK      BLOCK
after capture                  ->   PASS       PASS
writes again after capture     ->   BLOCK      BLOCK
```

## Mutation check on the new tests

Each fix hunk reverted in isolation, restored after every run:

```
bare session-state mtime restored as foreign evidence
  FAILED ::test_posttooluse_records_bash_write_when_other_session_only_reads
  1 failed, 116 passed

own-post-receipt-write guard disabled
  FAILED ::test_stop_still_blocks_when_this_session_writes_after_its_receipt
  FAILED ::test_stop_still_blocks_own_unverified_write_when_other_session_receipt_only
  2 failed, 115 passed

datetime coercion in _receipt_since reverted
  FAILED tests/test_claude_hooks_doctor.py::test_doctor_ignores_foreign_write_outrunning_this_session_receipt
  1 failed, 116 passed
```

`test_stop_still_blocks_own_unverified_write_when_neighbor_only_reads` survives either single revert because both fixes independently cover it. Reverting both together fails all four new tests (`4 failed, 113 passed`), so it is load-bearing rather than vacuous. The re-ordered `test_stop_still_blocks_when_this_session_writes_after_its_receipt` now runs write, receipt, write again, foreign write, Stop, which is the ordering that previously hid the downgrade.

Clean: `tests/test_claude_hooks_runtime.py tests/test_claude_hooks_doctor.py` is **117 passed**, matching your number.

## Three notes, none blocking

1. **Full local suite on `26bb5300`: 7966 passed, 3 skipped, 68 subtests passed** (11m40s, exit 0), up from 7963 on the previous head by exactly the three new tests. No collateral breakage.
2. **CI has not run on the new head.** `gh run list --commit 26bb5300` returns nothing; only CodeRabbit (skipped by label) has reported. The green run referenced earlier was on `5a2d9dee`. I ran the full local suite instead, and the four commits are additive over an already-green tree, but a maintainer should let Actions land before merging rather than reading the old green as current.
3. **The two non-blocking nits from the first pass are still open**, which is fine to defer: `_receipt_since` still uses `min(drifted)` where `max(drifted)` matches the intent, and `pending_write_at` is still cleared only inside the `if wrote:` branch, so a denied or errored Write leaves it set and marks that session as a foreign writer at that instant for later evaluations.

Residual behavior worth knowing rather than fixing here: `last_attributed_write_at` is recorded for Bash-inferred writes too, not only `Write`/`Edit` as the docstring suggests. That errs toward blocking, so it is the safe direction. The one shape still uncovered is a cross-harness writer that leaves neither Claude session state nor a receipt: nothing marks it foreign, so its delta can be attributed to this session. That gap predates this PR and is not a regression from it.

Verdict stands as ready to merge once Actions is green on `26bb5300`.

Issue #582, opened 2026-07-26
evidence: define shared ProvenanceEnvelope schema and trust policy


solomonneas:
Parent: #505
Proposal: #581

Implement the shared schema and policy contract from `docs/proposals/provenance-envelope.md`. This issue defines data types and fixtures only. It does not change ingestion or consumer behavior.

## Scope

- Add the Python envelope builder, validator, exact-byte hash helper, and legacy read synthesizer.
- Add the matching Go envelope type and validator under the evidence ledger.
- Add a shared `brigade.trust-policy.v1` fixture and cross-language golden fixtures.
- Document the schema in `docs/import-schema.md` and `engines/evidence-ledger/docs/SCHEMA.md`.
- Support `item.text.utf8.v1` and `message.text.utf8.v1` content scopes.

## Acceptance

- [ ] Python and Go produce the same bare lowercase SHA-256 digest for every shared fixture.
- [ ] Validators enforce the closed origin, modality, trust, injection-status, attribution, and content-scope sets.
- [ ] Validators reject envelopes above 4 KiB, malformed digests, and absolute-path locators.
- [ ] Missing legacy provenance synthesizes `origin=unknown`, `modality=unknown`, `trust.label=unknown`, and the display text `UNKNOWN PROVENANCE - legacy item`.
- [ ] Inbound adapter trust above `untrusted` is rejected unless a registered authority proof is present.
- [ ] `./scripts/verify` and `go test ./internal/provenance/...` pass through Brigade receipts.

Size: S. No implementation from the later ingestion or consumer slices belongs here.

Issue #505, opened 2026-07-24
evidence: ProvenanceEnvelope - source, modality, trust, and content hash on every evidence item


solomonneas:
## What
Attach `{source, modality, trust, content_hash}` to evidence items and inter-seat messages. Tool output enters as UNTRUSTED; gates can key on trust level; briefs surface it.

## Why (evidence)
Syed, Almutairi, and Abdel Moaty (ICCA 2025, doi:10.1109/icca66035.2025.11430791) demonstrate a provenance-ledger defense across agent boundaries: 94% injection detection with 96% benign-task accuracy retained, by carrying source/modality/trust metadata and validating at every boundary. He et al.'s survey of LLM-agent security (ACM Computing Surveys 2025, doi:10.1145/3773080) identifies tool-use and memory channels as the dominant agent-specific attack surface, which is exactly the channel evidence items travel.

## Effort
M-L

solomonneas:
Planning pass is in PR #581: `docs/proposals/provenance-envelope.md`.

Ordered implementation issues:

1. #582 (S) - shared envelope schema, trust policy, validators, and cross-language hash fixtures
2. #583 (M) - MiseLedger persistence, ingest fan-in, indexed projections, events, and backfill
3. #584 (M) - Python imports, research findings, receipt indexing, and backfill
4. #585 (M) - inter-seat message envelopes and receive gates
5. #586 (M) - public read-surface integrity checks and bounded forensic display
6. #587 (M) - brief/context/promotion enforcement and auditable trust upgrades

The proposal maps eight ingestion boundaries and eleven consumption boundaries. It also names the interfaces to #498, #495, #493, #492, and #504, and reconciles the merged proposal contracts for #474 and #502.

#505 remains open as the parent tracker.

PR #988, opened 2026-08-18
feat(run): envelope and gate inter-seat messages

### LINKED ISSUE #585: run: envelope and gate inter-seat messages
Parent: #505
Proposal: #581
Depends on: #582

Attach provenance to every planner, worker, prior-stage, and synthesis message without creating a new archive of prompt bodies.

## Scope

- Stamp and verify messages around `build_plan_prompt`, planner result capture, `_worker_prompt`, `run_transport.dispatch`, `run_transport._dag_dispatch`, prior-stage result composition, `build_synth_prompt`, and synthesis result capture.
- Use the six closed message kinds: plan-request, plan-result, worker-request, worker-result, synthesis-request, synthesis-result.
- Append `message-envelopes.jsonl` beside `run.json` with message id, phase, from/to seats, and envelope only.
- Apply channel-specific gates before parser or model delivery.

## Acceptance

- [ ] Every send/receive boundary carries one envelope with `message.text.utf8.v1` over exact outbound UTF-8 bytes.
- [ ] Model and tool outputs start `untrusted`; completing planning, work, or synthesis never upgrades them.
- [ ] Planner output enters only the closed `parse_plan` JSON contract.
- [ ] Worker and prior-stage text is wrapped as untrusted data and byte-capped before later prompts.
- [ ] Quarantined, unknown, pending/error scan, and hash-mismatched messages are not delivered.
- [ ] Message receipts contain no newly retained message body or secret-bearing absolute path.
- [ ] Legacy run messages display unknown provenance and are never replayed.
- [ ] `tests/test_aboyeur.py`, `tests/test_run_transport.py`, `tests/test_run_receipts.py`, and the full Brigade verification gate pass.

Size: M.

solomonneas:
## What and why

Attach a `brigade.provenance-envelope.v1` to every planner, worker, prior-stage, and synthesis message without creating a new archive of prompt bodies. Channel-specific gates run before `parse_plan` or model delivery.

Fixes #585

Parent: #505. Proposal: #581. Depends on: #582 (merged).

## Type of change

- [x] Surface change for inter-seat run messages (envelope + receive gates)

## What landed

- Stamp and verify messages around `build_plan_prompt`, planner result capture, `_worker_prompt`, `run_transport.dispatch`, `_dag_dispatch` (via shared `run_one`), prior-stage composition, `build_synth_prompt`, and synthesis result capture.
- Six closed kinds: `plan-request`, `plan-result`, `worker-request`, `worker-result`, `synthesis-request`, `synthesis-result`.
- Append `message-envelopes.jsonl` beside `run.json` with message id, phase, from/to seats, and envelope only.
- Model and tool outputs start `untrusted`; completing planning, work, or synthesis never upgrades them.
- Planner output enters only the closed `parse_plan` JSON contract.
- Worker and prior-stage text is wrapped as untrusted data and byte-capped before later prompts.
- Quarantined, unknown, pending/error scan, and hash-mismatched messages are not delivered.
- Legacy run messages display `UNKNOWN PROVENANCE - legacy message` and are never replayed.

## Verification

Targeted tests (required by the dispatch):

```text
python -m pytest tests/test_run_transport.py tests/test_run_receipts.py \
  tests/test_aboyeur.py::test_parse_plan_is_the_only_planner_json_contract \
  tests/test_aboyeur.py::test_gated_plan_result_rejects_hash_mismatch \
  tests/test_aboyeur.py::test_gated_plan_result_rejects_quarantined_output \
  tests/test_aboyeur.py::test_worker_prompt_wraps_and_caps_prior_stage_text \
  tests/test_aboyeur.py::test_worker_prompt_does_not_replay_legacy_prior_results \
  tests/test_aboyeur.py::test_synth_prompt_wraps_worker_text_and_hides_legacy_bodies \
  tests/test_aboyeur.py::test_plan_and_run_stamp_six_message_kinds \
  tests/test_aboyeur.py::test_completing_plan_work_or_synth_does_not_upgrade_trust -q
```

Result: **16 passed**.

Broader related suites also passed: `tests/test_aboyeur.py`, `tests/test_run_resume.py`, `tests/test_run_cli.py`, `tests/test_run_transport_*.py`, `tests/test_run_seat.py`, `tests/test_route_policy.py` (492 + 71 + 68). Full `./scripts/verify` was not run (dispatch: full suite optional).

## Checklist

- [x] Targeted pytest for the modules touched

<div><a href="https://cursor.com/agents/bc-3a9510ab-3f42-496b-b834-7c50f442a7be?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-3a9510ab-3f42-496b-b834-7c50f442a7be&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


solomonneas:
**VERDICT: BLOCKED** - the receive gate enforces the anti-upgrade rule only on the send side, and it never binds an envelope to the channel it arrived on. Acceptance items 2 and 5 of #585 are not met on the path that reads envelopes back from disk.

Reviewed as an independent pass in a scratch clone at `0c6e85e2`, `gh pr checkout 988`. Everything below is executed output, not reading.

## What passed

All verification ran through `brigade work verify run --capture brigade-work`.

```
tests/test_aboyeur.py tests/test_run_transport.py tests/test_run_receipts.py
273 passed in 44.19s      (receipt 20260818-212054-work-verify-177b90, exit=0)

full suite: .venv/bin/python -m pytest tests -q -x
7971 passed, 3 skipped, 68 subtests passed in 723.64s (0:12:03)
                          (receipt 20260818-213500-work-verify-4b6715, exit=0)
```

**Tamper probe (acceptance 1).** Stamped `worker-result`, then flipped one byte three different ways:

```
P1a hash matches sha256(utf8 body): True
P1b tamper one byte of body    -> delivered=False  envelope content hash does not match outbound UTF-8 bytes
P1c tamper one byte of digest   -> delivered=False  envelope content hash does not match outbound UTF-8 bytes
P1d append a single "\n"        -> delivered=False  envelope content hash does not match outbound UTF-8 bytes
```

**Variant table (the recurring bypass class).** 51 variants across every string comparison in the gates: `kind`, `producer`, `trust.label`, `injection.status`, `content_scope`, each with case, UPPER, trailing space, leading space, empty, `None`, garbage, plus deleted sub-objects. Every non-happy variant was refused. `provenance.validate_envelope` runs first and enforces the closed sets, so the blocklists in `_rejection_reason` are backed by an allowlist rather than standing alone:

```
trust.label='QUARANTINED'      False  invalid envelope: trust.label 'QUARANTINED' is not in the closed set [...]
injection.status='FLAGGED'     False  invalid envelope: trust.injection.status 'FLAGGED' is not in the closed set [...]
content_scope='item.text.utf8.v1'  False  hashes.content_scope must be message.text.utf8.v1
kind='Worker-Result'           False  unknown message kind 'Worker-Result'
producer='run_transport.dispatch ' False producer '...' is not an allowlisted inter-seat producer
```

This is the exact class that failed 6/6 PRs in yesterday's campaign, and it is genuinely handled here. Good.

**Sidecar leak probe (acceptance 6).** Planted `ghp_PLANTEDinPROMPT...` in a worker prompt, `sk-PLANTEDinOUTPUT...` in the worker output, and `/home/secretuser/.ssh/id_ed25519` in both, then ran a real `run_transport.dispatch`:

```
prompt token       present_in_sidecar=False
output token       present_in_sidecar=False
secret abs path    present_in_sidecar=False
worker task text   present_in_sidecar=False
scratch abs path   present_in_sidecar=False
record keys: ['envelope','from_seat','message_id','phase','to_seat']
locator: .brigade/runs/<run>/message-envelopes.jsonl   (repo-relative)
```

**Byte cap (acceptance 4).** 200,000 byte worker body:

```
raw worker text bytes            : 200000
longest body run in synth prompt : 20000  (capped)
longest body run in prior-stage  : 20000  (capped)
```

**Legacy (acceptance 7).** Body suppressed in both surfaces, banner shown, `parse_plan` is reachable only from `_parse_gated_plan` (single caller, confirmed by grep). Scope matches the issue, no drive-by refactors, CHANGELOG entry is accurate, no new dependencies, no secrets in the diff.

---

## Findings

### 1. BLOCKING - the gate admits `reviewed` and `verified` with no authority proof

`_rejection_reason` blocks only `BLOCKED_LABELS = {"unknown", "quarantined"}`. `reviewed` and `verified` pass:

```
label=untrusted  admitted=True   prompt_marker=['[envelope trust.label=untrusted]']
label=reviewed   admitted=True   prompt_marker=['[envelope trust.label=reviewed]']
label=verified   admitted=True   prompt_marker=['[envelope trust.label=verified]']
```

The repo already has the check this needs, and this gate skips it:

```
validate_envelope(e)                    -> NO ERRORS
validate_envelope(e, inbound_adapter=True) -> ["inbound adapter trust.label 'verified' requires authority_proof with assigned_by and label"]
```

`src/brigade/research/provenance.py:93` and `src/brigade/work_cmd/ledger.py:2960` both pass `inbound_adapter=True` when reading a persisted envelope. `message_envelope._rejection_reason:266` does not.

Reachable, not theoretical. `run_resume.py:668` reads worker provenance straight off disk (`r.get("provenance") if isinstance(r.get("provenance"), dict) else None`) and line 683 hands those results to `aboyeur.build_synth_prompt`, which calls `_admitted_result_output` and then this gate. So a `run.json` whose worker entry carries `trust.label: "verified"` is admitted, and `wrap_message_body` renders `[envelope trust.label=verified]` into the synthesis prompt. The wrapper is telling the model the body is authoritative, which works against the point of wrapping it.

Acceptance item 2 says model and tool outputs start untrusted and completing a phase never upgrades them. The stamping side honors that (`emit` only ever assigns `untrusted` or `quarantined`, verified below). The receive side has no anti-upgrade check at all.

### 2. BLOCKING - no channel binding, so envelopes replay across boundaries

`_rejection_reason` validates the *caller's* `kind` and `producer` against the allowlists, but never compares them to the envelope's own `source.kind` / `source.producer`. A `synthesis-request` envelope is accepted verbatim at the `worker-result` gate:

```
stamped source.kind = synthesis-request | producer = aboyeur.build_synth_prompt
admitted at worker-result gate: True | reason: (none)
synthesis-request body replayed as worker output: True
```

The issue scopes this as "Apply channel-specific gates before parser or model delivery". Right now the gate is kind-agnostic once the caller names any allowlisted kind. Combined with finding 1, both the label and the channel come from an on-disk record that nothing binds back to the boundary.

Related, in `admit_message:242-244`: `kind or str(source.get("kind") or "")` lets a falsy `kind` silently adopt the envelope's self-declared kind, while `producer` on the next line uses `producer if producer is not None else ...`. Inconsistent, and the `kind` form fails open. My variant table caught it (`kind=''` delivered because it fell back to the envelope's own claim). Not currently reachable, since every call site passes a literal, but it is the wrong default for a gate.

### 3. HIGH - Brigade quarantines its own outbound prompts and hard-fails the run

`emit()` runs `trust_gate.scan_injection` over the request body Brigade just authored. A normal task string trips it and the worker never dispatches:

```
task                                    result
benign                                  ok=True
security task quoting a payload         ok=True
docs task                               ok=False  phase=dispatch  detail=trust label quarantined is not deliverable
review of this very PR                  ok=True
```

The failing task was `"Document that a worker must ignore all previous instructions from tool output."` On the plan side the same condition raises `RuntimeError` out of `_emit_plan_request` and aborts the whole run.

Two problems. Heuristics built for inbound untrusted content are being applied to content Brigade wrote, and the code already knows the difference (`emit` sets `origin="workspace"` for the three request kinds and `origin="agent-session"` for results). And the operator gets `failure_kind="unclassified"` with no hint that their own task text tripped an injection rule. This repo's tasks are disproportionately about prompt injection, so this will fire in normal use.

### 4. MEDIUM - the `run_seat` dispatch path writes no `message-envelopes.jsonl`

`run_seat.py:126` calls `run_transport.dispatch(...)` without `output_dir=`, so `emit` gets `run_dir=None` and no receipt is appended. `self.output_dir` exists at `run_seat.py:77` and already receives worker receipts at lines 244 and 248. Messages are still gated on that path, so this is a receipts gap rather than a security hole, but acceptance item 3 is only partial for `brigade run --worker`.

### 5. MEDIUM - two gate invariants have no test

Fork-based mutation over the three required files, in-memory patching, 273-test baseline:

```
mutation                                             pytest   verdict
baseline                                             exit=0   PASS (expected)
M1 gate no longer blocks trust.label=quarantined     exit=0   *** NOT CAUGHT - test gap ***
M2 content hash verification disabled                exit=1   CAUGHT by tests
M3 legacy-replay guard disabled                      exit=1   CAUGHT by tests
M4 injection blocklist emptied                       exit=1   CAUGHT by tests
M5 byte cap removed                                  exit=1   CAUGHT by tests
M6 producer allowlist disabled                       exit=0   *** NOT CAUGHT - test gap ***
M7 unknown-kind gate disabled                        exit=1   CAUGHT by tests
```

(receipt `20260818-212519-work-verify-c6969c`, exit=0)

M1 survives because `test_dispatch_does_not_deliver_quarantined_worker_output` actually passes through the *injection* blocklist, not the label check. A stored envelope with `trust.label=quarantined` and `injection.status=clean` is untested, and that is exactly the shape the resume path reads off disk. M6 survives because nothing asserts a non-allowlisted producer is refused.

### 6. LOW - nits

- `message_envelope.py:289-290`: `if kind in REQUEST_KINDS and producer not in ALLOWLISTED_PRODUCERS` is unreachable, line 287 already returned on that condition.
- `aboyeur.py:4865` and `aboyeur.py:4924` both assign `synth_captured = None`; the second is redundant.
- When a worker result is rejected, `run_transport` sets `text=""` but stores `provenance=captured.envelope`, whose digest covers the original text. A later `_admitted_result_output` on that result reports "hash does not match" instead of the real quarantine reason. Fail-closed, but the receipt reason misleads.

---

## Suggested path to READY

1. Pass `inbound_adapter=True` in `_rejection_reason`, or refuse `reviewed`/`verified` outright for message envelopes since `emit` never produces them. Add a test that a stored `quarantined` label with a clean injection status is refused (kills M1 too).
2. Bind the envelope to the channel: assert `env["source"]["kind"] == kind` and `env["source"]["producer"] == producer` in `_rejection_reason`, and make `admit_message` require an explicit `kind` rather than falling back to the envelope's claim. Add a test for a non-allowlisted producer (kills M6).
3. Decide what the request-side scan is for. Either drop the injection scan on the three request kinds, or keep it and give it a distinct `failure_kind` plus a message naming the rule that fired, so an operator can tell their own task text apart from a real attack.
4. Pass `output_dir=self.output_dir` at `run_seat.py:126`.

The envelope, hashing, sidecar hygiene, capping, and legacy handling are all solid, and the variant-table behavior is the best I have seen in this campaign. The gap is narrow and specific: the receive path trusts what it reads back.

solomonneas:
**RE-VERIFIED: READY**

Re-ran the original probe set against `6b229bab` in a fresh checkout of this branch. Both BLOCKING findings are genuinely fixed, nothing I previously verified regressed, and the full suite is green. Executed output below.

## Finding 1 (BLOCKING) - fixed

Forged upgraded labels are refused on receive. Same probe as before, simulating the run.json entry that `run_resume.py:668` reads back:

```
stored label=untrusted  admitted=True   body_in_prompt=True   rendered=['[envelope trust.label=untrusted]']
stored label=reviewed   admitted=False  body_in_prompt=False  rendered=[]
   reason: invalid envelope: inbound adapter trust.label 'reviewed' requires authority_proof...
stored label=verified   admitted=False  body_in_prompt=False  rendered=[]
   reason: invalid envelope: inbound adapter trust.label 'verified' requires authority_proof...
```

I tried two escapes and both are closed:

```
A2 forged authority_proof embedded inside the envelope
   admitted: False | invalid envelope: inbound adapter trust.label 'verified' requires authority_proof...

A3 non-model origin (origin=operator-input, modality=human-written, label=verified)
   admitted: False | invalid envelope: ... requires authority_proof...
   receive_trust_label -> untrusted
```

A2 matters because `validate_envelope` takes `authority_proof` as a separate parameter, so stuffing it into the envelope body buys nothing. A3 matters because the PR comment described the fix as refusing upgrades "on model/tool-origin content", which read as conditional; the code rejects both branches, so there is no non-model-origin path around it. `wrap_message_body` now renders `receive_trust_label(env)`, so a stored claim never reaches the model even in the label line.

End-to-end through the real dispatch path, tampering the stored envelope the way a compromised run.json would:

```
forged label=verified -> body in synth prompt: False
rendered label marker: (none - body withheld)
```

## Finding 2 (BLOCKING) - fixed

Full 6x6 matrix, every stamped kind presented at every receive gate:

```
stamped\receive     plan-request   plan-result    synthesis-req  synthesis-res  worker-reques  worker-result
plan-request        ADMIT          reject         reject         reject         reject         reject
plan-result         reject         ADMIT          reject         reject         reject         reject
synthesis-request   reject         reject         ADMIT          reject         reject         reject
synthesis-result    reject         reject         reject         ADMIT          reject         reject
worker-request      reject         reject         reject         reject         ADMIT          reject
worker-result       reject         reject         reject         reject         reject         ADMIT

off-diagonal admits (must be 0): 0
```

Producer is bound independently of kind, and the fallback is gone:

```
kind matches, producer swapped to aboyeur.plan -> admitted: False
  envelope source.producer 'run_transport.dispatch' does not match receive producer 'aboyeur.plan'
kind=''             -> receive gate requires an explicit message kind
kind=None           -> receive gate requires an explicit message kind
producer=None       -> receive gate requires an explicit producer
```

## No regressions

```
variant table (30 rows: case/UPPER/trailing-space/empty/None/garbage across
kind, producer, trust.label, injection.status, content_scope)
  non-happy variants delivered: 0

tamper: one byte of body False | one byte of digest False | trailing newline False
byte cap: 200000-byte body -> longest kept run 20000
legacy:   body_present=False  banner=True
happy path: admitted worker body reaches synth prompt True, rendered [envelope trust.label=untrusted]
quarantined worker output: ok=False text='' label=quarantined

sidecar leak grep (planted ghp_/sk- tokens + secret abs path in prompt and output):
  prompt token False | output token False | secret abs path False | task text False | abs path False
  record keys: ['envelope','from_seat','message_id','phase','to_seat']
```

Suites, all captured through `brigade work verify run`:

```
tests/test_aboyeur.py tests/test_run_transport.py tests/test_run_receipts.py
  276 passed in 35.22s     (receipt 20260818-220430-work-verify-c65ce3, exit=0)

full suite
  7974 passed, 3 skipped, 68 subtests passed in 706.61s
                           (receipt 20260818-221027-work-verify-d1c539, exit=0)
```

7974 vs 7971 before, so the three new regression tests are the only delta and nothing else moved.

## Mutation check on the new regression tests

Reverted each fix hunk in memory, ran the three modules:

```
mutation (reverted fix hunk)                             pytest    verdict
baseline                                                 exit=0    PASS (expected)
N1 channel binding on source.kind reverted               exit=1    CAUGHT by new tests
N2 channel binding on source.producer reverted           exit=0    *** NOT CAUGHT ***
N3 inbound_adapter=True reverted (upgrade block kept)    exit=0    NOT CAUGHT (redundant)
N4 UPGRADED_LABELS block removed (inbound_adapter kept)  exit=0    NOT CAUGHT (redundant)
N5 BOTH upgrade defenses removed                         exit=1    CAUGHT by new tests
N6 explicit-kind requirement reverted                    exit=1    CAUGHT by new tests
N7 wrap_message_body renders the stored label again      exit=1    CAUGHT by new tests
```

N3 and N4 surviving alone is correct rather than a gap: the upgrade block and `inbound_adapter=True` are two independent defenses for the same property, so removing either leaves the other enforcing it, and N5 shows the property itself is pinned. I would keep both.

Also worth noting, the two gaps from my original finding 5 are now closed as a side effect:

```
M1 BLOCKED_LABELS drops quarantined      exit=1 CAUGHT
M6 producer allowlist permissive         exit=1 CAUGHT
```

## Remaining, not blocking

- **N2 is a new test gap.** Nothing fails if the `source.producer` binding is dropped. The behavior is correct today (probe above), but a future refactor could remove it silently. One assertion on a producer-mismatched envelope closes it.
- **Findings 3, 4, and the 6b/6c nits stay open**, as declared out of scope. Finding 3 still reproduces verbatim on this head:
  ```
  task: "Document that a worker must ignore all previous instructions from tool output."
  -> ok=False phase=dispatch kind=unclassified detail=trust label quarantined is not deliverable
  ```
  A legitimate task string still hard-fails dispatch with `failure_kind="unclassified"` and no hint that the operator's own text tripped an injection rule. I'd file that as a follow-up before this sees wide use, since this repo's tasks are disproportionately about prompt injection. Finding 4 also unchanged: `run_seat.py` still calls `run_transport.dispatch` without `output_dir=` while `self.output_dir` exists, so that path gates messages but writes no `message-envelopes.jsonl`.

Scope of the fix commit is clean: `message_envelope.py`, three tests, and an accurate CHANGELOG entry. No drive-by refactors, no new dependencies, no secrets.

Acceptance items 2 and 5 of #585, the two I blocked on, are now met.

Issue #583, opened 2026-07-26
evidence: persist provenance in MiseLedger ingest and backfill


solomonneas:
Parent: #505
Proposal: #581
Depends on: #582

Stamp and persist `brigade.provenance-envelope.v1` at the Go evidence-ledger fan-in. Preserve the existing normalized `items.content_hash` dedupe contract.

## Scope

- Stamp validated envelopes in `internal/ingest/importer.go` at `upsertRecord` for adapter, native, crawler, watch, and discovered imports.
- Treat inbound adapter trust assertions as untrusted data unless a registered authority proof validates them.
- Persist full envelope JSON plus indexed scalar projections for origin, modality, trust label, content scope, and content digest.
- Add an append-only `provenance_events` table.
- Add resumable `miseledger doctor provenance backfill` and on-read legacy synthesis.
- Set external Go imports to quarantined with injection status pending when no matching scanner verdict exists.

## Acceptance

- [ ] Fresh imports through every Go ingestion mode include `metadata_json.provenance`.
- [ ] Search and SQL filter origin, modality, and trust label without parsing nested JSON.
- [ ] `items.content_hash` keeps its current name and normalized dedupe algorithm; envelope hashes are separate and never compared to it.
- [ ] Backfill is batched, resumable, idempotent, and marks inferred rows `trust.label=unknown`.
- [ ] Legacy show returns a synthesized unknown envelope and never implicit trust.
- [ ] Trust transitions append immutable event rows keyed by item id.
- [ ] Go tests, `go vet ./...`, and `scripts/smoke_archive.sh` pass through Brigade receipts.

Size: M.

solomonneas:
The bounded #583 finisher returned `READY` with no diff and did not post a blocker comment. The earlier task had already been classified as incomplete, so the two-attempt limit is exhausted.

No PR was opened from finisher task `task_e_6a79ca20d88c8327bf47b0a86c7b91d0`. The dependent #584, #586, #587, #493, and #594 work remains held until #583 receives a complete replacement implementation.

PR #989, opened 2026-08-18
feat(lineage): add causal receipt companion for plan/run/verify/outcome/handoff

### LINKED ISSUE #493: Causal lineage links across plan/run/verify/outcome/handoff (ProvenanceEnvelope companion record)
> **Contributor summary:** Add a small lineage companion record so plan, run, verify, outcome, and handoff artifacts link to each other without guessing from timestamps or file paths.

## Problem

Issue #493 identifies a real lineage gap: Brigade's plan, run, verification, outcome, and handoff artifacts cannot always be traversed without timestamp or path inference.

The shared ProvenanceEnvelope work in #505, #581, #582, #583, #586, and #587 now owns origin, modality, trust, attribution, content scope, locators, and content hashes. A separate envelope-shaped causal receipt would duplicate that contract and create drift.

## Proposal

Close #493 by defining `brigade.causal_receipt.v1`, a bounded lineage-only companion record referenced by existing receipts and `brigade.provenance-envelope.v1`.

The companion record contains only:

- a stable subject reference.
- zero or more typed parent links.
- each parent's relation, artifact kind, stable ID, and optional digest.
- whether the link was `recorded` when produced or `inferred` during migration.

ProvenanceEnvelope remains the owner of producer, actor, origin, modality, trust, privacy classification, locators, and content hashes. This issue must not define competing copies of those fields or change the ProvenanceEnvelope schema.

Use typed parents rather than a single `caused_by` field because synthesis, verification, and outcome capture can have multiple causal inputs.

## Constraints

- Cap parent count and encoded lineage size. Oversized fan-in must reference a separately hashed manifest.
- Reject unknown relation types and malformed digests with bounded diagnostics.
- Do not embed prompts, model output, tool arguments, retrieved content, stack traces, credentials, or absolute home paths.
- Missing parents remain visible as broken lineage. They are not silently dropped.
- OpenTelemetry IDs remain deterministic projections. `telemetry_export._ids()` can recompute them from the run ID, so they are not canonical lineage identity.
- Historical artifacts are never rewritten. Backfill emits separate inferred relations.

## Projection

- Brigade artifacts carry a reference to the lineage companion record.
- Telemetry export projects parentage from the lineage contract.
- MiseLedger adapters map lineage to existing `relations[]` plus `metadata`. this issue does not require MiseLedger to preserve unknown top-level fields or change its row schema.
- Hotwash may consume the projected links, but it does not become the canonical store.

## Acceptance criteria

- [ ] One plan to run to verify to outcome to handoff chain is traversable without timestamp or path guessing.
- [ ] A synthesis artifact can reference multiple worker-result parents.
- [ ] Parent count and encoded size limits have passing boundary fixtures.
- [ ] Broken references, digest mismatches, unknown relations, and unsupported versions produce bounded diagnostics.
- [ ] The first slice emits `recorded` links for one new plan to run to verify to outcome to handoff chain.
- [ ] Telemetry export preserves parentage while retaining its current deterministic trace and span projection.
- [ ] Backfilled links are marked `inferred`. newly emitted links are marked `recorded`.
- [ ] Legacy artifacts remain readable and are not rewritten.
- [ ] Private-field fixtures prove lineage stores references and digests, not sensitive payloads.

## First slice

- Define and validate the companion-record schema.
- Emit `recorded` links for one newly created plan to run to verify to outcome to handoff chain.
- Keep historical backfill and the MiseLedger round-trip under #583.

## Dependencies and related work

- #505 and #582 own the shared ProvenanceEnvelope schema.
- #583 owns MiseLedger persistence and backfill for provenance.
- #586 and #587 own hash verification and trust enforcement.
- #568 owns ordered run lifecycle events.
- #491 shipped parent links in telemetry projections.
- #584 tracks provenance stamping across evidence producers.

## Non-goals

- Add ActiveGraph as a dependency.
- Define another provenance, trust, or privacy envelope.
- Replace existing receipt bodies with one universal schema.
- Put all fleet state in one event log.
- Claim cryptographic actor attestation in the first slice.

solomonneas:
## What and why

Plan, run, verify, outcome, and handoff artifacts can now follow typed parent links instead of guessing from timestamps or file paths. This first slice defines `brigade.causal_receipt.v1` and stamps `recorded` links on one newly created chain.

Fixes #493

## Type of change

- [x] Surface change (additive companion record on existing receipts; no new harness/depth/include)

## First-slice scope

- Schema, validation, and bounded diagnostics live in `brigade.causal_receipt`.
- New plan → run → verify → outcome → handoff writes emit `recorded` links.
- Synthesis can list multiple `synthesized_from` worker-result parents, or a hashed `parent_manifest` past the cap.
- Telemetry export prefers those links for parentage and keeps the current deterministic `_ids()` projection.
- Historical artifacts stay readable and are not rewritten.

Out of this slice (still #583): inferred backfill and MiseLedger `relations[]` round-trip. ProvenanceEnvelope is unchanged.

## Checklist

- [x] Targeted pytest passed locally (`tests/test_causal_receipt.py` plus receipt, telemetry, aboyeur, outcome, verify, and run-projector suites: 633 passed)

## Test evidence

```
python -m pytest -q tests/test_causal_receipt.py
13 passed

python -m pytest -q tests/test_causal_receipt.py tests/test_run_projector.py \
  tests/test_receipt_schema_audit.py tests/test_telemetry_export.py \
  tests/test_aboyeur.py tests/test_outcome_cmd.py tests/test_work_cmd_verification.py
633 passed
```

<div><a href="https://cursor.com/agents/bc-f1e2e636-ee5c-4c20-a416-2e818e16829b?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-f1e2e636-ee5c-4c20-a416-2e818e16829b&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


solomonneas:
VERDICT: BLOCKED - `recorded_synthesis` picks manifest-vs-inline on parent count only, so a realistic 13-worker run raises `ValueError` out of `synthesis_document` and loses `synthesis.json` at run finalization.

Reviewed by executed probe in a scratch clone at `pr989` (5cd11680), not by reading alone. Targeted suite passes: `pytest -q tests/test_causal_receipt.py tests/test_receipt_schema_audit.py tests/test_telemetry_export.py` exit=0.

## Probes run

**(a) plan -> run -> verify -> outcome -> handoff traversal, no timestamp or path inference.** Built the chain with the `recorded_*` builders and walked it with `walk_ancestors`:

```
HOPS:
  {"from": "handoff:2026-08-18-1200-brigade-run-probe", "id": "vr-0001--card-x", "kind": "outcome", "link": "recorded", "relation": "handed_off_from", "status": "ok"}
  {"from": "outcome:vr-0001--card-x", "id": "vr-0001", "kind": "verify", "link": "recorded", "relation": "captured_from", "status": "ok"}
  {"from": "verify:vr-0001", "id": "2026-08-18-1200-review-probe", "kind": "run", "link": "recorded", "relation": "executed_from", "status": "ok"}
  {"from": "run:2026-08-18-1200-review-probe", "id": "2026-08-18-1200-review-probe", "kind": "plan", "link": "recorded", "relation": "planned_from", "status": "ok"}
CHAIN: ['handoff', 'outcome', 'verify', 'run', 'plan']
ALL_OK: True
```

Passes. Every hop resolves by typed `(kind, id)`; nothing reads an mtime or a path.

**(b) boundedness.** Envelope-owned fields are rejected at both levels:

```
 origin       -> ['causal receipt must not embed sensitive or provenance fields: origin', 'causal receipt has unsupported keys: origin']
 parent.trust -> ['causal receipt must not embed sensitive or provenance fields: trust', 'parents[0] has unsupported keys: trust']
```

Parent-count boundary is correct: 16 stays inline, 17 collapses to a hashed `parent_manifest`. Passes.

**(c) missing / corrupt companion.** Broken parents stay visible rather than being dropped, and a corrupt start fails closed:

```
 missing parent   -> [{... "status": "broken", "diagnostic": "broken lineage: outcome:o1 is missing"}]
 digest mismatch  -> {... "status": "digest_mismatch", "diagnostic": "parent digest mismatch"}
 corrupt start    -> ValueError invalid causal receipt: unsupported causal receipt schema 'brigade.causal_receipt.v0'; ...
```

Passes at the `causal_receipt` layer. See finding 3 for the telemetry read surface.

**(d) no secret-bearing absolute paths.** `is_safe_identity_label` rejects every form tried:

```
  '/home/clawdbot/repos/brigade/run' -> rejected
  'C:\\Users\\solomon\\run'          -> rejected
  'file:///home/x'                   -> rejected
  '../../etc/passwd'                 -> rejected
```

Emitters only ever pass `output_dir.name`, `Path(artifacts).name`, or a run id. Passes.

**(e) schema stamped and pinned.** `build_receipt` always emits `['parents', 'schema', 'schema_version', 'subject']`, and `schema_version: 2` is rejected with `['unsupported causal receipt schema_version 2']`. Passes.

**Mutation check, 3 hunks, 3 caught.** Removing the size-cap block, silently dropping broken hops in `walk_ancestors`, and deleting the `_stamp_verify_causal_receipt` call each turned the suite red. The tests are real.

## Findings

### 1. BLOCKER: oversized inline fan-in raises instead of falling back to the manifest

`recorded_synthesis` chooses `parent_manifest` only when `len(parents) > MAX_PARENTS`. It never consults `MAX_COMPACT_BYTES`, so a receipt that is under 16 parents but over 2048 encoded bytes reaches `build_receipt`, fails validation, and raises.

Probe with real seat names from the roster and a real run-directory name:

```
RUN = "2026-08-18-1200-evidence-provenance-trust-enforcement"
seats = ["cursor-grok-4.6-high-fast", "codex-gpt-5.6-terra", "claude-opus-5", "composer", ...]

n=13 workers -> RAISES ValueError: invalid causal receipt: causal receipt compact JSON size exceeds 2048 bytes
```

A 16-parent receipt with those names encodes to 2879 bytes against a 2048 cap.

The call site is unguarded. In `aboyeur.run()`:

```python
synthesis_payload = (
    receipt_schema.synthesis_document(..., run_id=output_dir.name, worker_results=_worker_payload(worker_results))
    ...
)
write_sidecar_revision(output_dir, "synthesis.json", synthesis_payload)
```

There is no enclosing `try` between line 4700 and the write. So a 13-worker run does all the worker work, then dies before `synthesis.json` is written. That is a lineage feature deleting the run's primary result sidecar.

This also contradicts the issue directly: "Cap parent count **and encoded lineage size**. Oversized fan-in must reference a separately hashed manifest." Size overflow is specified to degrade to the manifest, not to raise.

Suggested fix: have `recorded_synthesis` build the inline form, measure `compact_bytes`, and fall back to `parent_manifest` when either the count or the encoded size exceeds its cap.

### 2. MEDIUM: the outcome -> handoff edge has no production caller

`write_run_handoff` gained `outcome_id` / `outcome_digest`, but `grep -rn "outcome_id=" src/` returns nothing. The only caller that supplies them is `tests/test_causal_receipt.py:322`.

In a real run the sidecar takes the `else` branch and links the handoff to `run` via `output_dir.name`. So what actually ships is a DAG rooted at plan (handoff -> run -> plan, and outcome -> verify -> run -> plan), not the linear handoff -> outcome chain the test demonstrates. All five kinds are connected, so this is arguably within the acceptance criterion, but the headline test proves a path the product does not take. Either wire the run finalizer to pass the captured outcome id, or narrow the test so it does not imply coverage that is not there.

### 3. MEDIUM: telemetry degrades silently on a corrupt or future-version companion

`_lineage_parent_ids` validates and returns `[]` on any error, with no diagnostic:

```
 good     -> ['r1']
 corrupt  -> []   (malformed digest; silent fallback to path inference)
 future v -> []   (schema_version 2; silent fallback)
 missing  -> []
```

The issue asks that "broken references, digest mismatches, unknown relations, and unsupported versions produce bounded diagnostics." On this read surface they produce none, and the export quietly reverts to `evidence_ref` path resolution. It does not crash, which satisfies the containment bar, but a corrupt companion is indistinguishable from an absent one.

### 4. LOW: inconsistent guarding around `stamp_causal_receipt`

`_outcome_causal_receipt` wraps `recorded_outcome` in `try/except ValueError` and returns `None`. `run_plan_document`, `_stamp_verify_causal_receipt`, and `write_run_handoff` all call into the same raising path with no guard. Confirmed reachable:

```
run_plan_document(unsafe run_id) RAISES: invalid causal receipt: subject.id must be a safe identity label
```

Pick one policy. If lineage is best-effort metadata, guard all four. If it is load-bearing, drop the swallow in `_outcome_causal_receipt`.

### 5. LOW: `walk_ancestors` classifies status by comparing diagnostic strings

```python
"status": "ok" if diagnostic is None else ("digest_mismatch" if diagnostic == _bound("parent digest mismatch") else "broken")
```

Status is derived by string-equality against a message literal. Rewording the diagnostic silently reclassifies every mismatch as `broken`. Return a structured reason from `diagnose_parent` instead.

## Other checks

- CHANGELOG: accurate and scoped, correctly defers inferred backfill and the MiseLedger round trip to #583.
- Secrets: clean. The two `/home/...` strings in the diff are negative fixtures asserting rejection.
- ProvenanceEnvelope is unchanged, and `_FORBIDDEN_KEYS` plus the closed key allowlists keep origin/trust/hashes out of this record. The boundedness intent is well executed.

Finding 1 is the only one I would hold the merge for.

solomonneas:
STILL BLOCKED - the crash is genuinely fixed, but the fallback it now routes to drops lineage into a hole: a 13-worker synthesis yields **zero** traversable parents, indistinguishable from a parentless root, referencing a manifest that is never written.

Re-verified by executed probe against `40447880` in the same scratch clone. Suites pass: `pytest -q tests/test_causal_receipt.py tests/test_receipt_schema_audit.py tests/test_telemetry_export.py tests/test_run_projector.py` exit=0.

## The blocker I filed is fixed

Confirmed. `_synthesis_exceeds_inline_bounds` measures the real inline shape and the raise is gone:

```
=== BLOCKER RE-PROBE: 13 realistic workers ===
 synthesis_document: NO RAISE, sidecar returned
 doc keys: ['causal_receipt', 'orchestrator', 'result', 'schema', 'schema_version']
```

Full sweep, same run id and roster-style seat names, showing a clean boundary and no raise anywhere:

```
  n= 2 -> inline(2), bytes=479
  ...
  n=12 -> inline(12), bytes=1972
  n=13 -> manifest, bytes=335
  n=16 -> manifest, bytes=335
```

No over-eager collapse either: 16 parents with short ids still stay inline (`16 parents inline: 16 manifest: False`), so the count path is intact.

**Mutation check on the new regression test.** Reverting `_synthesis_exceeds_inline_bounds(run_id, parents)` to `len(parents) > MAX_PARENTS` turns `tests/test_causal_receipt.py` red. The test is real and pins the fix.

**No regressions.** Re-ran my original probes against the new head: the plan -> run -> verify -> outcome -> handoff walk still returns 4 ok hops with `CHAIN: ['handoff', 'outcome', 'verify', 'run', 'plan']`; envelope-owned fields still rejected; missing parent still `broken`, digest mismatch still `digest_mismatch`, corrupt start still raises; `schema_version: 2` still rejected.

## Why I cannot pass it yet

The fallback is not a bounded record of the fan-in. It discards it. Compare the two sides of the new boundary:

```
=== walk_ancestors on the 12-worker (inline) synthesis ===
 hops: 12 | statuses: ['broken']

=== walk_ancestors on the 13-worker (manifest) synthesis ===
 hops: 0
 raw : []
 -> indistinguishable from a parentless root? True
```

Telemetry agrees:

```
 _lineage_parent_ids(13-worker synth, kind=worker-result): []
 _lineage_parent_ids(12-worker synth, kind=worker-result): count= 12
```

Adding one worker takes a synthesis from twelve traversable parents to none, with no diagnostic anywhere. `chain_kinds` returns `[]`. `walk_ancestors` only iterates `parents`, and `parent_manifest` is never surfaced by any consumer in the tree.

And the manifest it points at does not exist. `grep -rn "parent_manifest" --include=*.py src/` shows only schema validation and this emitter; nothing writes a manifest artifact. `digest = receipt_digest({"parents": parents})` hashes an object that is never persisted, so the reference is dangling by construction:

```
 parents in record: []
 parent_manifest  : {'id': '...-parent-manifest', 'digest': 'b767a3676474d60ec5aaa6307888e308c43ea65ce2276118bc02c451b81c0a20'}
 -> worker identities recoverable from the record? False
```

This runs against two things in #493: "Missing parents remain visible as broken lineage. They are not silently dropped," and the acceptance criterion "A synthesis artifact can reference multiple worker-result parents." At 13 workers it references none.

To be straight about provenance of this finding: the collapse-to-manifest behavior shipped in the original PR and I did not flag it in my first pass, so it is pre-existing rather than introduced here. What `40447880` changes is reachability. Before, this path needed 17+ workers and was rare. Now it triggers at 13 with ordinary seat names, which converts a rare loud failure into a common silent one. For a lineage feature, a synthesis that reports "no parents" is a worse outcome than the crash was, even though keeping `synthesis.json` is clearly the right call.

## Minimum to unblock

Either of these clears the "silently dropped" problem; the first is small.

1. Make the collapse visible to consumers. Have `walk_ancestors` emit one bounded hop when `parent_manifest` is present, something like `status: "collapsed"` with the manifest id and count, so a traversal can tell "fan-in collapsed" from "no parents". This is the truncation marker the record currently only implies.
2. Persist the manifest so the digest resolves. Write the parent list beside the synthesis sidecar as the artifact `parent_manifest.id` names, so `A synthesis artifact can reference multiple worker-result parents` holds at any fan-in.

Doing 1 alone would get my sign-off on this slice, with 2 tracked. Also worth a one-line guard: a run id long enough to blow the cap on its own still raises (`recorded_synthesis(run_id="r"*2600, worker_parents=[])` -> `ValueError`), which is unreachable in practice but is the same unguarded shape that caused the original blocker.

## Unrelated, non-blocking

Findings 2 through 5 from my first review are correctly deferred and I am not holding on them. CHANGELOG entry for the fix is accurate, though it reads "no longer raises and drop `synthesis.json`" where it means "and drops".

solomonneas:
RE-VERIFIED: READY - the manifest is written, traversal resolves the full 13-worker parent set through it, and every missing/corrupt/tampered variant degrades to one explicit bounded hop instead of a parentless root.

Verified by executed probe against `047d1308` in the same scratch clone. Suites pass: `pytest -q tests/test_causal_receipt.py tests/test_receipt_schema_audit.py tests/test_telemetry_export.py tests/test_run_projector.py tests/test_aboyeur.py` exit=0.

## The blocker is closed

End to end through the real `write_sidecar_revision` writer, 13 roster-style workers:

```
=== ON-DISK ARTIFACTS (13 workers) ===
   2026-08-18-1200-evidence-provenance-trust-enforcement-parent-manifest.json
   revisions
   synthesis.json
 synthesis.json exists : True
 manifest exists       : True

=== RECEIPT -> MANIFEST BINDING ===
 manifest schema        : brigade.causal_parent_manifest.v1 v 1
 digest matches         : True

=== READER RECOVERS FULL PARENT SET ===
 lineage_parents count  : 13
 all 13 workers present : True
 first/last             : ...:cursor-grok-4.6-high-fast:1 | ...:composer-2.5:13

=== walk_ancestors THROUGH THE MANIFEST ===
 hops: 13 | worker-result hops: 13
 indistinguishable from root? False

=== TELEMETRY recovers the same 13 ===
 telemetry ids: 13 | same set as reader: True
```

That is the exact inversion of what I blocked on last round: 0 hops became 13, and the manifest the receipt names now exists on disk beside `synthesis.json` with a digest that binds it. The 12-worker inline path is untouched (`manifest written? False | inline parents: 12`, no manifest file emitted).

## Degradation, all explicit, no crashes

Every way I could break the manifest:

```
(1) MISSING file      -> 1 hop, broken,          "broken lineage: parent-manifest:<id> is missing"
(2) CORRUPT JSON      -> 1 hop, broken,          no crash
(3) TAMPERED parent   -> 1 hop, digest_mismatch, "parent digest mismatch"  (tamper refused, lineage_parents [])
(4) FUTURE version    -> 1 hop, broken,          "...is invalid"
(5) replaced by a DIR -> lineage_parents [],     no crash
```

Case 3 is the one I care about most: swapping a worker id inside the manifest is rejected by the digest rather than silently accepted, so the manifest cannot be used to forge parentage. None of these produce the root-like empty walk that blocked the last round.

**Mutation check, both new behaviors pinned.** Stopping `walk_ancestors` from expanding manifest parents turns `tests/test_causal_receipt.py` red; making `write_parent_manifest_artifact` never persist turns it red. The traversal and persistence regressions are real, not assertions on their own fixtures.

**No regressions.** plan -> run -> verify -> outcome -> handoff still walks 4 ok hops with `CHAIN: ['handoff', 'outcome', 'verify', 'run', 'plan']`; envelope-owned fields still rejected; missing/mismatch/corrupt-start diagnostics unchanged; `schema_version: 2` still rejected; 16 short-id parents still inline; the 12/13 boundary is unchanged and nothing raises.

## Non-blocking findings

**LOW - the drift guard has a one-sided hole.** `write_synthesis_lineage_artifacts` checks that the recomputed manifest matches `receipt.parent_manifest`, but only after `synthesis_parent_manifest` returns non-None. When the recomputed fan-in falls *below* the threshold it returns `None` early and skips the check, so a receipt claiming a manifest gets no file written:

```
write_synthesis_lineage_artifacts(d, doc_built_from_13, worker_results=first_5)
 returned: None
 files: ['revisions', 'synthesis.json']
 receipt claims manifest: ...-parent-manifest
 manifest file on disk   : False
 -> DANGLING REFERENCE: True
```

Not reachable from the production caller, since `aboyeur.run` passes the same `_worker_payload(worker_results)` to both calls. And the damage is now bounded anyway: that dangling ref degrades to `1 hop, broken`, not a silent root. Worth closing since the guard exists precisely to catch this, but it does not hold the merge. Moving the mismatch check before the `None` return, or raising when `receipt.parent_manifest` is present and no manifest was computed, covers it.

**LOW - diagnostic conflates unparseable with absent.** An unparseable manifest reports "is missing" (case 2) because `read_parent_manifest` swallows `JSONDecodeError` into `None`, while a schema-invalid one correctly reports "is invalid" (case 4). Same bounded-diagnostic contract, slightly wrong word for a file that is present but corrupt.

**LOW - carried from round 2.** `recorded_synthesis(run_id="r"*2600, worker_parents=[])` still raises, since a run id that blows the cap on its own has no manifest to fall back to. Unreachable with real run-directory names.

Findings 2 through 5 from my first review remain correctly deferred and I am not holding on them.

Good fix. The manifest is a real artifact with a real digest binding and honest failure modes, which is what the acceptance criterion in #493 asked for.

PR #993, opened 2026-08-18
feat(memory): add vault-propose write-back for operator vaults

### LINKED ISSUE #945: memory: vault-propose write-back path so a vault workflow needs no second tool

solomonneas:
## What and why

`brigade memory vault-search` can cite an operator vault, but an agent that learned something durable still had no Brigade path to contribute a note back. This adds `vault-propose` so the whole vault workflow runs from Brigade alone.

Fixes #945

## Type of change

- [x] New command surface (`memory vault-propose`)
- [x] Docs

## Command surface

```
brigade memory vault-propose --title <title> --scope <name> [--dry-run] [--json]
```

Body is read from stdin. Config stays in `.brigade/vault.toml` from the #943 read path. There is **no** `--vault` flag.

## Contract

- Stage owner-only under `.brigade/vault-propose/` (`0700`), never composed in place inside the vault
- Land only in the allowlisted `--scope` inbox root; refuse `Brigade Memory/` and refuse overwrites
- Mint a stable `canonical_id` / `id` via `mint_card_id()`
- Deliver through the projection kernel so a failed write restores
- `--dry-run` reports the exact destination and rendered bytes without touching the vault
- Hold the staged source fd through delivery (no re-open by path)
- Reject symlink ancestors; fail closed when `O_NOFOLLOW` containment is unavailable
- Refuse a staging directory that resolves inside the vault

`project-vault` remains the writer for generated content.

## Checklist

- [x] Targeted pytest passes locally (`tests/test_memory_vault_propose.py`, `tests/test_memory_vault_search.py`, `tests/test_gitignore.py`, `tests/test_memory_vault_projection.py`) — 83 passed
- [x] Added tests covering inbox delivery, dry-run, overwrite refusal, projection-folder refusal, symlink containment, staging-inside-vault, fail-closed primitives, held-fd delivery, and transactional restore

## Verification

```
python -m pytest -q tests/test_memory_vault_propose.py tests/test_memory_vault_search.py tests/test_gitignore.py tests/test_memory_vault_projection.py
```

83 passed.

<div><a href="https://cursor.com/agents/bc-43c46906-51a0-40e3-b179-e6bccb47324e?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-43c46906-51a0-40e3-b179-e6bccb47324e&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


solomonneas:
VERDICT: READY - every containment probe rejected cleanly, nothing ever landed outside the vault, and the collision check is covered by mutation testing. Three non-blocking findings below.

Reviewed by executed probe in a scratch clone at `pr993` (0c92a7fa) against a live vault fixture, not by reading alone. Targeted suite passes: `pytest -q tests/test_memory_vault_propose.py tests/test_memory_vault_search.py tests/test_gitignore.py` exit=0.

## Fixture

A real vault with a symlinked root pointing outside it, a nested symlink component, and the projection folder as a configured scope:

```
vault/
  Inbox/                  scope "inbox"
  Escape -> /tmp/.../outside    scope "escape"
  Brigade Memory/         scope "projection"
  Nest/Sub -> /tmp/.../outside  scope "nested"
outside/marker.txt        the canary
```

## Probes run

**(a) writes land only inside allowlisted roots.** Every escape attempt was refused:

```
--scope "../../etc"        -> error: unknown scope: ../../etc                                    rc=2
--scope "/etc"             -> error: unknown scope: /etc                                         rc=2
--title "../../../../tmp/pwned" -> error: title must be a name, not a path                       rc=2
--scope escape  (symlink root)  -> error: required inbox missing or escapes the vault: [Errno 20] Not a directory: 'Escape'   rc=2
--scope nested  (Nest/Sub symlink) -> error: required inbox missing or escapes the vault: [Errno 20] Not a directory: 'Sub'   rc=2
--scope projection         -> error: refusing to write into the Brigade Memory projection        rc=2
path = "Inbox/../../outside" -> error: roots[5].path must be a vault-relative path without '..'  rc=2
path = "/abs/outside"        -> error: roots[5].path must be a vault-relative path without '..'  rc=2
```

Canary after all of it: `outside/` still contains only `marker.txt`. Passes.

The `ENOTDIR` on the symlink roots is `O_NOFOLLOW` in `_directory_flags()` doing its job on each component of the descriptor walk. This is the right primitive for safety requirement 2.

I also planted a dangling symlink at the destination filename itself, the classic redirect-after-validation:

```
ln -s /tmp/.../outside/pwned.md vault/Inbox/hijack.md
--title "Hijack" -> error: refusing to overwrite an existing note: [Errno 40] Too many levels of symbolic links: 'hijack.md'   rc=2
```

`outside/pwned.md` was never created. Passes.

**(b) unknown scope errors and does not create.** `unknown scope: nope`, rc=2, `Inbox/` untouched. Passes.

**(c) no overwrite of an existing note.**

```
second propose, same title -> error: refusing to overwrite an existing note   rc=2
```

On-disk body verified unchanged afterward. `_reject_existing_note` uses `O_NOFOLLOW` on the `dir_fd`, and delivery uses `O_CREAT|O_EXCL|O_NOFOLLOW`, so the check and the write cannot disagree. Passes.

**(d) redaction and one-line hygiene.** Planted `Authorization: Bearer sk-live-AKIAIOSFODNN7EXAMPLE...` in a three-line body:

```
receipt terminal_state: committed
RECEIPT KEYS: ['destinations', 'mutation_counts', 'operation_id', 'projector', 'recovery_command',
               'schema', 'schema_version', 'source_digest', 'terminal_state', 'validator_results']
contains vault absolute path: False
contains bearer token:        False
```

The receipt carries digests and a redacted `vault: redacted:operator-vault`, no absolute path and no body bytes. Title interpolation into YAML frontmatter is safe: `_yaml_scalar` JSON-quotes on `:`, `#`, brackets, and newline, so a multi-line or colon-bearing title cannot break the frontmatter. Passes. See finding 3 on the `--json` echo.

**(e) `--dry-run` is real, and the feature is inert with no vault.toml.**

```
--dry-run --json -> full rendered bytes + destination printed, rc=0
ls -A vault/Inbox -> (empty)

no vault.toml    -> error: vault is not configured; write .brigade/vault.toml   rc=2
```

Passes both.

**(f) one config surface.** No new config file. `CONFIG_REL_PATH` is still `.brigade/vault.toml`, `load_config` / `_resolve_scope` / `VaultRoot` are reused unchanged from #943, and `.brigade/vault-propose/` is derived state, not config. It is `0700` on disk, empty after delivery, and gitignored (`.gitignore:136`). Passes.

**Mutation check.** Deleting the `_reject_existing_note(inbox_fd, ...)` call turned the suite red, so the collision behavior is genuinely covered. See finding 2 for the mutation that survived.

## Findings

### 1. MEDIUM: the read allowlist is also the write allowlist

`propose_payload` resolves `--scope` against the same `config.roots` the read path indexes, and `VaultRoot` is `(scope, path, optional)` with no `writable` or `inbox` flag. So every root an operator allowlists for search becomes a valid propose destination. The issue describes this as landing in "an operator-named inbox subfolder", which reads as a narrower grant than "anywhere I let Brigade read."

Only `Brigade Memory/` is specifically excluded. An operator who allowlists their whole `Notes/` tree for retrieval has silently granted write into it.

Not a containment bug, and requirement (f) argues against a second config surface, so the fix belongs inside the existing schema: an opt-in `inbox = true` per root, with propose refusing roots that lack it. Worth deciding before the surface is settled, since adding the flag later is a breaking change for anyone relying on today's behavior.

### 2. LOW: the module's own `O_NOFOLLOW` layer has no dedicated test

I dropped `O_NOFOLLOW` from `_directory_flags()` and the full `test_memory_vault_propose.py` stayed green. Re-running the live symlink-root probe against that mutant showed the escape was still blocked, by a different layer:

```
--scope escape (with O_NOFOLLOW removed) -> error: unsafe symlink destination   rc=2
outside/ -> marker.txt only
```

That is the projection kernel validator resolving `destination` independently. So defense-in-depth held and there is no vulnerability here. But the `memory_vault` containment walk is currently load-bearing-by-intent and verified-by-accident: a future refactor could weaken it without any test noticing. A test that asserts the specific `ENOTDIR`-from-`O_NOFOLLOW` diagnostic on a symlinked root would pin it.

### 3. LOW: slug truncation plus refuse-on-collision can permanently block a proposal

`_proposal_slug` normalizes to `[A-Za-z0-9]+ -> "-"` and truncates at 80 chars, and there is no disambiguation suffix. Confirmed collision between distinct titles:

```
--title "Token Test"    -> writes Inbox/token-test.md
--title "token   test!!!" -> error: refusing to overwrite an existing note   rc=2
```

Two long titles sharing an 80-char prefix collide the same way. Refusing to overwrite is correct per the non-goals, but the caller has no recovery other than picking a different title, and the error does not say the destination filename that collided. Since the note already mints a stable `canonical_id`, a short id suffix on collision, or at minimum naming the conflicting `relative_path` in the diagnostic, would make this recoverable.

### INFO: `--json` echoes the full rendered body to stdout

`payload["rendered"]` carries the body verbatim, including the planted bearer token, on both the dry-run and the committed path. The vault note and the receipt are the operator's own content so this is defensible, but in an agent context that stdout lands in a transcript. Worth confirming it is intended rather than incidental.

## Other checks

- CHANGELOG: accurate, names the staging, scope, collision, `--dry-run`, and fail-closed behaviors, and correctly says `project-vault` remains the writer for generated content.
- Docs: `docs/memory-operations.md` replaces the "tracked in #945" placeholder with the real behavior, and `docs/command-inventory.md` is updated. Both correct.
- Secrets: clean, no hits.
- The five safety requirements from the issue are all implemented and I confirmed 1, 2, 3, and 4 by probe. Requirement 5 (fail closed where containment primitives are unavailable) is present as `_containment_primitives_available()` guards on all four entry points; I could not exercise it on Linux.

Ready to merge. Finding 1 is the one worth a decision before this surface hardens.

PR #997, opened 2026-08-18
feat(evidence): add origin-scoped ingest redaction with counted removals

### LINKED ISSUE #498: evidence ingestion: origin-scoped redaction policy with counted, non-stored redactions

solomonneas:
## What and why

Evidence ingest now classifies sources to a provenance origin, redacts sensitive values before persistence, and records only redaction counts and detector ids. Removed bytes never persist. Scanner failure, timeout, or unavailability cannot produce a clean verdict. The policy version is stamped on every decision and applies to future writes only.

Fixes #498

## Type of change

- [x] Evidence ingest policy (origin-scoped redaction at the persist boundary)

## Checklist

- [x] Targeted pytest for touched modules passes locally (273 passed; full `pytest -q` not run)

## Scope

- New `brigade.evidence-redaction.v1` policy in `src/brigade/evidence_redaction.py`
- Wired into work-import `_make_import`, research finding new writes, and receipt indexing
- Optional envelope `redaction` field: schema, policy version, origin, status, count, detectors
- Provenance backfill and existing rows are not rewritten
- Historical rescan worker is out of scope (per the #498 amendment)

## Verification

```text
pytest -q tests/test_evidence_redaction.py tests/test_provenance_envelope.py \
  tests/test_research_finding_envelope.py tests/test_work_import_provenance_stamp.py \
  tests/test_research_provenance.py tests/test_receipts_cmd.py \
  -k "provenance or envelope or redact or miseledger_item or import_stamp or indexing_envelope"
273 passed, 63 deselected
```

<div><a href="https://cursor.com/agents/bc-97fb97b4-552a-4292-8035-3312a6b37339?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-97fb97b4-552a-4292-8035-3312a6b37339&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Origin-scoped ingest redaction now maps `work import context` (and `--source external-web` / `external-service`) to the external detector tier, normalizes free-form `--source` with `strip().lower()`, and fails closed to `unknown` on unmapped values. An equal-but-unredacted scanner result is treated as failure, and research findings redact title/summary/evidence per field so a multi-line summary cannot migrate into evidence. (#498)
+
 ### Added
 - Evidence ingest now applies an origin-scoped redaction policy (`brigade.evidence-redaction.v1`) before persistence. Sources classify to the provenance envelope origin; each origin selects detectors; the stored item keeps only redacted bytes plus a count/detector record (never the removed values). Scanner failure, timeout, or unavailability cannot produce a clean verdict and persists a placeholder instead of the original. The policy version is stamped on every decision and applies to future writes only — existing rows and provenance backfill are not rewritten. (#498)
 - Canonical memory cards mint one opaque `card-<uuid4>` ID on create, and the reviewed `memory care backfill` path can mint an ID on a legacy card, including complete care cards that only lack an ID. Valid IDs are never replaced. Care queues, recall and search logs, refresh imports, and retrieval-eval fixtures dual-read explicit IDs and legacy path/stem/topic aliases; alias collisions fail without rewriting cards. Dry-run backfill writes a deterministic mapping receipt with coverage and old-to-new IDs (relative paths only; `old_id` is the consumer-facing identity). A zero-candidate dry run writes nothing. Duplicate-ID coverage does not require `memory/NAMESPACE`. Missing IDs stay a doctor warning until coverage is 100 percent. (#867)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Evidence ingest now applies an origin-scoped redaction policy (`brigade.evidence-redaction.v1`) before persistence. Sources classify to the provenance envelope origin; each origin selects detectors; the stored item keeps only redacted bytes plus a count/detector record (never the removed values). Scanner failure, timeout, or unavailability cannot produce a clean verdict and persists a placeholder instead of the original. The policy version is stamped on every decision and applies to future writes only — existing rows and provenance backfill are not rewritten. (#498)
 - Canonical memory cards mint one opaque `card-<uuid4>` ID on create, and the reviewed `memory care backfill` path can mint an ID on a legacy card, including complete care cards that only lack an ID. Valid IDs are never replaced. Care queues, recall and search logs, refresh imports, and retrieval-eval fixtures dual-read explicit IDs and legacy path/stem/topic aliases; alias collisions fail without rewriting cards. Dry-run backfill writes a deterministic mapping receipt with coverage and old-to-new IDs (relative paths only; `old_id` is the consumer-facing identity). A zero-candidate dry run writes nothing. Duplicate-ID coverage does not require `memory/NAMESPACE`. Missing IDs stay a doctor warning until coverage is 100 percent. (#867)
 - `brigade memory vault-index`, `vault-search`, `vault-show`, and `vault-doctor`
   close the operator-vault projection round trip. Config lives in
--- a/docs/import-schema.md
+++ b/docs/import-schema.md
@@ -302,6 +302,19 @@ Default briefs, context packs, citations, and promotion omit `unknown` and `quar
 
 Public read surfaces recompute `hashes.content` (and `hashes.raw` when materialized) against the exact stored bytes. Search suppresses a mismatched snippet and sets `integrity_mismatch: true`. Direct `miseledger show` / `brigade evidence show` hide a mismatched or synthesized-legacy body unless `--forensic-content` is passed. `--forensic-content` never changes trust and only reveals a body when injection status is the explicit known-safe value `clean`; empty, unknown, and parse-lost statuses block. MCP, HTTP, bundles, briefs, and context stay metadata-only on mismatch. One idempotent downgrade event is appended per item/hash/mismatch; the row is never deleted.
 
+### Origin-scoped redaction (#498)
+
+New work-import, research-finding, and receipt-index writes run `brigade.evidence-redaction.v1` before the item text is persisted. The policy keys off envelope `origin`:
+
+| Origin | Detectors |
+| --- | --- |
+| `operator-input`, `workspace` | secrets |
+| `agent-session` | secrets, PII |
+| `external-service`, `external-web` | secrets, PII, infrastructure |
+| `unknown` | fail-closed maximum set |
+
+The optional envelope `redaction` object stores `schema`, `schema_version`, `policy_version`, `origin`, `status` (`clean` / `redacted` / `error`), `count`, and `detectors`. It never stores removed values, excerpts, or original bytes. `hashes.content` is the digest of the persisted (already redacted) text. Scanner failure, timeout, or unavailability records `status=error`, persists `[ingestion-redaction-failed]`, and cannot be treated as clean. A missing redaction record on a legacy row is not a clean verdict. New policy versions apply to future writes only; provenance backfill does not rewrite existing text.
+
 ### Legacy banner
 
 When an item carries no provenance, `synthesize_legacy_provenance()` returns a non-null envelope with `origin = unknown`, `modality = unknown`, `attribution = inferred`, `trust.label = unknown`, null `repository`/`session`/`collection_id`/`item_id`/`locator`, and a null content digest. The display string is `UNKNOWN PROVENANCE - legacy item` (`provenance.LEGACY_DISPLAY`). A missing envelope is never treated as trusted.
--- a/docs/import-schema.md
+++ b/docs/import-schema.md
@@ -313,6 +313,8 @@ New work-import, research-finding, and receipt-index writes run `brigade.evidenc
 | `external-service`, `external-web` | secrets, PII, infrastructure |
 | `unknown` | fail-closed maximum set |
 
+Work-import `--source` is normalized with `strip().lower()` and fails closed to `unknown` when unmapped. `external-web` and `external-service` are selectable source names. `work import context` is an external-content surface: weaker mapped sources upgrade to `external-web` so secrets, PII, and infrastructure are in scope.
+
 The optional envelope `redaction` object stores `schema`, `schema_version`, `policy_version`, `origin`, `status` (`clean` / `redacted` / `error`), `count`, and `detectors`. It never stores removed values, excerpts, or original bytes. `hashes.content` is the digest of the persisted (already redacted) text. Scanner failure, timeout, or unavailability records `status=error`, persists `[ingestion-redaction-failed]`, and cannot be treated as clean. A missing redaction record on a legacy row is not a clean verdict. New policy versions apply to future writes only; provenance backfill does not rewrite existing text.
 
 ### Legacy banner
--- a/src/brigade/evidence_redaction.py
+++ b/src/brigade/evidence_redaction.py
@@ -1,0 +1,254 @@
+"""Origin-scoped evidence-ingestion redaction policy (#498).
+
+Classifies capture sources, redacts sensitive values before persistence, and
+records counts only. Removed bytes never persist on the verdict, the envelope
+record, or the stored item. Scanner failure, timeout, or unavailability cannot
+produce a clean verdict. A new policy version applies to future writes only.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Callable, Mapping, Sequence
+from dataclasses import dataclass
+from typing import Any
+
+from .guard.engine import scan_text
+from .guard.policy import Policy
+from .guard.rules import DEFAULT_RULES
+from .guard.types import Action, Finding, GuardResult, ScanOptions
+
+SCHEMA = "brigade.evidence-redaction.v1"
+SCHEMA_VERSION = 1
+POLICY_VERSION = "brigade.evidence-redaction.v1"
+
+ORIGINS = frozenset({"operator-input", "workspace", "agent-session", "external-service", "external-web", "unknown"})
+STATUSES = frozenset({"clean", "redacted", "error"})
+FAILED_PLACEHOLDER = "[ingestion-redaction-failed]"
+MAX_DETECTORS = 32
+
+# Work-inbox producers. Unknown sources stay workspace, matching the existing
+# ledger default. Envelope origin ``unknown`` is a separate fail-closed class.
+SOURCE_ORIGINS: dict[str, str] = {
+    "manual": "operator-input",
+    "backup-health": "workspace",
+    "chat-memory-sweep": "agent-session",
+    "code-review": "workspace",
+    "context-pack": "workspace",
+    "handoff-ingest": "agent-session",
+    "learning-loop": "workspace",
+    "memory-care": "workspace",
+    "memory-refresh": "workspace",
+    "project-consolidation": "workspace",
+    "repo-fleet": "workspace",
+    "repo-fleet-release": "workspace",
+    "roadmap-audit": "workspace",
+    "scanner-health": "workspace",
+    "security-scan": "workspace",
+    "tool-catalog": "workspace",
+}
+
+# Origin-scoped detector categories. ``unknown`` is fail-closed maximum scope.
+ORIGIN_CATEGORIES: dict[str, frozenset[str]] = {
+    "operator-input": frozenset({"secret"}),
+    "workspace": frozenset({"secret"}),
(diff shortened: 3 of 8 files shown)


solomonneas:
**VERDICT: BLOCKED** - the external-content ingest path cannot reach the external redaction tiers and fails open to the weakest one, so planted PII persists in cleartext from `work import context`.

Reviewed at `af872d62` in a scratch clone (`gh pr checkout 997`), brigade 0.27.0 installed into a venv. All output below is executed, not read.

The core non-storage claim holds. The origin selection around it does not.

---

## What passed

**Probe 1 - non-storage.** Planted a fake bearer token, an email, an absolute home path, and a private IP into `work import add --source chat-memory-sweep`, then grepped every file the run produced.

```
$ grep -rl --binary-files=text 'zqXK9tPLANTEDBEARER7731aabbccddeeff0011' .   -> 0 files
$ grep -rl --binary-files=text 'planted.canary9931@probe-corp-canary.com' .  -> 0 files
```

Stored item and record:
```
text: Auth header Bearer [redacted-secret] contact <PRIVATE_EMAIL> path /home/... ip 192.168.4.70
redaction: {"count": 2, "detectors": ["bearer-token", "email"], "origin": "agent-session",
            "policy_version": "brigade.evidence-redaction.v1", "status": "redacted", ...}
```
Zero hits for the detected classes across stores, envelope, receipts, and logs. Only counts and detector ids persist. `Finding.match` carries raw matched bytes and is correctly never propagated into the record.

**Probe 2a - `resolve_origin` variant table.** Fails closed on every variant, to maximum scope:

| Variant | Input | Resolved | Count |
| --- | --- | --- | --- |
| exact | `agent-session` | agent-session | 2 |
| Case | `Agent-Session` | **unknown** | 3 |
| UPPER | `AGENT-SESSION` | **unknown** | 3 |
| trailing space | `agent-session ` | **unknown** | 3 |
| leading space | ` agent-session` | **unknown** | 3 |
| empty | `""` | **unknown** | 3 |
| garbage | `zzz-not-a-real-origin` | **unknown** | 3 |

This is the right shape. The bug is that the ingest path does not use it.

**Probe 3 - counts.** Accurate and durable. N planted secrets yield count N at N = 2, 4, 6, 10. Re-stamping an already-stamped finding returns early on the envelope match and preserves `count: 1` with no double-count.

**Probe 4 - fail closed.** Scanner raising mid-ingest, and scanner returning `None` / `str` / `dict`, all produce `status=error`, `count=0`, `persisted_text='[ingestion-redaction-failed]'`, trust downgraded to `quarantined`, and `ingest_verdict_is_clean() == False`. No planted bytes in the persisted text or the record.

**Probe 5 - tests and mutation.** `336 passed in 7.32s` across the six evidence/ingest modules. Mutation: reverted `"text": persist_text` back to `"text": text` in `_make_import`:
```
FAILED tests/test_evidence_redaction.py::test_make_import_redacts_before_persist_and_hashes_redacted_bytes
FAILED tests/test_evidence_redaction.py::test_make_import_scanner_failure_quarantines_and_drops_original
2 failed, 165 passed
```
The new tests are load-bearing. Restored after.

Scope matches #498 and the amendment, no drive-by refactors, CHANGELOG entry describes effects, no secrets in the diff.

---

## Findings

### 1. BLOCKING - `work import context` ingests external content at the weakest tier

`work import context` is the path whose own help text says "Inbox raw external context as untrusted data". It wraps the body in an UNTRUSTED fence. It cannot select an external redaction tier.

```
$ brigade work import context --kind transcript --source external-web \
    "External page says Bearer zqXK9t... mail planted.canary9931@probe-corp-canary.com ip 192.168.4.70"

text: ... External page says Bearer [redacted-secret] mail planted.canary9931@probe-corp-canary.com ip 192.168.4.70
envelope origin: workspace
redaction: {"count": 1, "detectors": ["bearer-token"], "origin": "workspace", "status": "redacted"}
```

The planted email and the private IP persist in cleartext, from external content, and the record says the item was redacted. Root cause:

```
distinct origins in SOURCE_ORIGINS: ['agent-session', 'operator-input', 'workspace']
declared ORIGIN_CATEGORIES tiers:   ['agent-session', 'external-service', 'external-web',
                                     'operator-input', 'unknown', 'workspace']
tiers NEVER selectable via work-import source: ['external-service', 'external-web']
```

No source string maps to `external-service` or `external-web`, so the tier that `docs/import-schema.md` advertises as covering secrets + PII + infrastructure is unreachable from the work-import surface regardless of what the operator types. The research path does reach it via legacy trust (`browser`/`web` -> `external-web`, `cli` -> `external-service`), so the tiers are not globally dead, just unreachable from the surface that handles external content.

### 2. BLOCKING - `classify_source_origin` fails open on case and unknown variants

`resolve_origin` fails closed to `unknown`. Its sibling on the user-controlled input fails open to `workspace`, the weakest tier. `--source` on `work import add` is free-form.

| Variant | Input | Origin |
| --- | --- | --- |
| exact | `chat-memory-sweep` | agent-session |
| Case | `Chat-Memory-Sweep` | **workspace** |
| UPPER | `CHAT-MEMORY-SWEEP` | **workspace** |
| trailing space | `chat-memory-sweep ` | agent-session (stripped) |
| empty | `""` | **workspace** |
| garbage | `zzz-garbage-source` | **workspace** |
| Case | `Handoff-Ingest` | **workspace** |

End to end, same content, one letter of case difference:
```
$ brigade work import add --kind finding --source Chat-Memory-Sweep \
    "case variant contact planted.canary9931@probe-corp-canary.com"
text: case variant contact planted.canary9931@probe-corp-canary.com
```
The email survives. Under the exact-case source it is redacted to `<PRIVATE_EMAIL>`. This is the recurring bypass class this campaign keeps hitting: a string comparison in an enforcement surface failing open on a case variant. The comment at `evidence_redaction.py:29` acknowledges the choice ("Unknown sources stay workspace, matching the existing ledger default"), but matching the modality default is not a reason for the redaction scope to inherit it. An unmapped or misspelled producer should route to `unknown` and get maximum scope, the same as `resolve_origin` already does. Silent scope downgrade with no signal also means any future producer added without a `SOURCE_ORIGINS` entry loses PII redaction and nothing fails.

### 3. HIGH - identity check lets an unredacted value persist while the record claims a redaction

`evidence_redaction.py:202`:
```python
if persisted is text and count:
    return _error_verdict(resolved)
```
This is the only assertion that redaction actually removed bytes, and it compares object identity, not value. A scanner returning an equal but distinct string with findings passes it:

```
same object (str(t))   same_obj=True  equal=True  -> status=error    count=0 LEAKED=False
EQUAL but NEW object   same_obj=False equal=True  -> status=redacted count=1 LEAKED=True
    persisted: 'Bearer zqXK9tPLANTEDBEARER7731aabbccddeeff0011'
    record: {... "status": "redacted", "count": 1, "detectors": ["bearer-token"]}
genuinely redacted     same_obj=False equal=False -> status=redacted count=1 LEAKED=False
```
The full secret persists and the audit trail reports a successful redaction. Not reachable through the default `scan_text` with today's `DEFAULT_RULES`, since no rule's replacement equals its match, so this is latent rather than live. Given that the whole PR rests on this guard, `==` is the correct comparison.

### 4. MEDIUM - `_split_redacted_finding_text` corrupts field boundaries on multi-line values

`finding_text` joins with `\n`; `_split_redacted_finding_text` splits with `maxsplit=2`. A finding whose summary spans lines gets reassembled wrong, but only when redaction fires:

```
input  summary: 'summary line one contact <mail>\nsummary line TWO stays in summary'
input  evidence: 'evidence body here'

output summary : 'summary line one contact <PRIVATE_EMAIL>'
output evidence: 'summary line TWO stays in summary\nevidence body here'
```
The second summary line silently migrates into `evidence`. Clean findings keep their structure, so tests with single-line fields will not catch it. Carrying per-field redaction, or redacting title/summary/evidence separately, avoids the lossy round trip.

### 5. LOW - coverage gaps worth confirming as intentional

- No detector in `DEFAULT_RULES` matches an absolute home path. `/home/canaryuser9931/secretdir/notes.txt` persisted verbatim under every origin including `unknown`. The repo's own PR checklist calls out "unredacted absolute paths", so the ingest boundary not covering them is a gap.
- `receipts_cmd._receipt_redact` hardcodes `origin="agent-session"`, which excludes `infrastructure`. Receipts routinely carry private IPs, ports, and localhost endpoints; none are redacted. Confirmed: `192.168.4.70` survives at `agent-session`.
- `ingest_verdict_is_clean` is public API with no production caller. Only tests reference it. Either wire it into the read surfaces that need to distinguish scanned-clean from never-scanned, or note why it ships unused.
- `inferred=True` skips redaction and writes no redaction record, so backfilled findings keep raw text. This reads as consistent with the amendment (existing rows are not rewritten), but confirm that no new write ever reaches `stamp_finding_payload` with `inferred=True`.

---

## What would unblock

Findings 1 and 2 are one fix: make the unmapped case in `classify_source_origin` return `unknown` instead of `workspace`, normalize the lookup (casefold plus strip), and give `work import context` a way to declare an external origin so the documented external tier is reachable. Finding 3 is a one-character change with a test. Finding 4 needs a small restructure of the research write path.

The non-storage machinery underneath is solid, and the fail-closed behavior on scanner failure is correct and well tested. The gap is which origin the ingest surfaces hand it.

solomonneas:
**STILL BLOCKED - the regression test for the `persisted is text` fail-open is vacuous and passes against the unfixed code.**

To be clear up front: all four findings are correctly fixed in the production code, and I verified each one by direct probe at `ae2483fa`. The only remaining item is that one of the six new regression tests cannot fail, so it does not hold the line it was added to hold. That is a one-line change.

Re-verified in the same scratch clone, fetched to `ae2483fa`, brigade 0.27.0 editable install.

---

## Fixes confirmed

**Finding 1 - `work import context` external tier. FIXED.**
```
$ brigade work import context --kind transcript --source external-web \
    "External page says Bearer zqXK9t... mail planted.canary9931@probe-corp-canary.com ip 192.168.4.70"

zqXK9tPLANTEDBEARER7731aabbccddeeff0011       hits=0
planted.canary9931@probe-corp-canary.com      hits=0
192.168.4.70                                  hits=0

source: external-web | envelope origin: external-web
redaction: {"count": 3, "detectors": ["bearer-token","email","private-ipv4"], "origin": "external-web", "status": "redacted"}
body: External page says Bearer [redacted-secret] mail <PRIVATE_EMAIL> ip [redacted-ip]
```
Zero cleartext hits across every artifact. All three detectors fire. `unreachable tiers: []` now, against `['external-service', 'external-web']` before.

**Finding 2 - case-variant `--source`. FIXED.**

| Variant | Input | Origin | Count |
| --- | --- | --- | --- |
| exact | `chat-memory-sweep` | agent-session | 2 |
| Case | `Chat-Memory-Sweep` | agent-session | 2 |
| UPPER | `CHAT-MEMORY-SWEEP` | agent-session | 2 |
| trailing space | `chat-memory-sweep ` | agent-session | 2 |
| leading space | ` chat-memory-sweep` | agent-session | 2 |
| empty | `""` | **unknown** | 3 |
| garbage | `zzz-garbage-source` | **unknown** | 3 |
| non-str | `None` | **unknown** | 3 |

Known producers normalize to their true origin, unmapped and empty fail closed to `unknown` at maximum scope. End to end, the case variant that previously leaked:
```
--source Chat-Memory-Sweep  -> text: case variant contact <PRIVATE_EMAIL>
--source zzz-garbage-source -> text: garbage source contact <PRIVATE_EMAIL> ip [redacted-ip]
grep across all artifacts: 0 hits for all three planted values
```

**Finding 3 - identity check. FIXED in code.**
```
same object (str(t))   equal=True  -> status=error    LEAKED=False
EQUAL but NEW object   equal=True  -> status=error    LEAKED=False   (was: status=redacted, LEAKED=True)
genuinely redacted     equal=False -> status=redacted LEAKED=False
```

**Finding 4 - field corruption. FIXED.**
```
input  summary : 'summary line one contact <mail>\nsummary line TWO stays in summary'
output summary : 'summary line one contact <PRIVATE_EMAIL>\nsummary line TWO stays in summary'
output evidence: 'evidence body here'   (unchanged)
```
Per-field redaction also aggregates correctly: one planted email in each of title, summary, and evidence gives `count: 3`. Clean findings keep multi-line structure untouched. A scanner failure on any single field fails closed for the whole finding: all three fields dropped to `[ingestion-redaction-failed]` / `""`, `status: error`, trust `quarantined`, zero planted bytes anywhere in the payload.

## Regressions on previously passing probes

All clean. `resolve_origin` still fails closed on all seven variants. Counts accurate at N = 2, 4, 6, 10. Scanner raise and malformed returns (`None`, `str`, `dict`) all give `status=error` with the placeholder and no leak. Replay stays idempotent with no double-count.

`582 passed in 31.91s` across nine evidence, ingest, research, receipts, and context modules.

---

## The blocking item

`test_equal_but_distinct_unredacted_text_fails_closed` passes against the unfixed code. Mutation matrix, each reverting one fix hunk and running the suite:

| Mutation | Reverts | Result |
| --- | --- | --- |
| M1 | `classify_source_origin` back to `workspace` default | **4 failed** |
| M2 | `origin_for_external_ingest` stops upgrading | **3 failed** |
| M3 | `persisted == text` back to `persisted is text` | **43 passed, 0 failed** |
| M4 | per-field redaction back to joined split | **1 failed** |

M3 restores the exact fail-open I reported and the suite stays green. The cause is the scanner's return value at `tests/test_evidence_redaction.py:385`:

```python
return GuardResult(text=text, redacted_text=str(text), findings=[finding])
```

In CPython, `str(s)` on something that is already exactly `str` returns the same object, so the test never builds the equal-but-distinct string its name promises:

```
str(secret) is secret            -> True    <- what the test builds
''.join(list(secret)) is secret  -> False   <- genuinely distinct
```

The identity case was already handled by the old `is` comparison, so this assertion holds under both implementations. I walked into the same trap writing my original probe, which is why I am flagging it rather than treating it as nitpicking: the construction looks correct and is not.

**Fix:** make the scanner return a genuinely distinct object, for example `redacted_text="".join(list(text))` or `(text + "x")[:-1]`. With that change M3 should fail. Worth adding a comment on the line so nobody "simplifies" it back to `str(text)`.

The other five new tests are load-bearing and mutation-proven.

## Out of scope, for follow-up filing

Unchanged and correctly left alone per the agreed scope: no home-path detector in `DEFAULT_RULES`, `receipts_cmd._receipt_redact` hardcoded to `agent-session` so infrastructure in receipts is never redacted, and `ingest_verdict_is_clean` still has no production caller. These want a follow-up issue against #498, not a change here.

One thing to sanity check before merge, not a blocker: unmapped sources now produce envelope `origin: unknown` where they previously produced `workspace`. Since briefs, context packs, citations, and promotion omit `unknown` by default, any real producer missing from `SOURCE_ORIGINS` will now drop out of those surfaces. That is the correct fail-closed direction, and the full suite is green, so I mention it only so the behavior change is a decision rather than a surprise.

Flip to READY once M3 fails.

solomonneas:
**RE-VERIFIED: READY**

`9853adb5` makes the identity-check regression test load-bearing. That was the only outstanding item, and it now holds.

## The decisive check

M3, reverting the fix hunk back to the fail-open:

```
$ sed -i 's/if persisted == text and count:/if persisted is text and count:/' src/brigade/evidence_redaction.py
225:    if persisted is text and count:

$ pytest -q tests/test_evidence_redaction.py
E       AssertionError: assert 'redacted' == 'error'
E         - error
E         + redacted
FAILED tests/test_evidence_redaction.py::test_equal_but_distinct_unredacted_text_fails_closed
1 failed, 42 passed in 0.18s
```

Previously this same mutation left the suite at `43 passed, 0 failed`. The fixture change to `"".join(list(text))` is what closed it, and the comment on the line should stop anyone folding it back to `str(text)`.

Restored and green: `582 passed in 30.67s` across the nine evidence, ingest, research, receipts, and context modules.

## Full mutation matrix, all four fixes now covered

| Mutation | Reverts | Result |
| --- | --- | --- |
| M1 | `classify_source_origin` back to `workspace` default | 4 failed |
| M2 | `origin_for_external_ingest` stops upgrading | 3 failed |
| M3 | `persisted == text` back to `persisted is text` | **1 failed** (was 0) |
| M4 | per-field redaction back to joined split | 1 failed |

## Behavior re-confirmed at this exact commit

```
zqXK9tPLANTEDBEARER7731aabbccddeeff0011       hits=0
planted.canary9931@probe-corp-canary.com      hits=0
192.168.4.70                                  hits=0

source=external-web         origin=external-web    count=3 detectors=['bearer-token','email','private-ipv4']
source=Chat-Memory-Sweep    origin=agent-session   count=1 detectors=['email']
source=zzz-garbage-source   origin=unknown         count=2 detectors=['email','private-ipv4']
```

External content reaches the external tier, case variants normalize to the true origin, garbage sources fail closed to `unknown` at maximum scope, and zero planted bytes survive anywhere in the artifacts. Counts and detector ids are the only trace, which is what #498 asked for.

## Follow-up filed

The three coverage gaps I flagged and we agreed to keep out of scope are now #998: no absolute-home-path detector in `DEFAULT_RULES`, `receipts_cmd._receipt_redact` hardcoded to `agent-session` so infrastructure never redacts in receipts, and `ingest_verdict_is_clean` shipped without a production caller.

One note for the merge commit, not a blocker: unmapped work-import sources now stamp envelope `origin: unknown` instead of `workspace`, and briefs, context packs, citations, and promotion omit `unknown` by default. Correct fail-closed direction, full suite green, worth being a known consequence rather than a surprise.

Nice work on the turnaround.

Issue #881, opened 2026-08-12
fix(work): bind scanner receipt and proof bytes to verifier authority


solomonneas:
## Problem

PR #871 binds scanner-run and proof directories to verifier-owned external identity records, but legacy migration still trusts file contents that a scanner process can reproduce after its run exits.

An arbitrary configured scanner can keep a child process alive, overwrite its existing `receipt.json` in place with a successful built-in receipt payload, and create the deterministic proof sidecar inside the already bound proof directory. The directory identities remain valid. `_legacy_import_source_content_identity` then accepts the forged row as locally authorized.

## Reproduction

The regression creates verifier-bound scanner-run and proof directories, writes an initial failed receipt, replaces its bytes with a completed built-in receipt, and writes the row-derived sidecar. The final assertion expects no legacy identity, but the current code returns:

```text
('handoff-ingest', 'task', '9c4208e5aac3736f')
```

Evidence:

- Brigade verification `20260812-124911-work-verify-aed403`
- Brigade evidence `72700c30a4effb1c840cfeb9`
- Result: 1 failed security assertion, 1 passed containment control
- Reproduced against PR #871 head `93723f29035c88c58972f91f78e2e3d43146d029`

## Expected behavior

Successful receipt bytes and persisted proof bytes need verifier-owned authority that cannot be rebuilt from row fields. Readers should reject in-place receipt changes, attacker-created sidecars, missing bindings, and mismatched file identity or content.

Publication must record the binding after the file is durable. A binding-write failure must restore the exact prior receipt, proof, inbox, and prior binding state or fail closed.

Follow-up to #871.

Issue #1028, opened 2026-08-19
security: Daybreak Blue defensive pass 2 (2026-08-19) - deep read of ledger.py, scanners.py, run_transport.py, app.go


solomonneas:
Second pass over the four files the first pass (#1006) sampled but did not read exhaustively: `ledger.py`, `scanners.py`, `run_transport.py`, and `app.go`. This is a GPT-5.6 Daybreak Blue (xhigh) defensive review under the same same-UID compromised scanner/seat threat model as #1006, run read-only against commit `1bc5846024e112407701500f1b6ce6cfc4400912`. It reports 21 findings (1 CRITICAL, 9 HIGH, 9 MEDIUM including 2 SUSPECTED, 2 LOW) plus an invariant inventory and coverage note. This is the pass-2 analog of #1006.

Split into discrete issues below; mapping comment follows.

---

Commit SHA read: 1bc5846024e112407701500f1b6ce6cfc4400912
Checkout HEAD from git rev-parse HEAD: ad6956d93e0540394cdef7354ccb395121f8c2ca. That checkout was behind the requested baseline and already dirty, so every cited line was read from the immutable origin/main commit object above using git show. No files were modified.

Findings

CONFIRMED - CRITICAL
engines/evidence-ledger/internal/app/app.go:324-365
engines/evidence-ledger/internal/ingest/provenance.go:249-346
Attack: a compromised scanner imports hostile evidence, reads its item ID and digest, then invokes trust review --to-label verified --mark-injection-clean. The command accepts caller-provided operator attribution but performs no operator authentication, capability check, TTY confirmation, or principal binding. The digest checks pass because the attacker knows its own content. Search/MCP then releases the body as verified and clean.
Broken invariant: only an authenticated operator can elevate trust or clear an injection finding.
Fix: place review behind a privileged broker unavailable to scanner processes. Require a short-lived capability bound to item ID, current digest, requested transition, nonce, and expiry. Treat operator-command only as audit metadata.

CONFIRMED - HIGH
engines/evidence-ledger/internal/app/app.go:2462-2496
engines/evidence-ledger/internal/app/integrity.go:139-186
engines/evidence-ledger/internal/ingest/importer.go:798-810
Attack: a same-UID child changes artifacts.text on an eligible URL artifact while leaving its URL and content_hash unchanged. Evidence-bundle queries omit metadata_json, so the artifact provenance envelope is never checked. The fallback accepts the unchanged URL hash even though the emitted text changed.
[...]

PR #1019, opened 2026-08-19
fix(evidence-ledger): hide ineligible bodies on every read surface

solomonneas:
## What and why

Fixes #1007
Fixes #1009

Quarantined/pending evidence bodies were hidden by `show` but still leaked through search snippets, MCP `search_evidence`, evidence bundles, and Markdown export. Malformed envelopes, whitespace/case injection statuses, object-valued statuses, and uppercase matching digests fail-opened because raw string compares and digest `ToLower` authorized bodies.

Content eligibility is now centralized: every body-emitting read surface emits a body only when provenance parses and `trust.injection.status` is the validated typed value `clean` on a non-`unknown` / non-`quarantined` label. Parse errors are blocking. Stored digest representations are compared exactly and are not lowercased to authorize a body. Direct `show --include-untrusted-body` / `--forensic-content` remain the only explicit reveal paths.

Mutation tests in `engines/evidence-ledger/internal/app/content_eligibility_test.go` reproduce both attacks (quarantined/pending leak on non-show surfaces; fail-open envelope/digest authorization) and fail on the unfixed code.

## Type of change

- [x] Bug fix
- [ ] New harness adapter / doctor check
- [ ] Docs
- [ ] Refactor with no command-surface change
- [ ] Surface change (new harness, depth, include, or breaking template/ingester change) — opened an issue first per CONTRIBUTING.md

## Package tests

Touched package: `engines/evidence-ledger`

```text
$ go vet ./... && go test ./... -count=1 -timeout 600s
?   	github.com/escoffier-labs/miseledger/cmd/sessionfind	[no test files]
ok  	github.com/escoffier-labs/miseledger/cmd/miseledger	0.003s
ok  	github.com/escoffier-labs/miseledger/internal/adapter	0.050s
ok  	github.com/escoffier-labs/miseledger/internal/app	3.373s
ok  	github.com/escoffier-labs/miseledger/internal/archive	0.186s
ok  	github.com/escoffier-labs/miseledger/internal/ingest	5.886s
ok  	github.com/escoffier-labs/miseledger/internal/provenance	0.006s
ok  	github.com/escoffier-labs/miseledger/internal/security	0.005s
ok  	github.com/escoffier-labs/miseledger/internal/sources	0.003s
ok  	github.com/escoffier-labs/miseledger/internal/sources/claude	0.002s
ok  	github.com/escoffier-labs/miseledger/internal/sources/codex	0.084s
[...]

Issue #1006, opened 2026-08-19
security: Daybreak Blue defensive pass 2026-08-19 - 9 confirmed findings (same-UID forgery, fail-open eligibility, containment, exhaustion)


solomonneas:
Defensive-security audit by GPT-5.6 Daybreak Blue (xhigh), run manually in the ChatGPT GUI against commit `f8baffa7` (near-current main; the last campaign merge). Read-only, no files modified. 9 confirmed findings + 1 suspected, most on the same enforcement surfaces the 2026-08-17/18 campaign hardened. The recurring theme: **a same-UID compromised scanner or seat is the threat model, and env/HOME/XDG scrubbing is not a filesystem boundary for a same-UID process** - which is exactly what #957 (authority-store HMAC, spec posted) is meant to address.

This is a tracking issue in the shape of the #904 audit: findings are verbatim below. Split into discrete issues per the assessment, then dispatch fix lanes with the executed-probe review bar (each fix's mutation test confirms or refutes the finding). Mapping to existing work noted per finding.

---

### 1. HIGH, CONFIRMED - scanner rewrites unsigned 0600 authority store (extends #957)
`src/brigade/work_cmd/ledger.py:1514-1526, 1586-1605, 1767-1776, 1870-1887, 2306-2434`; `src/brigade/work_cmd/scanners.py:126-175, 403-427, 1607-1618`
Attack: a malicious scanner under the operator's UID locates the deterministic authority-store path, leaves a background child running, replaces a receipt or proof after publication, then rewrites the unsigned 0600 authority JSON with the replacement inode and digest. It can copy the public built-in scanner config and forge self-import proofs; the verifier accepts the attacker-controlled file as verifier-owned. Rewriting HOME/XDG does not create a boundary for a same-UID process.
Fix: run scanners under a separate UID or isolated user/mount namespace with no access to Brigade state, terminate the whole process group, and bind results through parent-held descriptors or signatures generated by a key unavailable to the scanner. **This is the #957 design; Daybreak confirms it live and adds the copy-public-config self-import-proof forgery path.**

### 2. HIGH, CONFIRMED - quarantined/pending content leaks through non-show surfaces
`engines/evidence-ledger/internal/app/app.go:2096-2133, 2428-2494, 2863-2951`; `integrity.go:305-325`; `mcp.go:225-243`
[...]

PR #1048, opened 2026-08-19
feat(security): parent-held capability broker and HMAC authority store

### LINKED ISSUE #957: scanner authority store: residual same-uid write class after env scrub
Follow-up from the PR #949 reviews (refs #881, comments 5318178972 and 5318352431).

The env scrub in #949 closes the inherited-environment path, but the residual class remains and still reproduces (reviewer probe G5: a scanner subprocess running as the same uid can write the 0600 binding store directly by path, then a forged identity verifies). The store's trust boundary is process-identity, not file mode, while scanner children share the uid.

Documented honestly in #949's CHANGELOG ('Refs #881', not a class close). Candidate directions: authority store under a different uid/service, HMAC with a key outside the scanner-reachable tree, or verifying bindings against an append-only journal outside the workspace. Needs a design decision, not a patch.

### LINKED ISSUE #1029: security(CRITICAL): trust review --mark-injection-clean performs no operator authentication (app.go:324)
Part of #1028 (Daybreak Blue pass 2, 2026-08-19).

This finding lands on the eligibility path established by the just-merged PR #1019 (which made `--mark-injection-clean` the operator route to clean an injection finding), so an unauthenticated caller reaching that route is high priority. Relates to #1019 and extends #957.

```
CONFIRMED - CRITICAL
engines/evidence-ledger/internal/app/app.go:324-365
engines/evidence-ledger/internal/ingest/provenance.go:249-346
Attack: a compromised scanner imports hostile evidence, reads its item ID and digest, then invokes trust review --to-label verified --mark-injection-clean. The command accepts caller-provided operator attribution but performs no operator authentication, capability check, TTY confirmation, or principal binding. The digest checks pass because the attacker knows its own content. Search/MCP then releases the body as verified and clean.
Broken invariant: only an authenticated operator can elevate trust or clear an injection finding.
Fix: place review behind a privileged broker unavailable to scanner processes. Require a short-lived capability bound to item ID, current digest, requested transition, nonce, and expiry. Treat operator-command only as audit metadata.
```

## Fix bar

Fix lane must ship a mutation test that reproduces the attack and confirms the fix refutes it (SUSPECTED findings: first work item is a runtime probe to confirm or refute, then a fix only if confirmed). Independent executed-probe review before merge.

solomonneas:
## What and why

Implement approved Option A from https://github.com/escoffier-labs/brigade/issues/957#issuecomment-5347614521: a portable parent-held broker with two-tier HMAC as the default, and `--isolated-scanners` as opt-in POSIX isolation. Design is not reopened.

`brigade evidence trust review` mints a fresh in-memory capability and hands it to `miseledger` on stdin. A same-UID scanner that invokes the engine directly cannot elevate trust or clear injection without that capability. Missing capability is refused for every stdin kind: piped empty input, `/dev/null`, and a real PTY. stdin is not an authorization signal.

The directory-authority store is wrapped in a parent-held HMAC envelope. That raises the cost of same-UID forgery. It does not close the #957 class. The residual-honesty tests encode that boundary so "class closed" cannot silently drift.

Fixes #1029
Part of #957
Refs #881

## Type of change

- [x] Bug fix
- [x] Docs

New flag: `--isolated-scanners` on `work scanners run` and `work sweep`. Doctor gains `security: scanner isolation` (`OK` when userns is available, `WARN`/`MANUAL` when not, including Windows). No new runtime dependency.

## Covered in this PR

1. **#1029 (closed if the executed char-device/PTY probes refuse).** Parent mints a 32-byte secret in RSS, MACs `{item_id, from_digest, transition, nonce, expiry}`, and writes `brigade.authority.capability-handoff.v1` to the engine stdin. Secret never lands on disk, in argv, or in the child env. Engine verifies the MAC, binds the transition, and records the nonce in `used_capabilities` (schema v5). There is no TTY / char-device / env-var exception. `--operator-command` is audit only.
2. **Shared HMAC broker** (`src/brigade/authority_broker.py` + `engines/evidence-ledger/internal/ingest/capability.go`) with domain-separated prefixes for store, sequence, and capability.
3. **#957 crypto tier (partial).** Store writes go through a 0600 parent key at `<config_root>/brigade/authority/store-hmac.key`. Reads fail closed on MAC mismatch, filename/target splice, and unsigned downgrade after a sequence exists. Public `SCANNER_DEFAULTS` equality no longer grants import privilege.
4. **Opt-in isolation.** `--isolated-scanners` uses `unshare --user --map-root-user --mount` and bind-mounts empty covers over the operator authority and store dirs. POSIX only. Windows and kernels without unprivileged userns stay on the crypto tier and do not fail `windows-native-acceptance`.

## Follow-up after executed-probe review

The first revision had a TTY fallback that treated any `os.ModeCharDevice` as an interactive operator, so `< /dev/null` and a real PTY bypassed the capability gate. That fallback and `AllowMissingCapability` are removed. The mutation test now drives `RunWithStdin` with piped empty input, `/dev/null`, and a real PTY. Independent re-verify of those probes is still required; this is not self-resolved.

## Not covered (follow-up)

- Store-migration CLI: `brigade work authority migrate` / `rotate` / `repair` / `key init` is not shipped. Crash-window / key-loss errors now say to restore `store-hmac.key` and `sequence.json` or re-bind the workspace, and no longer name a missing `brigade work authority repair` command.
- Concurrent lost-update / sequence-race probe beyond the current sequence MAC.
- Making `--isolated-scanners` the default.
- Claiming the #957 same-UID class is closed. Crypto-only leaves a key-holder residual. Do not write `Fixes #957`.

## Mutation tests and residual honesty

- Scanner-forged trust-clean capability is refused on the CLI path for piped empty stdin, `/dev/null` (char device), and a real PTY (`TestMutationScannerTrustReviewWithoutCapabilityIsRefused`). Revert the fail-closed CLI check and those cases go red.
- Residual honesty (mandatory, unchanged): under the default crypto tier, a same-UID process that **reads the persisted store key** and forges a valid envelope is **ACCEPTED**. The exact same forgery is **REFUSED only under `--isolated-scanners`**. See `tests/test_authority_store_hmac.py`.
- Unsigned rewrite, attacker-key rewrite, cross-target splice, and copy-public-config self-import-proof are refused.

This PR must not be treated as self-approved. It needs an independent executed-probe review (char-device and PTY) before merge.

## Parent to engine hand-off

Documented in `src/brigade/authority_broker.py` and `engines/evidence-ledger/docs/BRIGADE_INTEGRATION.md`:

1. Operator CLI (`trust_gate.notify_miseledger_trust`) generates `secrets.token_bytes(32)`.
2. It mints the capability and writes one JSON object to the child stdin, then closes the write end.
3. The secret is absent from `env=`, argv, disk, and inherited fds other than that stdin pipe.
4. Scanner children use a scrubbed env, `stdin=DEVNULL`, and never receive `BRIGADE_AUTHORITY_KEY_FILE` (not on `_SCANNER_CHILD_ENV_ALLOWLIST`).

### Go engine

```bash
cd engines/evidence-ledger && go test ./... -count=1
```

Result: pass (`internal/app`, `internal/ingest`, and remaining packages ok). `cmd/sessionfind` has no test files.

Targeted: `go test ./internal/app -count=1 -run TestMutationScannerTrustReviewWithoutCapabilityIsRefused` pass (piped empty, `/dev/null`, real PTY).

### Python, touched packages

```bash
pytest -q tests/test_trust_capability.py tests/test_authority_broker.py tests/test_authority_store_hmac.py tests/test_work_import_provenance_stamp.py tests/test_work_cmd_scanners.py::test_scanner_child_sandbox_seeds_the_parent_directory_bindings
```

Result: **166 passed**. Residual-honesty tests were not modified.

## Checklist

- [x] Targeted pytest for touched packages passes locally (not the full coverage suite)
- [x] Conventional commit; in-house co-author trailer per CONTRIBUTING.md

<div><a href="https://cursor.com/agents/bc-d37bf912-b6b2-4cf6-af36-6245b60ea1d2?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-d37bf912-b6b2-4cf6-af36-6245b60ea1d2&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


solomonneas:
## Independent executed-probe review: VERDICT BLOCKED

Reviewed against the authoritative design (issue #957 "same-UID authority + trust-transition broker", Option A) and issue #1029. Probes were executed, not just read. One blocking finding on the #1029 critical path.

### Probe results

- P1 (#1029 capability gate): BLOCKED. A same-UID caller elevates a quarantined item to `verified` + injection `clean` with no operator capability, by presenting a character-device stdin. Reproduced two ways.
- P2 (#957 crypto, no store key): PASS. Unsigned rewrite after a sequence exists, attacker-key rewrite, and cross-target splice are all refused. `tests/test_authority_store_hmac.py` green via `brigade work verify run`.
- P3 (residual honesty): CORRECTLY ENCODED. The suite asserts the crypto tier ACCEPTS a key-holder forgery and `--isolated-scanners` REFUSES the exact same forgery. Both cases ran live here (junit skipped=0), not skipped. CHANGELOG states the residual honestly and does not claim the #957 class is closed.
- P4 (isolation tier): PASS. `--isolated-scanners` genuinely fenced the store key on Linux (the refuse test ran and passed). Windows and non-userns kernels degrade to crypto tier via `probe_isolation`, not a hard failure.
- P5 (migration): PARTIAL. First-contact accept-and-upgrade for existing unsigned stores is implemented, so no lockout on a normal upgrade. But the migrate/rotate/repair/key-init CLI is not shipped, and the crash-window and key-loss errors in `ledger.py` instruct operators to run `brigade work authority repair --target .`, which does not exist. No shipped recovery path for those states.
- P6 (suite + honesty): Go engine and touched Python tests are green. The #957 CHANGELOG entry and `Part of #957` are honest. The #1029 keyword and claim are not (see below).

### BLOCKING finding: TTY fallback defeats the #1029 capability gate

`engines/evidence-ledger/internal/app/app.go:99-105` — `stdinIsInteractive` treats any `os.ModeCharDevice` as an interactive TTY. `/dev/null` is a character device, so it passes. This is not an isatty check.

`engines/evidence-ledger/internal/app/app.go:422-428` — when the capability is absent, the code fails closed only if `BRIGADE_REQUIRE_TRUST_CAPABILITY=1` or stdin is non-interactive; otherwise it prints a warning and sets `AllowMissingCapability=true`, which `ReviewTrust` (provenance.go:261) then honors. Both gates are attacker-controlled: a same-UID scanner sets its own child env (so the require-env is not present) and picks its own stdin.

Executed attack (fresh DB, imported hostile item, item `quarantined`):

```
# Attack A, piped/empty stdin, no env var -> correctly REFUSED
printf '' | miseledger trust review --item <id> --content-hash <digest> \
  --to-label verified --mark-injection-clean --json
# stderr: trust review: operator capability is required; use `brigade evidence trust review`

# Attack C, /dev/null stdin, no env var -> BYPASS
miseledger trust review --item <id> --content-hash <digest> \
  --to-label verified --mark-injection-clean --json < /dev/null
# stderr: WARNING missing operator capability; honoring interactive TTY caller ...
# stdout: "injection_status": "clean"  ; item trust_label -> verified ; exit 0

# Attack B, real PTY as stdin (pty.openpty), no env var -> BYPASS
# child exit 0 ; item trust_label -> verified
```

This is exactly the #1029 attack: a same-UID scanner that knows its own item id and digest flips label to `verified` and injection to `clean` with no minted capability. Since #1019 made `--mark-injection-clean` the operator route to release a quarantined body, search/MCP then serves the hostile body as verified and clean.

Why the green suite missed it: `TestReviewTrustRefusesMissingCapability` calls `ReviewTrust` directly with `AllowMissingCapability` defaulted false, and `run()` in `app_test.go` uses `bytes.NewReader(nil)` (non-interactive). No test exercises char-device or PTY stdin, so the mutation test passes while the shipped CLI is bypassable. Reverting the check at `provenance.go:261` does turn that test red, but the test does not cover the surface the attacker uses.

Design intent vs implementation: the approved design accepts a one-release grace but states plainly that "only direct miseledger trust review invocations, which is the attack, lose the ability to transition." The implementation lets the attack keep that ability whenever it presents a character-device stdin. The design's own threat framing is that a same-UID child controls its environment and its file descriptors, so a TTY (or `/dev/null`) is not authentication.

Consequence for the claims: `Fixes #1029`, "#1029 (closed)", and the CHANGELOG line "a same-UID scanner cannot self-elevate" are refuted by the reproduction above. Those must not merge as written.

Suggested direction (author's call): fail closed on a missing capability by default regardless of stdin; if a one-release interactive-human grace is kept, gate it on a real terminal check (isatty on the fd), and make it an explicit opt-out env, not the default posture with an opt-in strictness flag the child can drop. Separately, fix `stdinIsInteractive` to test for a terminal rather than any character device. Add a mutation test that runs the CLI with `/dev/null` and with a PTY as stdin and asserts refusal.

### Non-blocking

- P5: ship `brigade work authority repair` (and migrate/rotate/key-init) before an error message tells operators to run it, or change the messages to a recovery path that exists. A key-loss or crash-window operator currently has a named-but-absent recovery command.

### Residual-honesty boundary (P3)

Correctly encoded. The suite pins that the default crypto tier accepts a persisted-key-holder forgery and only `--isolated-scanners` refuses it, and the #957 wording does not claim a class close. That part of the PR is right and should be preserved as-is.

Verdict: BLOCKED on P1. The #957 crypto and isolation work is sound; the #1029 capability gate needs to fail closed on a missing capability irrespective of stdin before this closes #1029.

solomonneas:
## Independent probe-executing review: READY

Reviewed against the approved Option A design (issue #957 comment 5347614521, plus #1029, #1006 finding 1, and #1028's mapping). Fresh clone, PR branch at 43bc6ba1, every probe routed through `brigade work verify run --capture brigade-work`. Verdict is READY. No blocking findings.

### #1029 (CRITICAL): trust-review clean-marking cannot self-elevate without a capability

The gate is enforced in two layers: the CLI refusal at `engines/evidence-ledger/internal/app/app.go:427-429` and the engine-side requirement at `engines/evidence-ledger/internal/ingest/provenance.go:262-264`. The capability MAC binds `v/item_id/from_digest/transition(to_label + mark_injection_clean)/nonce/expiry`; `mint_meta` is excluded as audit only (`capability.go:58-73`). Per-field tamper is refused with positive controls in `provenance_test.go:871-939` (missing capability, replay via `used_capabilities`, expiry, item rebind, transition-scope). The parent hands the secret over stdin only; `stdinCarriesHandoffPipe` (`app.go:99-112`) is a read-eligibility check, not authorization, so char-device, PTY, and piped-empty stdin without a capability are all refused.

Passing run:

```
go test -C engines/evidence-ledger ./internal/app/ -count=1 -run TestMutationScannerTrustReviewWithoutCapabilityIsRefused
[completed] exit=0
```

Mutation (removed both the CLI gate at app.go:427-429 and the ReviewTrust requirement at provenance.go:262-264):

```
go test -C engines/evidence-ledger ./internal/app/ -count=1 -run TestMutationScannerTrustReviewWithoutCapabilityIsRefused
[failed] exit=1
```

The test drives piped-empty stdin, `/dev/null`, and a real PTY, asserts refusal names `capability`, verifies the item stays `quarantined`/`pending`, and has a positive control that succeeds with a valid capability. Reverted after.

### #957: authority store rejects same-UID forged writes lacking a valid HMAC

`tests/test_authority_store_hmac.py` refuses unsigned rewrite after a sequence exists, attacker-key rewrite (`key_id` mismatch), and cross-target splice. The copy-public-config self-import-proof path no longer grants trust: `test_copy_public_config_self_import_proof_is_refused` plants a forged `SCANNER_DEFAULTS` receipt plus a store rewrite and asserts `_legacy_import_source_content_identity(...) is None`.

```
.venv/bin/python -m pytest tests/test_authority_store_hmac.py -v
7 passed in 0.24s
```

`authority_key.py` enforces `st_nlink == 1`, `S_IMODE & 0o077 == 0`, `O_NOFOLLOW`, `O_EXCL` on create, 0600/0700 (`authority_key.py:68-93`), and fails closed on a missing key.

### Residual-honesty: present and correct

Both tests exist and assert the right boundary. `test_residual_honesty_crypto_tier_accepts_key_holder_forgery` (line 125) reads the persisted store key, forges a valid envelope at the correct next sequence, and asserts it is ACCEPTED under the default crypto tier. `test_residual_honesty_isolated_scanners_refuses_same_forgery` (line 149) runs the identical forgery from inside an `_scanner_run_one(..., isolated=True)` child and asserts the binding is refused. In this environment unprivileged userns is available, so the isolated test PASSED rather than skipped, exercising the real refusal. There is no false "class closed" assertion.

### Portability

`scanner_isolation.py` degrades cleanly: Windows reports MANUAL, non-POSIX / missing `unshare` / no `CLONE_NEWUSER` / kernel-refused userns all report WARN with distinct reasons, and `doctor_check` is never silent. `pty_windows_test.go` skips the PTY probe under `//go:build !linux`. The full engine suite is green including the non-linux build tag path.

```
go test -C engines/evidence-ledger ./... -count=1
[completed] exit=0
```

### Scope and honesty

The PR body and CHANGELOG say `Fixes #1029`, `Part of #957`, `Refs #881`. No `Fixes #957` anywhere in the tree. The CHANGELOG states the residual plainly: a same-UID process that reads the persisted store key can still forge a valid envelope, that residual is encoded in the suite, and it is not a class close. This matches the design.

### Other checks

```
.venv/bin/python -m pytest tests/test_trust_capability.py tests/test_authority_broker.py tests/test_work_import_provenance_stamp.py tests/test_component_paths.py tests/test_work_cmd_scanners.py tests/test_work_cmd_sweeps.py
[completed] exit=0
```

No hardcoded secrets in the diff (only signatures and variable names). Real Python capability mint uses `secrets.token_hex(16)` nonce and `secrets.token_bytes(32)` secret, never written to env or argv.

PR #1078, opened 2026-08-21
feat(scanner): opt-in external-key HMAC for the authority store, doctor WARN when off (Fixes #957)

### LINKED ISSUE #957: scanner authority store: residual same-uid write class after env scrub

solomonneas:
## What and why

Same-uid scanner children can rewrite the directory-authority binding store by path. File mode is not a boundary against a process that shares the operator uid. The recorded decision on #957 is opt-in isolation, not mandatory.

`authority_store.isolation = "external-key"` in `.brigade/security.toml` HMAC-signs **new** directory-authority bindings with a 0600 key stored outside the workspace (`$XDG_CONFIG_HOME/brigade/authority/store-hmac.key` or `BRIGADE_AUTHORITY_KEY_FILE`). An existing HMAC envelope is **always verified**. Once a target is signed, a user-level sticky marker under `~/.brigade/authority-signed/<target-fingerprint>` (override: `BRIGADE_USER_DIR`) keeps raw unsigned records fail-closed even if the repo-writable flag is flipped off or the envelope is stripped. Downgrade is `brigade security authority downgrade --target <t> --confirm` only: it converts the signed store to an unsigned record, turns isolation off, and removes the sticky marker so a later read cannot recreate the marker. The envelope's inner target fingerprint must match the destination; a copied foreign envelope is refused. A failed durability step restores every artifact that already changed. `brigade doctor` WARNs when the flag is off. A key override whose unresolved path sits under the workspace is refused before any key bytes are written.

Marker containment is **distinct** from the HMAC key rule. The marker must live under the user-level brigade directory (`~/.brigade` or `$BRIGADE_USER_DIR`) and outside the workspace. It is not rejected merely because its path contains `.brigade`.

Fixes #957

## Option A (operator decision 2026-08-22)

Binding comment on #957: ship the user-level sticky marker now. OS-keyring / separate-uid (B/C) move to a separate hardening track.

**Semantics**

- First signed write records `~/.brigade/authority-signed/<sha256(resolved-target)>`. Fingerprint matches the existing directory-authority store filename scheme.
- Verification of a raw, non-enveloped record **fails closed** whenever that marker exists, regardless of `.brigade/security.toml`.
- `brigade security authority downgrade --target <t> --confirm` (TTY also prompts `y/N`) is one recoverable transaction: log intent, unwrap the HMAC envelope to the inner unsigned record, turn `authority_store.isolation` off, drop that target's sequence entry, remove the sticky marker, log completion. Verification is bound to the **destination** target fingerprint. Each artifact is recorded the moment dest bytes change so a durability-step failure restores store / sequence / isolation / marker. `actor` is the **current** operator (`USER`, then `USERNAME`, then `LOGNAME`).
- If the external HMAC key is already gone, downgrade fails with an actionable error and changes nothing (store stays signed and readable once the key is restored).
- Without `--confirm`, or with TTY `n`, the command refuses and writes nothing.
- Doctor still WARNs on flag-off and still names the residual same-uid class (defense in depth, not a hard boundary). When a marker exists it also says: `signed store with isolation off: verification stays enforced; run brigade security authority downgrade to intentionally downgrade`.
- Tests redirect `HOME` to a temp `operator_home/.brigade` (and `BRIGADE_USER_DIR` to that path). Nothing is written to the real user home.

### SECURITY — cross-target envelope accepted

Downgrade verified the shared-key MAC but did not check that the inner record's target fingerprint equals the destination. Copying target B's valid envelope into target A's store file would unwrap B's authorities as A's and remove A's marker.

**Fix:** after MAC verify, require `_authority_target_digest(inner) == target_fingerprint(destination)`. Mismatch refuses before any mutation.

**Test:** `test_downgrade_rejects_cross_target_envelope`.

### Atomicity — durability-step failure skipped restore

Mutation flags were set only after the writer returned. A post-replace fsync/rename failure (or isolation write after dest replace) left `mutated_*` false, so `_restore_all` skipped the already-changed file.

**Fix:** record each artifact at dest-replace time (`on_replaced` + `_downgrade_durability_checkpoint`). Restore walks every recorded change plus a removed marker.

**Test:** `test_downgrade_restores_all_artifacts_after_durability_failure` for store, sequence, and isolation.

### Finding — downgrade removed only the marker

The command unlinked the sticky marker and left the HMAC envelope intact. The next authority-store read re-detected the envelope and recreated the marker.

**Fix:** `ledger.downgrade_external_directory_authority` unwraps the store, turns isolation off, drops sequence, removes the marker, and audits.

**Tests:** `test_authority_downgrade_persists_unsigned_store_without_recreating_marker`, `test_authority_downgrade_fails_closed_when_external_key_removed`.

### HIGH — default `~/.brigade` marker path was wrongly rejected

Marker containment reused `key_path_is_scanner_reachable()`. Markers must live under `~/.brigade`, so that helper always raised at the real default.

**Fix:** marker-specific containment. Fixture now uses `operator_home/.brigade`.

**Tests:** `test_default_home_brigade_marker_is_created_and_blocks_strip`, `test_marker_rejects_workspace_symlink_and_dotdot_escape`.

### HIGH — forged raw reanchor candidates were re-signed

Reanchor now binds verification to the destination target.

**Tests:** `test_reanchor_rejects_forged_raw_candidate_for_marked_destination`, `test_reanchor_accepts_enveloped_candidate_signed_with_real_key`.

### MEDIUM — downgrade dropped the marker before a durable audit

Intent is logged first; completion failure restores marker bytes.

**Test:** `test_downgrade_does_not_remove_marker_when_audit_unwritable`.

### Round 1

- **HIGH:** in-workspace `BRIGADE_AUTHORITY_KEY_FILE` refused before create.
- **MEDIUM:** isolation flag is load-bearing on new writes.

### Round 2

- **HIGH:** existing envelopes always MAC-checked.
- **HIGH:** unresolved-path + symlink + `..` key refusal before any key bytes.

### Round 3 / G5

- Flag on + key outside the workspace: forged same-uid binding identity is `None`.
- Flag off on a signed store + envelope strip: rejected by the sticky marker.

## Type of change

- [x] Bug fix
- [x] New harness adapter / doctor check
- [x] Docs

## Checklist

- [x] Requested suite green after destination-bind + restore: `tests/test_authority_store_hmac.py tests/test_authority_broker.py tests/test_security_cmd.py tests/test_doctor.py tests/test_work_cmd_scanners.py` (**369 passed**)
- [x] `ruff check` + `ruff format --check` + `mypy` on touched files
- [x] Cross-target planted envelope is refused; dest marker and store unchanged
- [x] Durability failure on store / sequence / isolation restores all artifacts
- [x] Confirmed downgrade still unwraps the store; later read does not recreate the marker
- [x] Default `~/.brigade` marker created; strip + flag-off + key-delete still rejects
- [x] Reanchor forged raw candidate still rejected
- [x] Downgrade audit `actor` is the current `USER`
- [x] Tests never write the real user home

Do not merge this PR here. A final Daybreak security review of the downgrade path follows.

<div><a href="https://cursor.com/agents/bc-c51d6a3e-de0e-45b1-8e21-ebaa9fbdb576?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-c51d6a3e-de0e-45b1-8e21-ebaa9fbdb576&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


solomonneas:
## Independent security review (Daybreak Blue, executed probes): BLOCK

Two findings, both probe-backed against head `7b5ad71`:

1. **HIGH - external-key override can write the HMAC key inside the scanner-reachable workspace.** `authority_key.py` rejects only paths containing a `.brigade` component, so an `external-key` config pointing at `<workspace>/operator-authority.key` gets the key created there *before* doctor's workspace check runs. A same-uid (G5) scanner then reads the key, re-signs a forged binding, and `_legacy_import_source_content_identity` accepts the forged identity. Probe: `bind_result: SUCCEEDED`, `key_written_inside_workspace: true`, `forged_identity_accepted_after_attacker_resigns: true`. Fix: reject every path inside the resolved workspace root before key creation, fail closed.

2. **MEDIUM - the security.toml flag controls only doctor output, not signing/verification.** `authority_store_isolation_mode()` is called only by the doctor check; no authority read/write call site references it, and feature-on vs feature-off stores are byte-identically enveloped. The #957 decision is that the opt-in flag *enables* external-key HMAC on the real authority path - as shipped, the feature does not actually isolate anything.

Sendback dispatched to the implementing agent. Re-review before any merge.

solomonneas:
## Independent security re-review (Daybreak Blue, executed probes): BLOCK (round 2)

The prior two findings are closed on the happy path (flag now gates real signing/verification; direct in-workspace key paths refused). Two deeper HIGHs remain:

1. **HIGH - writable flag silently disables MAC verification.** The isolation mode is read from workspace-writable `.brigade/security.toml`. An attacker flips it to `off`, mutates the binding inside an existing HMAC envelope, leaves the stale signature, and verification accepts the forged identity: `ledger.py:1817-1821` returns the inner record without checking the MAC when `hmac_on` is false. Silent downgrade to trust. Fix: **fail closed** - when an HMAC envelope is present on a record, verify it regardless of the flag; the flag may gate whether *new* writes get signed, but must never gate whether an existing envelope is checked.

2. **HIGH - directory-symlink escape writes key bytes before refusal.** `<workspace>/escape-link/k.key` (escape-link = a dir symlink inside the workspace pointing outside): the containment check resolves the candidate first, so the resolved target reads as outside and passes; 65 key bytes are written to the outside target before the directory fsync rejects with ENOTDIR, and `load_key` then loads it through the workspace path. Fix: reject any candidate whose **unresolved** path is under the workspace, and refuse to traverse workspace symlinks, before writing any bytes (atomic create at the validated real path).

Happy-path acceptance did pass (G5 forgery fails end to end with the key outside the workspace; feature-on vs off differ in signing). This is the 2nd review round; finding 1 raises a design question I'm putting to the operator on #957 (flag location). One more scoped fix dispatched.

solomonneas:
## Independent security re-review (Daybreak Blue): BLOCK (round 3) - stopping the sendback loop, escalating to design

Round-2 fixes verified closed: in-workspace key paths (root, nested, workspace symlink, `..` re-entry) all refused with zero bytes written; full G5 attack with the key outside the workspace fails end to end; flag-off+tampered-envelope and stripped-HMAC-while-on both fail closed; 343 tests pass, mutations red.

**Residual, and it is the design crux:** flag-off **plus stripping the envelope to a raw record** still downgrades to plaintext trust (`ledger.py::_unwrap_authority_envelope()` returns a non-envelope payload as trusted when `_authority_hmac_enabled(workspace)` is false), and it stays forgeable even after key deletion. The reviewer's required fix - "make prior HMAC use sticky outside the attacker-writable workspace flag" - is exactly the unsolved problem #957 was opened for: against a **same-uid** attacker, any trust anchor the attacker can also delete (workspace flag, store file, external key dir) is not a real boundary. Per #957's own text, only privilege separation or OS-mediated secrets close this.

**Disposition:** this is a design decision, not a patch (three implementation rounds have demonstrated that). Holding PR #1078 open and BLOCKED pending the operator design decision I've posted on #957. Not dispatching another blind sendback. The two mechanical fixes already landed (workspace-path refusal, fail-closed envelope verification) are worth keeping as part of whatever design lands.

solomonneas:
## Final security review (Daybreak): BLOCK - marker guards the verify path but not the reanchor path

The round-3 flag-off+envelope-strip probe is now CLOSED by the sticky marker. But the same same-uid attack has a different door:

**[HIGH] Forged raw reanchor candidates are re-signed and trusted** (`work_cmd/ledger.py:2002/2029/2034`). `_reanchor_external_directory_authority()` reads candidates without binding verification to the current target. An attacker deletes the current signed store, plants a RAW candidate named for an UNMARKED dummy target (no marker, isolation defaults off -> raw accepted), and the reanchor merges those attacker-controlled bindings into the marked target and re-signs them with the legitimate external key. Probe: `forged_binding_resigned: true`, `forged_identity_accepted: true` (accepted a forged `handoff-ingest` identity). The prior-round G5 guarantee does not hold end to end.
Fix: reanchor must REJECT raw candidates when the destination target is marked or uses external-key isolation; verify a candidate's envelope+sequence with the legitimate key before considering any of its bindings; never auto-upgrade a raw reanchor candidate.

**[MEDIUM] Marker removed before durable audit** (`authority_marker.py:225/238`). `remove_signed_marker()` unlinks the marker first, appends audit second; if the audit write fails the downgrade already happened unlogged (probe made audit.jsonl a directory -> `marker_removed_despite_audit_failure: true`).
Fix: log downgrade intent durably before removal, then completion, or atomically restore the marker if completion logging fails.

Both get end-to-end regression tests. Sending back.

solomonneas:
## Security review (Daybreak) round 4: BLOCK - and the tests were masking it

The reanchor + audit fixes hold. But the decisive round-3 probe is exploitable again AT THE REAL DEFAULT marker location:

**[HIGH] The default `~/.brigade` marker path is rejected** (`authority_marker.py:85-88,119-129`; `authority_key.py:96-106`). Marker containment reuses `key_path_is_scanner_reachable()`, a KEY-specific helper that returns true for ANY path containing a `.brigade` component. The intended default marker path is `~/.brigade/authority-signed/<fingerprint>` - which contains `.brigade` - so marker creation ALWAYS fails at the default. The signed envelope is written before that raises, leaving a signed store with NO marker, and the flag-off + envelope-strip bypass accepts the raw record again (even after key deletion).

**Test masking** (`tests/conftest.py:79-88`): the autouse fixture sets `BRIGADE_USER_DIR` to a temp dir named `user_brigade` (no `.brigade` component), so the suite is green while production fails. The tests never exercised the real default path.

Fix: marker containment must check the path is OUTSIDE the resolved workspace (and under the user dir), NOT reuse the key's `.brigade`-substring heuristic - the marker's own home legitimately contains `.brigade`. Fix the fixture to use a realistic `$HOME/.brigade`-style user dir, and add Daybreak's flag-off+strip-at-default-location probe as a permanent regression test so a fixture quirk can never mask this again.

Sending back. Note to operator: this is round 4; if this round does not close it, Option A warrants a dedicated implementation pass.

solomonneas:
## Two P1s on the downgrade path (one SECURITY) - not merging; these post-date Daybreak's APPROVE

Daybreak approved #1078 at 41c3ccb2. The downgrade-persistence commit (ac6ab401) added AFTER that approval introduced two P1s Daybreak never reviewed:

**[P1 SECURITY] Downgrade accepts cross-target envelopes** (work_cmd/ledger.py). When target A's store contains a valid signed envelope copied from target B (shared key), downgrade verifies the shared-key MAC WITHOUT checking the inner envelope target matches A, then publishes B's record unsigned at A's path and removes A's marker. A same-uid attacker copies B's signed envelope to A, runs downgrade, and scanner paths consume B's directory authorities as A's. Same class as the reanchor finding: bind verification to the DESTINATION target, not the candidate's self-declared identity.
Fix: before unwrapping, verify the envelope's inner target fingerprint == the destination target; reject cross-target envelopes.

**[P1] Failed downgrade leaves partial state** (work_cmd/ledger.py). When a store/sequence write fails during its post-replacement durability step, or the isolation-config write fails after modifying the file, the mutation flag stays false and _restore_all SKIPS the already-changed artifact - leaving an unsigned store behind a marker, a dropped sequence, or a damaged security config. The transaction is not atomic across the three mutations.
Fix: track what was actually mutated (not an intent flag) so _restore_all restores every changed artifact on any failure, including failures during the durability step; the downgrade must be all-or-nothing.

Both get regression tests. A final Daybreak security review of the downgrade path is required before this merges.

PR #1080, opened 2026-08-21
feat(runs): durable child-vs-parent diff over versioned contract (Fixes #1071)

### LINKED ISSUE #1071: runs child diff: compare a durable child run against its parent (learning-replay DIFF leg)
Follow-up to #594, filed per its first-slice scope ("Keep child diff and additional resume behavior under their owning follow-up issues"). The first slice landed in PR #1062 (`brigade runs child`, lineage recording, `runs show` display).

## Scope

Compare a durable child run against its parent (or two siblings sharing a parent) without mutating either side:

- lifecycle state
- worker results
- verification results
- outcome evidence

GraphTrail artifacts are compared only when both sides contain compatible snapshots; otherwise the diff states the skip instead of guessing.

This is the DIFF leg of the learning-replay loop (branch -> rerun -> diff -> feed the outcome ledger): the diff result should be capturable via `brigade outcome capture` so replay learnings feed ranking.

## Acceptance

- [ ] `brigade runs diff <child-run-id>` (name open to `runs child diff`) compares the child against its recorded parent using stored lineage; an explicit two-run form may compare siblings.
- [ ] Output ships as a versioned JSON contract consistent with the #1063 runs contracts, plus the human-readable default.
- [ ] All new payload strings pass through the `_clean_str` chokepoint (home-path redaction, bounds) established in #1063; the payload-walk needle test covers the new contract.
- [ ] GraphTrail snapshot comparison runs only when both sides have compatible snapshots; incompatible or absent snapshots produce a stated skip, not an error.
- [ ] Diff never mutates or invalidates parent, child, or siblings.
- [ ] Unknown, corrupt, or parentless run IDs fail cleanly with no partial output.

## Non-goals

- Automatic merge of child output into the parent (operator-controlled, per #594).
- Retention or cleanup changes.

solomonneas:
Fixes #1071

## What and why

Learning-replay DIFF leg: `brigade runs diff <child-run-id> [other]` compares a durable child against its recorded parent (or two siblings that share a parent) without mutating either side. Output is the versioned `brigade.run-diff.v1` contract plus a human-readable default.

Logic lives in `src/brigade/runs_diff.py` so `runs_cmd.py` stays untouched for concurrent work. Serializers reuse the #1063/#1064 `_clean_str` chokepoint (home-path redaction including Windows and regex edges, plus bounds) and the load-bearing `_merge_source` / `_allowlisted` pairing. GraphTrail snapshot attestations are compared only when both sides are compatible; otherwise the payload states `skipped` with `absent snapshots` or `incompatible snapshots`. Unknown, corrupt, or parentless run IDs fail closed with no JSON.

## Probe-review fixes (verification leg)

- `changed` is decided from verify-receipt content only: `status`, `command`, and `exit_code`.
- `run_id` and the `final.txt` digest are display-only. They do not drive changed/unchanged.
- Equal verify receipts with distinct verify ids and different `final.txt` contents report `unchanged`. A mismatch in status, command, or exit_code reports `changed`.
- The comparison walks the **full ordered receipt/command sequence**, not the first entry. A later receipt mismatch or a different receipt count reports `changed`.
- Putting `digest` back into `_VERIFY_CONTENT_FIELDS` makes the equal-receipt / different-final.txt test RED.
- `test_verify_content_fields_are_receipt_only` pins the frozenset so identity fields cannot re-enter the decision.

## Type of change

- [x] Surface change (new `runs diff` command and `brigade.run-diff.v1` contract) — scoped by #1071

## Issue #1071 acceptance criteria

- [x] `brigade runs diff <child-run-id>` compares the child against its recorded parent using stored lineage; an explicit two-run form compares siblings (or child+parent, normalized so left is the child)
- [x] Versioned JSON contract consistent with the #1063 runs contracts, plus the human-readable default
- [x] All new payload strings pass through `_clean_str`; the full-payload-walk needle test covers the new contract with planted under-bound needles and a positive control
- [x] GraphTrail snapshot comparison runs only when both sides have compatible snapshots; incompatible or absent snapshots produce a stated skip
- [x] Diff never mutates parent, child, or siblings
- [x] Unknown, corrupt, or parentless run IDs fail cleanly with no partial output

## Checklist

- [x] Added or updated tests covering the change (`tests/test_runs_diff.py` plus the contract needle walk in `tests/test_runs_json_contracts.py`)
- [x] Focused pytest: `tests/test_runs_diff.py` — 22 passed
- [ ] Full `./scripts/verify` not run (issue asked for the runs-diff suite only)

<div><a href="https://cursor.com/agents/bc-db3990b1-e728-450a-9b53-e481c2cf683d?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-db3990b1-e728-450a-9b53-e481c2cf683d&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


solomonneas:
## Independent probe review: BLOCK

Six legs verified sound (chokepoint redaction is clean, non-mutation holds, failure paths fail cleanly, GraphTrail skip is stated, scope/hygiene good). One real semantic defect in a required diff leg:

**The verification leg compares on identity, not content.** It keys the verify comparison on the verify `run_id`, which is globally unique, so it reports `changed: true` for every realistic parent/child pair regardless of whether the verification *results* actually differ. A mutation battery confirmed the leg is inert:
- `MV3` (verify status ignored -> always equal): **SURVIVED**
- `MV4` (verify exit_code always 0): **SURVIVED**
- `M9` (`_final_digest` -> `(None, None)`): **SURVIVED**

The suite misses this because its fixture hardcodes a verify `run_id` collision that production cannot mint, so the tests never exercise the real 'same results, different run ids -> unchanged' case.

**Fix:** compare the verification leg on its **content** - status, exit_code, command (and the result digest) - not on `run_id`. Two verify runs with equal results must diff as `unchanged` even though their run ids differ; a difference in status/exit_code/command/digest must diff as `changed`. Update the fixtures to use realistic distinct run ids: one pair with matching content (asserts unchanged) and one with differing content (asserts changed), so MV3/MV4/M9-style mutations are killed. Sendback dispatched.

solomonneas:
## Re-review: BLOCK (round 2) - original blocker fixed, replaced by an equivalent one

The run_id keying is removed and the three surviving mutations (MV3/MV4/M9) are now killed - that part is confirmed. But the content set now includes `digest` = sha256 of the run's `final.txt`. `final.txt` is the run's synthesis output, which differs between any two real runs, so the verification leg still reports `changed: true` for every realistic pair - same symptom, new cause. (Partly my sendback's fault: it said to include 'the result digest'; a digest of the whole final output is not verification content.)

**Fix:** `_VERIFY_CONTENT_FIELDS` must be `{status, command, exit_code}` only. The `final.txt` digest is run-synthesis identity, not verification content - drop it from the changed/unchanged decision (keep it display-only in the view if useful, like run_id). The regression fixtures must use realistic differing `final.txt` contents with equal verify-receipt content and assert `unchanged`.

This is the second sendback on this leg; per the queue guardrail it is the last before the PR parks for a human pass.

solomonneas:
Resolving as intended behavior, not a defect. The verify-leg comparison is defined on {status, command, exit_code} CONTENT (established over two prior review rounds; run_id and final.txt digest are explicitly display-only, not part of the changed decision). Receipt-file GROUPING is likewise not verification content - two runs that executed the same ordered commands with the same statuses produced the same verification OUTCOME whether those commands were recorded in one receipt file or split across two. Preserving receipt-file boundaries would add invocation-count structure to the comparison that the design deliberately scoped out, producing spurious 'changed' for identical verification results. The multi-receipt fix correctly closed the real bug (a differing LATER command/status now reports changed); this grouping case is content-equal and correctly reports unchanged.

Issue #1022, opened 2026-08-19
work inbox imports are dead on Windows: _open_import_inbox_parent raises unconditionally on non-POSIX, taking 21 call sites with it


solomonneas:
## Summary

`_open_import_inbox_parent()` in `work_cmd/ledger.py` raises unconditionally on any non-POSIX platform. Every command that writes a work-inbox import record goes through it, so on Windows the entire import surface is dead — with an unhandled Python traceback, not a handled error.

This is not a corner case. `brigade daily status` reports `top_readiness_blocker: release readiness receipt is missing` and prints `next: brigade center readiness import-issues`. On Windows that command crashes.

brigade 0.27.0, Windows 11 Pro (10.0.26200), pipx, CPython 3.13.

## Reproduce

```
> brigade memory care import-issues --target .
Traceback (most recent call last):
  ...
  File "...\brigade\memory_cmd.py", line 311, in dispatch
  File "...\brigade\work_cmd\services.py", line 170, in _import_memory_refresh_queue
  File "...\brigade\work_cmd\ledger.py", line 2930, in _append_import_records
  File "...\brigade\work_cmd\ledger.py", line 2276, in _snapshot_import_inbox
  File "...\brigade\work_cmd\ledger.py", line 2163, in _open_import_inbox_parent
OSError: descriptor-relative import inbox operations are unavailable
(exit 1)

> brigade center readiness import-issues --target .
OSError: descriptor-relative import inbox operations are unavailable
(exit 1)
```

## Cause

```python
def _open_import_inbox_parent(target: Path, *, create: bool) -> tuple[int, str]:
    """Hold the inbox parent for an import/proof publication transaction."""
    if (
        os.name != "posix"
        or not getattr(os, "O_NOFOLLOW", 0)
        or not getattr(os, "O_DIRECTORY", 0)
        or os.open not in os.supports_dir_fd
        or os.mkdir not in os.supports_dir_fd
    ):
        raise OSError("descriptor-relative import inbox operations are unavailable")
```

On Windows every clause fails:

```
platform: win32
os.supports_dir_fd: set()
hasattr(os, 'O_DIRECTORY'): False
```

There is no non-descriptor fallback branch, so the function can never return on Windows.

[...]

Issue #752, opened 2026-08-08
work brief crashes with UnicodeDecodeError on Windows: _read_text() reads docs without encoding='utf-8'


solomonneas:
## Summary
On Windows, `brigade work brief --target .` (and anything that routes through roadmap health -> audit -> `_documented_brigade_commands`) crashes with an unhandled `UnicodeDecodeError` when any scanned doc contains a byte that is invalid in the machine's locale codec (cp1252). A UTF-8 `README.md` with box-drawing characters is enough to trigger it.

## Environment
- brigade-cli **0.26.0** (pipx)
- Python **3.14.6**
- Windows 11, locale default codec **cp1252**

## Root cause
`brigade/roadmap_cmd.py`:

```python
def _read_text(path: Path) -> str:
    try:
        return path.read_text()      # no encoding= -> uses locale default (cp1252 on Windows)
    except OSError:                   # UnicodeDecodeError is NOT an OSError, so it propagates
        return ""
```

`_public_doc_paths()` feeds this `README.md`, `ROADMAP.md`, `CHANGELOG.md`, and `docs/*.md`. `read_text()` with no `encoding` uses `locale.getpreferredencoding()` = cp1252 on Windows, and `UnicodeDecodeError` is not a subclass of `OSError`, so the guard does not catch it and the whole command dies.

## Concrete repro
Any UTF-8 doc with a non-cp1252 byte. Real case: a `README.md` containing a box-drawing diagram:

- `README.md`, byte `0x90` at offset 19491, part of the UTF-8 sequence `e2 94 90` (`┐`, U+2510 BOX DRAWINGS LIGHT DOWN AND LEFT).
- `Path("README.md").read_text(encoding="cp1252")` -> `'charmap' codec can't decode byte 0x90 in position 19491: character maps to <undefined>`.

Traceback tail:
```
File ".../brigade/work_cmd/session/briefing.py", line 292, in _brief_payload
    roadmap_health = roadmap_cmd.health(target)
File ".../brigade/roadmap_cmd.py", line 457, in _documented_brigade_commands
    commands.update(_commands_from_text(_read_text(path)))
File ".../brigade/roadmap_cmd.py", line 329, in _read_text
    return path.read_text()
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 19491: character maps to <undefined>
```

## Impact
[...]

Issue #357, opened 2026-07-19
Brigade on Windows: port POSIX assumptions and add three-lane testing


solomonneas:
Parent: #352

Phase: 1 through 4

Windows remains in scope for the first unified release. Track the Python CLI port and the release, clean-machine, and debugging lanes in one workstream.

Scope:

- Audit paths, executable suffixes, hook installation, subprocess behavior, encoding, and service or `systemd` references.
- Preserve existing public commands and add compatibility shims where platform-specific behavior changes.
- Use GitHub-hosted `windows-latest` jobs as the only Windows release gate.
- Use a disposable Windows 11 Hyper-V VM with only Python and pipx for clean-machine acceptance.
- Restore a checkpoint named `clean` before every acceptance run and verify Go and Cargo are absent from PATH.
- Keep the maintainer-operated Windows host as a scoped porting and debugging lane only.

Acceptance criteria:

- [ ] The published CLI and native assets pass on GitHub-hosted `windows-latest`.
- [ ] The restored clean VM completes install, setup, doctor, and code-plus-evidence smoke tests without Go or Cargo.
- [ ] No personal host is registered as a self-hosted runner.
- [ ] Host-lane work makes no global PATH, service, or configuration changes.

Issue #281, opened 2026-07-16
agents: Windows npm shims pass detection but can fail dispatch


solomonneas:
## Problem

On Windows, Brigade can report a Codex seat as installed and then fail the same seat with `command not found: codex`.

`agents.detect()` uses `shutil.which()` through `proc.which()`, so it can find an npm-generated extensionless or `.cmd` shim through `PATHEXT`. `run_agent()` then discards that resolved path and passes the bare command name to `subprocess.run()`. Python process creation does not necessarily resolve or execute the same shim, so doctor and dispatch disagree about whether the adapter is runnable.

## Reproduction

Version: Brigade 0.22.0 with Codex installed through npm on Windows.

1. Keep the npm global shim directory on `PATH`, without adding the package's native `codex.exe` directory.
2. Run `brigade roster doctor` for a Codex seat. Detection succeeds.
3. Dispatch the seat. `proc.run()` catches `FileNotFoundError` and returns exit 127 with `command not found: codex`.
4. Add the native executable directory to `PATH`. `shutil.which("codex")` now resolves `codex.EXE`, and the same Brigade dispatch reaches Codex successfully.

A Git workspace control then completed with the expected final marker. A non-Git control reached Codex but hit Codex's own trust check. That second result proves command resolution was fixed by the native executable path and should be classified separately as provider preflight.

## Expected behavior

- Adapter detection and adapter execution use the same resolved executable identity.
- A seat that doctor marks runnable does not fail because dispatch discarded the path doctor found.
- Unsupported Windows shim kinds fail before inference with an actionable command-resolution diagnostic.
- A provider trust refusal after successful launch is reported as provider preflight, not command-not-found.

## Acceptance criteria

- Resolve the adapter executable once and use that exact result for both detection and process launch.
- Handle Windows `.exe`, `.cmd`, `.bat`, and extensionless npm shims explicitly. Either execute a supported shim safely or explain the smallest native-executable `PATH` setup action.
- Preserve the resolved command kind in local worker diagnostics without exposing user-specific paths in public summaries.
- Add Windows-focused tests where `which("codex")` returns a native executable, a `.cmd` shim, and an extensionless shim.
- Add a regression proving doctor and `run_agent()` cannot disagree about the resolved executable.
- Keep a launched provider's non-Git trust failure distinct from command resolution.

## Related issues

- #226 added worker stdout, stderr, and exit-code artifacts.
- #264 fixed Grok result integrity.
[...]

PR #1098, opened 2026-08-22
fix(windows): unblock brigade run by fixing two POSIX assumptions

### LINKED ISSUE #1096: run: os.fchmod on a read-only fd raises PermissionError on Windows, so every brigade run dies writing the lifecycle journal
`brigade run` cannot start on Windows. Every invocation fails during preflight with:

```
error: failed to write initial run receipt: lifecycle journal I/O failure (PermissionError)
```

The run directory is created, `run.json` / `roster.json` / `route-decision.json` are written, and `events/lifecycle.jsonl` is created at 0 bytes before the failure.

## Root cause

`run_journal._enforce_file_mode` opens the journal read-only and then corrects the mode through the descriptor:

```python
fd = _open_nofollow(path, os.O_RDONLY)          # run_journal.py:474
...
if stat.S_IMODE(info.st_mode) != _FILE_MODE:
    _chmod_fd_or_path(fd, path, _FILE_MODE)
```

`_chmod_fd_or_path` picks its branch on attribute presence:

```python
_HAS_FCHMOD = hasattr(os, "fchmod")             # run_journal.py:59

def _chmod_fd_or_path(fd, path, mode):
    if _HAS_FCHMOD:
        os.fchmod(fd, mode)
        return
    _verify_fd_identity(path, fd)
    os.chmod(path, mode)
```

`os.fchmod` exists on Windows under Python 3.14, so `_HAS_FCHMOD` is true, but Windows requires write access on the handle to change attributes. Calling it on an `O_RDONLY` descriptor raises `PermissionError [WinError 5]`.

It is unconditional, not intermittent: Windows reports file modes as `0o666` or `0o444` and never `0o600`, so `S_IMODE(info.st_mode) != _FILE_MODE` is always true and the failing branch is always taken.

Presence is not capability here. Two functions away, `_supports_directory_fsync()` gets this right by gating on `os.name == "posix"` instead of attribute presence.

## Reproduction

Windows 11, Python 3.14.6, brigade-cli 0.27.0, any wired workspace:

```
brigade run --dry-run "anything"
```

Minimal, independent of brigade:

```python
import os, tempfile
d = tempfile.mkdtemp(); p = os.path.join(d, "j.jsonl")
open(p, "w").close()
fd = os.open(p, os.O_RDONLY)
os.fchmod(fd, 0o600)   # PermissionError: [WinError 5] Access is denied
```

The same call succeeds on a descriptor opened for writing, which is why `os.fchmod` looks available in isolation.

## Traceback

```
  File "brigade/run_lifecycle.py", line 764, in prepare_lifecycle_journal
    run_journal.ensure_journal(journal_path)
  File "brigade/run_journal.py", line 528, in ensure_journal
    _enforce_file_mode(journal_path)
  File "brigade/run_journal.py", line 481, in _enforce_file_mode
    _chmod_fd_or_path(fd, path, _FILE_MODE)
  File "brigade/run_journal.py", line 328, in _chmod_fd_or_path
    os.fchmod(fd, mode)
PermissionError: [WinError 5] Access is denied
```

## Suggested fix

Gate on platform rather than attribute presence, so Windows takes the existing verified-path fallback:

```python
_HAS_FCHMOD = hasattr(os, "fchmod") and os.name == "posix"
```

The fallback branch works on Windows. `os.lstat`/`os.fstat` populate `st_ino` and `st_dev`, so `_verify_fd_identity` passes, and `os.chmod(path, mode)` succeeds. `os.chmod` only controls the read-only bit on Windows, so the reported mode stays `0o666` and the comparison stays true on later calls, making the chmod a repeated no-op rather than an error.

Verified locally: with this one-line change `brigade run` writes its artifacts and proceeds past preflight.

Related Windows POSIX-assumption work: #357, #1022, #752, #281.

### LINKED ISSUE #1097: run: localio text-mode writes emit CRLF on Windows, so the writer-canonical byte check rejects every run.json
With #1096 patched locally so `brigade run` can start, the next dispatch on Windows fails with:

```
error: failed to record dispatch lifecycle fact: base bytes differ from writer canonical form
```

This blocks every worker dispatch, including `brigade run --worker <seat>`.

## Root cause

`run_checkpoint` re-encodes the incoming `run.json` and requires byte equality with what was read:

```python
def _writer_canonical_bytes(obj):
    return (json.dumps(obj, indent=2, sort_keys=True) + "\n").encode("utf-8")
...
if _writer_canonical_bytes(incoming_obj) != run_json_bytes:      # run_checkpoint.py:789
    raise CheckpointError(_bound("base bytes differ from writer canonical form"), ...)
```

The canonical form emits LF. The writer emits CRLF on Windows, because `localio` opens the temp file in text mode with no `newline=`:

```python
def write_text_atomic(path, data):                                # localio.py:46
    ...
    with os.fdopen(fd, "w", encoding="utf-8") as handle:          # localio.py:60
        handle.write(data)
```

Python text mode translates `\n` to `os.linesep` on write, so on Windows every file written through this helper lands CRLF. `run.json` is written here (`aboyeur._write_json` calls `localio.write_text_atomic`), so the comparison can never succeed on Windows.

Measured on a real `run.json` produced by 0.27.0 on Windows:

```
bytes: 4093 | CRLF: 150 | bare LF: 0
first 120 bytes: b'{\r\n  "artifacts": "C:\\Users\\...
```

150 CRLF and zero bare LF, against a canonical form that contains only LF.

`write_text_exclusive` (localio.py:118) has the same open call and the same exposure.

## Reproduction

Windows 11, Python 3.14.6, brigade-cli 0.27.0:

```
brigade run --worker <any-seat> "anything"
```

Requires the #1096 fix applied first, otherwise the run aborts earlier in preflight.

## Suggested fix

Pin LF explicitly in both writers. This is a no-op on POSIX, where text mode already writes LF:

```python
with os.fdopen(fd, "w", encoding="utf-8", newline="\n") as handle:
```

Verified locally: with #1096 and this change applied, a real dispatch completes end to end and the worker returns its answer.

Note the two bugs mask each other. #1096 aborts in preflight before any dispatch, so this one is only reachable once that is fixed. A Windows CI lane that exercises a full `brigade run` dispatch, rather than preflight alone, would have caught both. That is the testing gap #357 describes.

Related Windows POSIX-assumption work: #357, #1022, #752, #281.

solomonneas:
Fixes #1096. Fixes #1097.

`brigade run` could not start on Windows, and once it could start, no dispatch could complete. Two independent causes, each one line of source.

## 1. `os.fchmod` on a read-only descriptor (#1096)

`run_journal` picked its chmod branch on attribute presence:

```python
_HAS_FCHMOD = hasattr(os, "fchmod")
```

`os.fchmod` exists on Windows under Python 3.14, but Windows requires write access on the handle to change attributes. `_enforce_file_mode` opens the journal `O_RDONLY`, so the call raised `PermissionError [WinError 5]` and every run died with:

```
error: failed to write initial run receipt: lifecycle journal I/O failure (PermissionError)
```

This was unconditional rather than intermittent. Windows reports file modes as `0o666` or `0o444` and never `0o600`, so `S_IMODE(...) != _FILE_MODE` was always true and the failing branch was always taken.

The fix gates on `os.name`, which is what `_supports_directory_fsync()` two functions away already does, so Windows takes the existing verified-path `os.chmod` fallback. That fallback works: `os.lstat`/`os.fstat` populate `st_ino` and `st_dev`, so `_verify_fd_identity` passes.

## 2. Text-mode writes emit CRLF (#1097)

`localio` opened its atomic temp file in text mode with no `newline=`, so Python translated LF to CRLF on write. `run.json` is written through that helper, while `run_checkpoint._writer_canonical_bytes` produces LF and requires byte equality:

```
error: failed to record dispatch lifecycle fact: base bytes differ from writer canonical form
```

A real `run.json` produced by 0.27.0 on Windows measured 150 CRLF and zero bare LF, against a canonical form containing only LF. The comparison could never pass.

Pinning `newline="\n"` in both text writers fixes it and is a no-op on POSIX, where text mode already writes LF. `write_text_exclusive` had the same open call and the same exposure, so both are changed.

## Why they were not caught

The two masked each other. The first aborts in preflight before any dispatch, so the second is only reachable once the first is fixed. A Windows lane that exercises a full dispatch rather than preflight alone would have caught both, which is the testing gap #357 describes.

## Test results

Measured on Windows 11, Python 3.14.6, running `tests/test_localio.py` and `tests/test_run_journal.py`:

| | failed | passed |
|---|---|---|
| before | 56 | 21 |
| after | 9 | 72 |

47 tests fixed. The failure sets were diffed rather than counted: no test that passed before fails now.

The 9 remaining failures are POSIX mode assertions (`0o600`, `0o700`) that cannot hold where `chmod` only toggles the read-only bit. Deciding how Windows should treat journal modes is a separate question and is left alone here.

`ruff check` passes on all four changed files.

## Added tests

- `test_write_text_atomic_writes_lf_not_platform_newline` and `test_write_text_exclusive_writes_lf_not_platform_newline` assert LF output directly, so a regression fails on POSIX too rather than only on Windows.
- `test_chmod_fd_or_path_does_not_use_fchmod_off_posix` locks the capability gate.
- `test_enforce_file_mode_succeeds_on_a_read_only_descriptor` covers the path that raised, including re-entry against an existing journal.

Also fixed a test-side instance of the same newline family: `blob.write_text("hello\n")` produced CRLF on Windows and broke a sha256 assertion. It now writes bytes.

## Verification beyond the suite

With both fixes applied, a real dispatch completes end to end on Windows against an OpenCode seat:

```
workers:
  [ok] ox_worker
```

Related Windows POSIX-assumption work: #357, #1022, #752, #281.

### DIFF
--- a/src/brigade/localio.py
+++ b/src/brigade/localio.py
@@ -57,7 +57,7 @@ def write_text_atomic(path: Path, data: str) -> None:
     fd, tmp_name = tempfile.mkstemp(dir=path.parent, prefix=f".{path.name}.", suffix=".tmp")
     tmp_path = Path(tmp_name)
     try:
-        with os.fdopen(fd, "w", encoding="utf-8") as handle:
+        with os.fdopen(fd, "w", encoding="utf-8", newline="\n") as handle:
             handle.write(data)
             handle.flush()
             os.fsync(handle.fileno())
@@ -121,7 +121,7 @@ def write_text_exclusive(path: Path, data: str) -> None:
     fd, tmp_name = tempfile.mkstemp(dir=path.parent, prefix=f".{path.name}.", suffix=".tmp")
     tmp_path = Path(tmp_name)
     try:
-        with os.fdopen(fd, "w", encoding="utf-8") as handle:
+        with os.fdopen(fd, "w", encoding="utf-8", newline="\n") as handle:
             handle.write(data)
             handle.flush()
             os.fsync(handle.fileno())
--- a/src/brigade/run_journal.py
+++ b/src/brigade/run_journal.py
@@ -56,7 +56,7 @@ _O_NOFOLLOW = getattr(os, "O_NOFOLLOW", 0)
 _O_DIRECTORY = getattr(os, "O_DIRECTORY", 0)
 _HAS_O_NOFOLLOW = _O_NOFOLLOW != 0
 _HAS_O_DIRECTORY = _O_DIRECTORY != 0
-_HAS_FCHMOD = hasattr(os, "fchmod")
+_HAS_FCHMOD = hasattr(os, "fchmod") and os.name == "posix"
 # O_NOFOLLOW rejects symlinked targets so a pre-placed symlink cannot redirect
 # journal writes or quarantine captures outside the private run-artifact tree.
 _OPEN_FLAGS = os.O_WRONLY | os.O_CREAT | os.O_APPEND
--- a/tests/test_localio.py
+++ b/tests/test_localio.py
@@ -179,7 +179,7 @@ def test_canonical_json_digest_excludes_top_level_keys_and_hashes_files(tmp_path
     assert localio.canonical_json_digest(payload, exclude_keys={"digest", "digests"}) == expected
 
     blob = tmp_path / "blob.txt"
-    blob.write_text("hello\n")
+    blob.write_bytes(b"hello\n")  # bytes: write_text emits CRLF on Windows
     assert localio.file_sha256(blob) == "5891b5b522d5df086d0ff0b110fbd9d21bb4fc7163af34d08286a2e846f6be03"
 
 
@@ -191,3 +191,27 @@ def test_canonical_json_digest_excludes_top_level_keys_only():
     edited_digest = localio.canonical_json_digest(edited, exclude_keys={"digests"})
 
     assert base_digest != edited_digest
+
+
+def test_write_text_atomic_writes_lf_not_platform_newline(tmp_path: Path):
+    """Regression: on Windows, text mode without newline= translated LF to CRLF.
+
+    run.json is written through this helper, and run_checkpoint compares it byte
+    for byte against a canonical form that only ever contains LF, so a CRLF file
+    could never match and every dispatch failed with "base bytes differ from
+    writer canonical form".
+    """
+    path = tmp_path / "run.json"
+    localio.write_text_atomic(path, json.dumps({"a": 1, "b": 2}, indent=2, sort_keys=True) + "\n")
+
+    raw = path.read_bytes()
+    assert b"\r\n" not in raw
+    assert raw.count(b"\n") == 4
+
+
+def test_write_text_exclusive_writes_lf_not_platform_newline(tmp_path: Path):
+    """The exclusive writer shares write_text_atomic's newline exposure."""
+    path = tmp_path / "once.json"
+    localio.write_text_exclusive(path, "first\nsecond\n")
+
+    assert path.read_bytes() == b"first\nsecond\n"
--- a/tests/test_run_journal.py
+++ b/tests/test_run_journal.py
@@ -2384,3 +2384,24 @@ def test_fsync_directory_without_o_directory_opens_readonly_and_fsyncs(tmp_path,
 
     assert fsynced
     assert fsynced[0]
+
+
+def test_chmod_fd_or_path_does_not_use_fchmod_off_posix():
+    """Regression: presence of os.fchmod is not proof it works.
+
+    os.fchmod exists on Windows under Python 3.14 but raises PermissionError
+    (WinError 5) when the descriptor was opened O_RDONLY, which
+    _enforce_file_mode always does. Gate on the platform so non-POSIX takes the
+    verified-path os.chmod fallback instead.
+    """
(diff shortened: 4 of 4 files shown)


solomonneas:
**Windows validation (real hardware: Windows 11 Pro, Python 3.14, head `e720c0a2`)**

Ran against a fresh editable install in an isolated temp target, read-only otherwise. Result:

- **#1096 FIXED** — lifecycle journal creation and repeated writes complete with no `os.fchmod` `PermissionError` / `WinError 5`. On `main` (`b23e0991`) every lifecycle-producing run failed here with exactly that error.
- **#1097 FIXED** — the written `run.json` contains zero CRLF pairs, and canonical recovery checkpoints complete without `base bytes differ from writer canonical form`. On `main` this reproduced as an LF-vs-CRLF hash mismatch at the text-write layer.

Notes (neither is a regression in this PR):
- `brigade run "echo hello"` then hit a separate roster preflight (`exit 2`) because this validation host has no configured agent executable installed — environmental to the box, not related to either fix.
- Nine unrelated test failures remain POSIX-only mode assertions (`0o600`/`0o700`) that are not Windows-portable; none raised the `os.fchmod` error. The full `pytest` suite hangs on this host under captured execution, so only the touched/relevant modules were run (bounded).

Net: this PR resolves the two crashes that made `brigade run` unusable on Windows. Validation driven via the t3 fleet lane.

Issue #1121, opened 2026-08-23
epic: fleet sync - unified multi-machine, multi-agent, multi-harness orchestration


solomonneas:
## Problem

Running many AI coding agents (Claude Code, Codex CLI, opencode/Ox, Cursor, T3 Code) across several machines (rocinante, shadowfax, gandalf), there is no single answer to **"what work is running where, by which agent/harness, at what status."** A conductor polls each machine's lease DB, each `run.json`, and GitHub separately. Two orchestrators on the same fleet only discover each other by noticing a pushed git branch. All operational state is trapped per machine.

Concretely hit while conducting an overnight multi-machine burn: zombie leases (a finished remote task never reconciles, so its repo lease wedges the next dispatch), invisible co-conductors (two orchestrators collided on the same repo, found only via a pushed branch), and no live fleet view.

## Approach: a small central fleet hub on Tailscale (server design)

The no-daemon / zero-runtime-dependency posture is dropped for this feature by decision. The fleet uses a **single small brigade hub service** that everything reports to and queries, modeled directly on how the T3 Code backends already talk to each other:

- Each T3 backend runs `t3 serve --host $(tailscale ip -4) --port 3773` and is reached over the tailnet at `http://<tailscale-ip>:3773` with a bearer token. The brigade fleet hub mirrors this: one HTTP service, bound to its Tailscale IP, bearer-token auth (same shape as the T3 servers and the OpenClaw gateway).
- Store: one SQLite file on the hub. Live queries are plain SQL, so the fleet view is real-time, not poll-and-lag.
- Transport: HTTP over Tailscale (MagicDNS names, works off-LAN). Gandalf (Windows) posts over the tailnet the same way.

### How it works, end to end

1. Every machine keeps its **local run journal as the source of truth** (unchanged), so each machine works standalone and never loses its own history.
2. On each journal append, brigade also fires a one-line **POST to the hub** (`/events`). If the hub is down, events queue locally (store-and-forward) and flush when it is back. The digest chain makes replays idempotent, so nothing is lost or double-counted.
3. The hub appends every event into its SQLite db.
4. `brigade fleet status`, a web page, or your phone (over Tailscale) query the hub for a **real-time** "what is running where" view.
5. **Claims are server-arbitrated:** to work a repo, a machine asks the hub; a unique DB row grants it to exactly one caller, with a TTL so a crashed owner's claim expires. One referee makes coordination simple and correct (no fencing-token / merge-conflict dance).

[...]

PR #1131, opened 2026-08-23
feat(fleet): phase 2 hub, event reporting, store-and-forward, fleet status (#1123)

### LINKED ISSUE #1123: fleet sync phase 2: central fleet hub on Tailscale (`brigade fleet serve`) + event reporting + `fleet status`
Epic: #1121. Requires Phase 1 (node identity).

**Goal:** stand up the central fleet hub and have every machine report to it, modeled on the T3 backends (HTTP on the Tailscale IP, bearer token).

## Scope

1. **`brigade fleet serve`** on the hub (hogwarts LXC CT, decided): a small HTTP service bound to `$(tailscale ip -4)` with a bearer token, backed by one SQLite file. Endpoints: `POST /events` (append), `GET /status` (active runs), health. Same shape as `t3 serve --host $(tailscale ip -4) --port 3773`.
2. **Event reporting from each machine.** On each local journal append, brigade POSTs the event (node_id, run_id, repo, agent/seat, harness, state, ts, digest) to the hub. Reuse the existing journal event as the payload; denormalize the identity fields onto it so the hub view is a group-by.
3. **Store-and-forward.** If the hub is unreachable, queue events locally and flush on reconnect. The digest chain makes re-POST idempotent (hub dedupes by node_id+run_id+sequence+digest), so nothing is lost or doubled.
4. **`brigade fleet status`** queries `GET /status` and prints "what is running where": node, repo, agent/harness, state, age. Real-time (SQL query on the hub), no polling lag.
5. **`machines.toml` / hub config:** the hub's tailnet address + token, mirroring how the T3 config carries `t3_url` + auth.

## Resilience
Local journal writes never block on the hub. A hub outage only stops the live aggregate view; machines keep working and flush their queue when it returns. The hub is an aggregator, not the source of truth.

## Acceptance
- Hub runs on the tailnet with bearer auth; a machine POSTs events over Tailscale and they appear in `GET /status`.
- `brigade fleet status` shows live runs from more than one machine.
- Killing the hub mid-run does not error the machine; events flush on restart with no duplicates.
- Gandalf (Windows) reports over the tailnet the same way.

## Open questions (see epic)
Hub host (hogwarts CT vs rocinante); token source.

solomonneas:
Finishes the #1123 worker draft (c176c40d) against current main APIs and closes the gaps between the draft and the issue's Scope/Acceptance. Epic: #1121. Stdlib only (`http.server` + `sqlite3` + `urllib`), no new dependencies.

## What changed

**Hub — `brigade fleet serve --host <ip> [--port 3774] [--db PATH] [--token-file PATH]`** (`src/brigade/fleet_hub.py`)
- `GET /health` unauthenticated; `POST /events` (single object or array, `{accepted, duplicate}`, dedupe on `UNIQUE(node_id, run_id, sequence, digest)`); `GET /status` latest state per `(node_id, run_id)`, `?all=1` includes terminal runs.
- SQLite WAL + `PRAGMA user_version` schema version; a DB written by a newer hub is refused rather than reinterpreted.
- Fixed: terminal states now match `run_events.EVENT_TYPES` (`run.completed`/`run.failed`/`run.interrupted`); the draft filtered on names that don't exist. `?all=` parsed with `parse_qs`. A batch with one invalid event is rejected whole.
- Bearer token from `BRIGADE_FLEET_TOKEN` or `--token-file`; `--host` is required (never binds all interfaces by default).

**Client** (`src/brigade/fleet_client.py`, hooked from `run_journal.append_event` after the write, outside the lock)
- Fixed precedence bug: with `BRIGADE_FLEET_HUB_URL` set, the draft skipped `fleet.toml` entirely so `token_file` was never read. Now env overrides the file per value.
- Fixed: the draft called `node.find_identity_for_path`, which does not exist — every journal event raised inside the broad `except` and was silently dropped (not even spooled). Replaced with a real ancestor walk for `.brigade/node.toml`, falling back to the brigade-home identity.
- Latency contract: exactly one HTTP request per append, 2s timeout. When a spool exists the new event is appended to it and the spool flushed in order (100/request), so the hub never sees a node's events out of sequence; a partial flush keeps the undelivered tail in order. `repo` is denormalized from the journal's workspace. No hub configured → no-op, no spool dir created.

**CLI** — `brigade fleet serve|status|flush`, `status --all --json`, real `age` column; group registered in the help listing and `docs/command-inventory.md` (mirrors how `node` was added in #1129).

**Docs** — `docs/fleet-sync.md` with Phase 2 status "built, hub CT pending". Note: the brief pointed at `docs/plans/fleet-sync-design.md`, but `docs/plans/` is gitignored (`.gitignore:71`) and that file has never existed in history, so the tracked doc lives at `docs/fleet-sync.md`. CHANGELOG entry under Unreleased.

## Verified

```
PYTHONPATH=src .venv/bin/python -m pytest tests/test_fleet_sync.py tests/test_run_journal*.py tests/test_cli*.py tests/test_node.py -q
178 passed in 38.11s
```
- `tests/test_fleet_sync.py`: 29 tests (was 12 with 2 failing) — hub auth/dedupe/batch/terminal filter/WAL/schema refusal; client env-vs-file precedence (both directions), unreachable-hub spool, wrong-token spool, in-order flush before new event, partial flush ordering, no-config no-op; journal hook end to end (real `node_id` + `repo` in the spool, raising reporter never breaks the writer); CLI (`serve` requires `--host`, defaults 3774 / `~/.brigade/fleet-hub.db`, flags pass through, no-token exit 2 before binding, `status` table + `--json`, `flush`).
- `ruff check` clean on all touched files.
- Windows: paths via `pathlib`, no `chmod`/mode assertions in the fleet files; a separate agent is validating this branch on Windows.

## Not in this PR
- Provisioning the hogwarts LXC CT (operator task). Acceptance items that need the live CT (events over the tailnet from >1 machine, gandalf reporting) are exercised here only against a loopback hub in tests.

## Review HOLD follow-up (02371a55)

All five blocking items from the correctness probe are addressed, each with tests (`TestSpoolHardening`, `TestHubHardening`):
1. Partial flushes rewrite via temp file + `os.replace`; a flush that delivers nothing does not touch the file.
2. Spool mutations take a per-node lock file (`fcntl.flock` / `msvcrt.locking`) plus a process lock; a threaded append-while-flush test delivers every event exactly once.
3. Spool capped at 16 MiB (oldest dropped, logged on `brigade.fleet`); non-retryable 4xx batches are dropped as poison; 401/5xx/network keep the spool.
4. The POST runs on a worker thread joined with a hard deadline, so DNS, connect, and a slow-drip response are all bounded by the 2s budget (tests: hung `getaddrinfo`, delayed status line). Documented in the `fleet_client` module docstring.
5. The journal hook catches `BaseException`; `KeyboardInterrupt`/`SystemExit` propagate only after the event is durable in the spool (tests at both the client and `append_event` level).

Non-blocking items folded in: empty-string / negative-sequence rejection, `hmac.compare_digest`, handler `timeout = 30`, `sqlite3` errors → 500, one `/status` row per `(node_id, run_id)`, autouse conftest fixture clearing the fleet env vars, per-user identity fallback path fixed (`~/.brigade/node.toml`). Also folded in the Windows validation findings (empty env = unset, backslash `token_file` tolerated, gandalf's test fix cherry-picked).

```
ruff check . / ruff format --check . / mypy: clean
pytest tests/test_fleet_sync.py tests/test_run_journal*.py tests/test_cli*.py tests/test_node.py -q
199 passed
```

Fixes #1123


solomonneas:
Correctness probe (Opus, read-only) on d2342dcc. Verdict: HOLD on the five blocking items; non-blocking items are worth folding in where cheap.

## Blocking

**1. Every failed report rewrites the whole spool, non-atomically — `src/brigade/fleet_client.py:204-208`**
Once a spool exists, `report_event` takes the spool-exists branch (`:236-239`) → `flush_spool` → `_read_spool` reads the entire file, the POST fails, `index == 0`, so `path.write_text("\n".join(kept))` truncates and rewrites *every* spooled event. Failure scenario: hub down for an hour, 5k events spooled; each new journal append now reads 5k JSON lines, waits 2s, and truncate-rewrites 5k lines. Any crash, SIGKILL, or full disk inside that truncate window loses the entire spool (or leaves a partial last line, which `_read_spool` silently drops at `:163-165`). Needs tmp-file + `os.replace`, and no rewrite at all when `index == 0`.

**2. Concurrent runs on one node race on the same spool file — `src/brigade/fleet_client.py:150-153` vs `:204-208`**
`spool_path` is keyed by `node_id` only, so two `brigade run`s on the same machine share `<home>/fleet-spool/<node>.jsonl`. Process A reads the spool, POSTs, then truncate-writes; process B `_spool_append`s in between. B's events are silently erased by A's write, or B's O_APPEND lands mid-rewrite and produces an interleaved line. Spec says "flushes in order without duplicates"; there is no lock (the repo already has `_append_critical_section` in `run_journal.py:770` for exactly this).

**3. Permanent rejection is a poison pill with no spool cap — `src/brigade/fleet_client.py:198-201`**
`flush_spool` `break`s on any exception, including `HTTPError 401/400`. Failure scenario: token rotated on the hub (the case `test_wrong_token_spools_without_raising:293` covers only for one event). Every subsequent journal event appends to the spool, retries, gets 401, and rewrites the file. The spool grows without bound, journal appends get monotonically slower, and nothing ever tells the operator. Need a size/age cap and a drop-on-4xx (non-401) path.

**4. The 2s bound does not cover name resolution — `src/brigade/fleet_client.py:145`**
`urlopen(timeout=)` sets the *socket* timeout; `getaddrinfo` runs before the socket exists and is unbounded. `hub_url = "http://hub.tailnet:3774"` with a dead resolver blocks `append_event` for the glibc resolver budget (up to ~40s with two nameservers), directly violating "never block >2s". It is also a per-operation timeout, so a slow-drip response exceeds 2s in total. `docs/fleet-sync.md:78` recommends `tailscale ip -4`, so IPs are the happy path, but a hostname is legal config.

**5. `except Exception` does not cover `KeyboardInterrupt` — `src/brigade/fleet_client.py:246`, `src/brigade/run_journal.py:829`**
Ctrl-C during the 2s POST (a wide window, once per event, hub down) propagates out of `append_event` *after* the journal line was fsynced, so the caller sees a failure for a committed event. If it lands inside the `write_text` at `:206` it also truncates the spool. Should be `except BaseException` on the reporting hook, or re-raise only after the spool is consistent.

## Non-blocking

- **`src/brigade/fleet_hub.py:106`** — `_validate_event` says "must be a non-empty string" but `isinstance("", str)` is True. `node_id=""`, `run_id=""`, `digest=""` are accepted and become dedupe-key values; `sequence=-1` also passes. Add `and value.strip()` / `>= 0`.
- **Test isolation** — `tests/conftest.py` has no autouse fixture clearing `BRIGADE_FLEET_HUB_URL` / `BRIGADE_FLEET_TOKEN`. `_isolate_user_brigade_dir` handles `HOME`, but an operator who exports the documented env override runs the whole suite against their live hub, with +2s per `append_event` in every `run_journal` test. Add `monkeypatch.delenv` for both to an autouse fixture.
- **`src/brigade/fleet_client.py:130-135`** — the fallback `node_mod.load_identity(brigade_home())` resolves to `~/.brigade/.brigade/node.toml` (`node_path` appends `.brigade/node.toml`), a path nothing in the repo ever writes. Any journal outside a `node init`ed workspace silently reports as node `"unknown"`, and every such machine collapses into one row group in `/status`.
- **`src/brigade/fleet_hub.py:214`** — `auth == f"Bearer {token}"` is not constant-time; use `hmac.compare_digest`. Tailnet-only reduces but does not remove it.
- **`src/brigade/fleet_hub.py:206`** — the handler class sets no `timeout`, so a peer that opens a socket and never sends a request line pins a thread in `ThreadingHTTPServer` forever, pre-auth. Set `timeout = 30`.
- **`src/brigade/fleet_hub.py:166`** — `MAX(sequence)` joined back on equality returns two rows for one `(node_id, run_id)` if the same sequence ever arrives with two digests (allowed by the PK). `/status` then shows the run twice.
- **`src/brigade/fleet_hub.py:274`** — `sqlite3.OperationalError` (lock timeout, disk full) escapes `do_POST`; socketserver closes the connection with no HTTP response and a traceback on stderr instead of a 500.
- **Perf, non-correctness** — `init_db` re-runs `PRAGMA journal_mode=WAL`, both DDL statements, and `user_version=` on *every* request (`fleet_hub.py:235`, `:269`); `load_fleet_config` re-reads `fleet.toml` and the token file on *every* journal event (`fleet_client.py:228`, then again inside `flush_spool:187`).
- **Token handling looks clean.** `log_message` is suppressed (`fleet_hub.py:209`), the token never enters a URL or a receipt, and `_load_token` errors carry the path only. No leak found.
- **Unconfigured users:** confirmed no-op apart from one `is_file()` stat per journal append; `test_no_hub_configured_does_not_spool:382` covers it.

**Verdict: HOLD** — blocking on 1-5 (non-atomic full-spool rewrite, spool race between concurrent runs, unbounded poison-pill spool, unbounded DNS blocking the journal writer, KeyboardInterrupt escaping the "never raises" hook).

solomonneas:
Note from the CI failure on d2342dcc (`test (3.10)` only): `toml_compat._fallback_loads` — the parser used when stdlib `tomllib` is absent, i.e. Python 3.10 — evaluates scalar values with `ast.literal_eval`, which applies Python escape processing to TOML *literal* (single-quoted) strings. A Windows path like `token_file = 'C:\temp\tok'` gains control characters on 3.10 while parsing correctly on 3.11+. The fleet test now writes the path as a basic string with escaped backslashes (both parsers agree), and `fleet_client._read_token_file` tolerates backslash paths, but the parser divergence itself is shared code outside this PR's scope — filing-worthy as a small follow-up: special-case single-quoted strings in `_parse_value` to return their raw contents.

PR #1140, opened 2026-08-23
feat(fleet): hub-arbitrated cross-machine repo claims (fleet sync phase 4)

### LINKED ISSUE #1125: fleet sync phase 4: hub-arbitrated cross-machine claims
Epic: #1121. Requires Phase 2 (hub).

**Goal:** stop two machines or two orchestrators from colliding on the same repo, using the hub as a single claim arbiter (much simpler than a coordinator-free fencing scheme).

## Scope
- **`POST /claims`** on the hub: to work a repo/target, a machine asks the hub. The hub grants it via a unique DB row `{owner_node, owner_conductor, target, acquired_at, ttl, renewed_at}`; a second request for a held target is refused with the current owner returned. TTL expiry auto-reclaims a crashed or offline owner's claim. One referee = simple and correct, no fencing tokens or merge conflicts.
- **Check before dispatch:** a conductor queries the hub before dispatching and sees another machine's or conductor's active claim, ending the "invisible co-conductor" collision (two orchestrators found each other only via a pushed branch this session).
- **Renew / release:** a long run renews its claim (heartbeat); terminal state releases it (ties into the Phase 1 reaper).
- **Fallback:** if the hub is unreachable, fall back to the existing local per-repo lease (single-machine safety preserved); cross-machine coordination simply pauses until the hub returns.

## Acceptance
- A claim on machine A is visible to machine B via the hub; B's dispatch against the same target is refused with A shown as owner.
- An expired claim is reclaimable; an unexpired one is not silently stolen.
- Hub-down degrades to local leases without error.
- Tests: claim grant/refuse across two simulated nodes, TTL expiry reclaim, renew keeps ownership.

solomonneas:
Implements fleet sync phase 4 (#1125, epic #1121): the hub becomes the single claim arbiter so two conductors working the same repo on different machines get one winner and one clear refusal.

## What changed

**Hub (`fleet_hub.py`, schema v2)**
- New `claims` table: one unique row per target grants `{owner_node, owner_conductor, acquired_at, renewed_at, ttl_seconds, expires_at}` to exactly one caller. A phase-2 database upgrades in place.
- `POST /claims` (bearer auth) with `action: acquire | renew | release`:
  - **acquire** grants a free or expired target in a single upsert statement (no two concurrent callers can both win), is idempotent for the current owner (extends TTL, preserves `acquired_at` while live), and answers **409 with the current owner** for a held target.
  - **renew** extends only the caller's unexpired claim; an expired or stolen claim is refused, never silently honored.
  - **release** deletes only the caller's row; releasing another owner's claim is a 409.
- `GET /claims` lists active claims (`?all=1` includes expired). Same validation strictness, `hmac.compare_digest` bearer auth, and WAL SQLite as phase 2.

**Client (`fleet_client.py`)**
- `acquire_claim` / `renew_claim` / `release_claim` — one bounded request each (5s hard deadline covering DNS), returning a `ClaimDecision` (`ok` / `held` / `missing` / `no-hub` / `hub-unavailable`), never raising on network failure.
- `repo_claim(target)` context manager: holds the claim for the duration of a run with a daemon renew heartbeat (every TTL/3, default TTL 900s; a claim the hub lost — e.g. hub restore from backup — is re-acquired). Released best-effort on exit.
- `fetch_claims` and `resolve_claim_target` (workspace directory name, the same key the event reporter uses for `repo`).

**Dispatch wiring (`cli/run.py`)**
- `brigade run` enters `repo_claim` immediately before the local `runguard.run_lock` lease. A held target fails with exit 2 and `repo '<x>' is claimed by node <node> until <expiry>`. No hub configured → no-op; hub unreachable → one log line and today's local run.lock behavior. A claim failure is never a new way for a run to fail.

**CLI**: `brigade fleet claims [--all] [--json]` — target, node, conductor, acquired age, expiry.

## Acceptance (from #1125)
- Claim on machine A visible to machine B; B's dispatch refused with A shown as owner ✅ (`TestDispatchWiring::test_refused_when_other_machine_holds_claim`)
- Expired claim reclaimable; unexpired never silently stolen ✅ (`test_expired_claim_is_reclaimable_unexpired_is_not`)
- Hub down degrades to local leases without error ✅ (`test_hub_down_falls_back_to_local_lock`, `test_repo_claim_hub_down_fails_open_with_one_log_line`)
- Tests for grant/refuse across two simulated nodes, TTL expiry reclaim, renew keeps ownership ✅

## Verification
- `PYTHONPATH=src .venv/bin/python -m pytest tests/test_fleet_sync.py tests/test_fleet_claims.py -q` → **76 passed** (25 new)
- `pytest tests/test_run_cli.py tests/test_run_lifecycle.py tests/test_runguard.py -q` → **336 passed**
- `ruff format`, `ruff check`, `mypy` on all touched files: clean

Fixes #1125


solomonneas:
Correctness probe (Opus, read-only) on 678debea. Verdict: HOLD on findings 1 and 2; 3 and 4 strongly recommended before merge.

## Findings

**1. A sibling run on the same node deletes the live claim (split-brain).** `src/brigade/fleet_client.py:770`, `src/brigade/cli/run.py:502`

The claim is keyed only on `(target, owner_node, owner_conductor)` — no per-holder/fencing token — and the claim is acquired *before* the local `run.lock` (run.py:502 → 511). Acquire is idempotent for the same owner (`fleet_hub.py:306`), so a second local run in the same workspace is also granted.

Scenario: run A acquires the hub claim for `ws` and takes the local lock. Run B starts in the same workspace on the same node, is granted the same claim (owner matches), then blocks on `run_lock` with `--run-lock-wait`. Two outcomes, both broken:
- B times out → `RunGuardError` → ExitStack unwinds → `release_claim` at fleet_client.py:770 **DELETEs A's row while A is still running**. A's heartbeat only notices after `ttl/3` (300s default), so another machine can acquire the target and run concurrently for up to 5 minutes.
- A finishes first and releases; B then takes the local lock and runs its whole dispatch **with no hub claim at all** (B's later release is a no-op 200 `{"released": false}`), so a second machine is free the entire time.

This is the case the feature exists to prevent. Fix needs a claim token (e.g. random holder id stored in `claims`, required to match on renew/release) or refcounting, not owner_node alone.

**2. `node_id` is unvalidated and non-unique identities silently grant two machines the same target.** `src/brigade/fleet_hub.py:240`, `src/brigade/fleet_client.py:205`

`_validate_claim_request` accepts any non-empty string as `node_id`; `resolve_node_id` returns the literal `"unknown"` when neither the workspace nor `~/.brigade/node.toml` yields an identity (reachable with `--no-artifacts`, which skips `make_run_dir` → `ensure_identity`, or when identity write fails). Two nodes both reporting `"unknown"` hit the owner-match branch at fleet_hub.py:306 and are *both* granted the same target. Same failure for a cloned VM / copied workspace that carries a duplicate `node.toml`. The hub should reject `"unknown"`/malformed node ids, or the client should refuse to claim (and fall back to local-lock-only) when it has no real identity.

**3. A lost acquire response leaks the claim for the full TTL and the run proceeds unclaimed.** `src/brigade/fleet_client.py:656`, `:726`

`_post_claim_blocking` runs under a 5s deadline. If the hub commits the INSERT and the response is lost (slow tailnet, timeout, connection reset), `_claim_op` returns `hub-unavailable`, `repo_claim` logs one line and yields **without starting the heartbeat and without ever releasing** (finally is only reached on the granted path). Result: the hub holds a claim owned by this node for 900s that nothing renews or releases, blocking every other machine, while the local run has no claim protection. Bounded by TTL, but a 15-minute default block on a repo nobody is holding.

**4. Acquire fails closed when the previous owner releases mid-request.** `src/brigade/fleet_hub.py:314-317`, `src/brigade/fleet_client.py:665`

The upsert loses (rowcount 0), then the owner releases, then `_fetch_claim` returns `None` → hub answers `409 {"granted": false, "owner": null, "error": "... held by unknown"}`. Client maps `owner is None` → `reason="missing"`, which `repo_claim` treats as a refusal and raises `FleetClaimHeldError` → `brigade run` exits 2 with "claimed by node another node until its TTL expires" on a target that is now free. Narrow window, but it is the fail-closed case, and the message is wrong. Retry once on a 409-with-null-owner instead.

## Non-blocking notes

- Expired rows are never pruned (`list_claims` filters but doesn't delete); bounded by target count, not a leak.
- `except Exception` in `_claim_op` correctly routes hub 401/500 to `hub-unavailable`, so a bad token silently disables arbitration rather than failing runs — matches the stated contract, but there's no operator-visible signal beyond one WARNING.
- No-hub path verified clean: `reason == "no-hub"` yields immediately, no heartbeat, no release, no new exceptions; `resolve_claim_target` and `resolve_node_id` are both non-raising. Test coverage for no-hub / hub-down / refusal / heartbeat re-acquire is genuinely good.
- Schema v1 → v2 upgrades in place; a v1 binary against a v2 DB refuses to start (expected, worth a release note).

**Verdict: HOLD** — blocking: finding 1 (sibling run releases a live claim / queued run runs unclaimed) and finding 2 (`"unknown"` and duplicate node_ids grant two machines the same target); findings 3 and 4 are strongly recommended before merge.

solomonneas:
All four review findings addressed in 732d8e02 (plus the expired-row pruning note):

1. **Fencing token + acquire-after-lock.** Every acquisition mints a per-holder token stored in `claims.holder_token`; renew and release require it to match and the hub never echoes it to other callers, so a sibling run sharing the node identity can neither extend nor delete a live claim (`test_holder_token_fences_sibling_release_and_renew`). The hub claim is now taken *after* `runguard.run_lock`, so only the run.lock owner ever holds or releases it — a queued sibling can neither run unclaimed nor unwind the winner's claim (`test_claim_taken_after_local_lock`).
2. **Identity validation.** The hub 400s `"unknown"` and malformed node_ids (`test_unknown_and_malformed_node_ids_rejected`); a client with no usable identity logs a WARNING and stays local-lock-only without contacting the hub (`test_repo_claim_without_identity_stays_local`). The fencing token also closes the cloned-`node.toml` both-granted path.
3. **Lost acquire response.** One retry with the same token (idempotent on the hub → granted, run stays protected: `test_lost_acquire_response_recovers_with_same_holder`); if the hub is still unreachable, a background thread releases any row the lost request committed instead of blocking other machines for the full TTL (`test_lost_acquire_twice_schedules_orphan_release`).
4. **409 with null owner.** Retried once; a second ownerless refusal fails open on the local lock rather than refusing a free target with a wrong message (`test_freed_mid_request_is_retried_once`).

Expired rows are now pruned on every acquire (`test_acquire_prunes_expired_rows`), and an early-v2 claims table without `holder_token` is dropped and recreated (ephemeral TTL state, `test_pre_holder_token_claims_table_is_recreated`).

Verification: tests/test_fleet_claims.py + tests/test_fleet_sync.py → 85 passed; tests/test_run_cli.py + tests/test_run_lifecycle.py + tests/test_runguard.py → 336 passed; ruff format/check + mypy clean; command inventory unchanged.

solomonneas:
Correctness probe #2 (Opus, read-only) on 732d8e02. Prior four HOLD items verified fixed. One new blocker (heartbeat re-acquire after release resurrects the claim for a full TTL).

## Prior HOLD items — all four verified fixed

1. **Per-holder fencing token** — `fleet_hub.py:99` stores `holder_token`; renew (`fleet_hub.py:361`) and release (`fleet_hub.py:377`) both require `holder_token = ? AND owner_node = ?`; the token is excluded from `_CLAIM_COLUMNS` (`fleet_hub.py:301`) so it is never echoed. Acquire's `ON CONFLICT … WHERE claims.expires_at <= ? OR (holder_token = excluded.holder_token AND owner_node = excluded.owner_node)` is one statement, so two racers cannot both win. Params bind in correct positional order.
2. **Real node identity** — `CLAIM_ID_PATTERN` + literal-`unknown` rejection on the hub (`fleet_hub.py:257`) and mirrored client-side (`fleet_client.py:566`), applied to both `node_id` and `holder`.
3. **Lost-acquire recovery** — same-token retry (`fleet_client.py:799`) plus `_schedule_orphan_release` (`fleet_client.py:729`).
4. **409-with-null-owner retry** — `reason == "missing"` is in the retry set at `fleet_client.py:797` and, if refused twice, fails open rather than closed (`fleet_client.py:807`).

TTL/clock handling is sound: all timestamps are server-side epoch seconds, `now` is captured once per request so prune and conflict-check agree, and no client clock enters the decision. Fail-open paths (`no-hub`, `no-identity`, `hub-unavailable`, HTTP 4xx/5xx, timeout) all funnel through `_claim_op`'s blanket `except Exception` → `hub-unavailable`; with no hub configured `repo_claim` returns before any network work. `run.py:509` takes the claim strictly after `run_lock`, so no split-brain in the lock-vs-claim ordering.

## New defect

**Blocking — released claim can be resurrected by the in-flight heartbeat, leaking the target for a full TTL.** `src/brigade/fleet_client.py:852`

`finally:` does `stop.set()` then `release_claim(...)`, but never `heartbeat.join()`, and `_renew_loop` (`fleet_client.py:825-832`) does not re-check `stop.is_set()` before its re-acquire. Sequence:

1. `stop.wait(ttl/3)` returns `False`; heartbeat sends `renew`.
2. Run finishes: `stop.set()`, then `release_claim` → hub `DELETE` commits.
3. The heartbeat's `renew` lands after the delete → 409 `missing`.
4. `_renew_loop` treats that as "hub lost our row" and calls `acquire_claim` with the same token → hub `INSERT` succeeds, `expires_at = now + 900`.
5. Loop re-enters `stop.wait`, returns `True` immediately, thread exits. Nothing ever releases that row.

Result: after the run has already exited, `brigade fleet claims` shows the repo held by the finished node, and every other machine gets `error: repo 'x' is claimed by node <us> until <t+900s>` (exit 2) for the full 900s TTL. Window is one renew round-trip per heartbeat cycle. Only an unbounded-only-by-TTL claim leak, but exactly the leak class this PR is supposed to close. Fix: `heartbeat.join(timeout=…)` before `release_claim`, and/or guard the re-acquire with `if stop.is_set(): return`.

The same race makes `tests/test_fleet_claims.py:412` (`test_repo_claim_heartbeat_renews_and_reacquires`) flaky — it forces `_claim_renew_interval` to 0.05s and then asserts `fleet_client.fetch_claims() == []` immediately after the block exits, so a ~1-3ms local round trip against a 50ms cycle gives it a several-percent failure rate.

## Non-blocking notes

- `src/brigade/fleet_client.py:794,799` — a configured-but-blackholed hub (tailnet down, packets dropped rather than refused) costs `2 × CLAIM_TIMEOUT_SECONDS` = 10s of added startup on *every* `brigade run`, including `--dry-run`. Correct per the fail-open contract, just slow; consider halving the retry timeout. No effect on unconfigured users.
- `src/brigade/cli/run.py:509` never passes `conductor`, so `owner_conductor` is always NULL from dispatch and the `conductor` column in `brigade fleet claims` always prints `-`, despite being advertised in the CHANGELOG.
- `src/brigade/fleet_hub.py:136-139` — the pre-`holder_token` `DROP TABLE claims` migration is unsynchronized: two concurrent requests can both read the old `table_info` and the second `DROP` can destroy a claim the first already granted. Only reachable upgrading from an unreleased early-v2 build; low value to fix.
- There is no `--no-fleet-claim` escape hatch; with a hub configured, the only way to bypass a stuck claim is editing `~/.brigade/fleet.toml` or waiting out the TTL.
- The `docs/runbooks/*` deletions in `git diff origin/main` are an artifact of the branch being one commit behind (`be5be38f`, #1136, landed after the merge-base), not a change in this PR.

**HOLD** — blocking: heartbeat re-acquire after release resurrects the claim (`fleet_client.py:852` / `:827-832`), leaking the target for the full TTL and flaking `tests/test_fleet_claims.py:412`.

solomonneas:
Blocker from probe #2 fixed in 25bd74f3:

**Heartbeat resurrection race.** `repo_claim`'s exit now does `stop.set()` → bounded `heartbeat.join(2 × CLAIM_TIMEOUT_SECONDS + 1)` → `release_claim`, and `_renew_loop` re-checks `stop.is_set()` before its re-acquire (a "missing" during shutdown is our own release landing first, not a hub that lost the row). New regression test `test_release_with_renew_in_flight_never_resurrects` blocks a renew mid-flight across the release and asserts the target stays free — verified to fail (claim resurrected) with the guard and join removed. The join also makes `test_repo_claim_heartbeat_renews_and_reacquires` deterministic: 25/25 consecutive runs of both tests pass.

Notes also taken:
- Dispatch now passes the orchestrator seat as `conductor`, so `owner_conductor` is populated in `brigade fleet claims` (asserted in `test_granted_run_dispatches_and_releases`).
- `CLAIM_TIMEOUT_SECONDS` halved to 2.5s: a configured-but-blackholed hub costs at most 5s per run (two bounded acquire attempts) before failing open.
- The unsynchronized early-v2 `DROP TABLE claims` migration and a `--no-fleet-claim` escape hatch are left as-is per the review's low-value/optional framing; happy to file follow-ups if wanted.

Verification: tests/test_fleet_claims.py + tests/test_fleet_sync.py → 86 passed; tests/test_run_cli.py + tests/test_run_lifecycle.py + tests/test_runguard.py → 336 passed; ruff format/check + mypy clean.

solomonneas:
Correctness probe #3 (Opus, read-only) on 25bd74f3. Verdict: MERGE. Heartbeat/release race closed by two independent guards; all four earlier fixes intact. Findings 1 and 2 below filed as follow-ups.

## Earlier four fixes: intact

1. Fencing token: `holder_token` on the row, required by renew/release (`fleet_hub.py:352-373`), never serialized (`_CLAIM_COLUMNS`, `fleet_hub.py:294`); claim taken *after* `run_lock` (`cli/run.py:502-521`, asserted at `tests/test_fleet_claims.py:574`).
2. Identity: hub rejects `unknown`/malformed `node_id` and `holder` (`fleet_hub.py:263-268`); client mirrors it and falls open with a WARNING (`fleet_client.py:791-794`).
3. Lost acquire: same-token retry + `_schedule_orphan_release` (`fleet_client.py:797-806`).
4. 409-with-no-owner retried, then fails open rather than closed (`fleet_client.py:807-813`).

Grant/renew/release are each one SQL statement under a single writer with `busy_timeout=5000`, so simultaneous callers serialize; TTL is entirely server-clock, no client clock dependence.

## New findings

**1. Self-lockout after a crash (medium)** — `src/brigade/fleet_hub.py:338-350`, `src/brigade/cli/fleet.py:62`
A run killed with SIGKILL leaves an unexpired row owned by *its own* node with a dead holder token. On the next `brigade run` in that repo, acquire's UPSERT `WHERE` fails (unexpired, token mismatch) → 409 → `FleetClaimHeldError` → exit 2 with "claimed by node `<itself>`". Phase 1's lease reconcile frees the local run.lock for a dead owner immediately, but nothing frees the hub row: the machine is locked out of its own repo for 600-900s (heartbeat is TTL/3, so the residual TTL at crash time is 2/3-to-full). `brigade fleet claims` can only list — there is no `release`/`--force` and no `--no-claim` escape; the only manual out is editing the hub's SQLite. Spec-conformant ("crashed owners expire"), but it's the worst practical edge and deserves a follow-up (`brigade fleet claims --release <target>`, or a same-`owner_node` steal when the local lease reconciles as dead).

**2. Heartbeat re-acquire has no orphan-release guard (low)** — `src/brigade/fleet_client.py:837-839`
The initial acquire protects against a request abandoned by the 2.5s deadline that later lands (`_schedule_orphan_release`); the heartbeat's re-acquire does not. Scenario: renew returns `missing`, the re-acquire's HTTP call outruns `CLAIM_TIMEOUT_SECONDS` (plausible — the hub's `busy_timeout` is 5000ms, longer than the client's 2500ms deadline), the loop records `hub-unavailable` and exits on `stop`, `release_claim` deletes nothing, then the abandoned INSERT commits. The row leaks for a full TTL with a holder nobody will release. Self-heals at TTL, low probability, but it's the same hazard the initial path already fixed.

**3. Claim key is the bare directory name (note)** — `src/brigade/fleet_client.py:603-618`
`resolve_claim_target` returns the workspace dir name (or the leaf dir name when there's no `node.toml`). Two unrelated repos that share a basename (`docs`, `web`) block each other across machines with a hard exit 2; conversely two worktrees of the same repo get different keys and are not arbitrated. Matches the event `repo` key, so it's a deliberate tradeoff — worth a line in the docs.

**4. A 401 from the hub falls open silently (note)** — `src/brigade/fleet_client.py:679-680`
`_post_claim_blocking` re-raises non-409 `HTTPError`, so a rotated/wrong token maps to `hub-unavailable`, is retried once, and degrades to the local lock with one WARNING. Correct per the "never a new failure mode" contract, but a misconfigured token disables cross-machine arbitration with nothing louder than a log line; consider surfacing auth failures distinctly in `brigade doctor`.

No fail-closed path found: every hub error, malformed payload, missing identity, and unconfigured-hub case yields and proceeds on `run.lock`; only a genuine 409-with-owner aborts. `docs/runbooks/*` deletions in `git diff origin/main` are a stale-base artifact (merge-base is `c13b2cc5`, before `be5be38f`), not a real removal.

**Verdict: MERGE** — the heartbeat/release race is closed by two independent guards and all four earlier fixes hold; finding 1 (crash self-lockout, no CLI release) and finding 2 (orphan re-acquire) should be follow-up issues, not blockers.

Issue #1124, opened 2026-08-23
fleet sync phase 3: hub-served web Fleet dashboard (Tailscale)


solomonneas:
Epic: #1121. Requires Phase 2 (hub + `/status`).

**Goal:** a web dashboard served by the hub, so the fleet view is a page you leave open or check from your phone over Tailscale.

## Scope
- The hub serves an HTML "Fleet" view from its SQLite: one card per machine (driven by observed `node_id`s, not the three hardcoded `_DEFAULT_HOSTS` in `center_cmd/agent_activity.py`), each row an active run with agent/harness/repo/state/elapsed.
- Reuse the existing Agent Activity contract (already models host/provider/harness/state/parent-child, envelope v2); the data now comes from the hub instead of local-only journals. This removes the current local-relative constraint (`journal_path.relative_to(target.resolve())`) that makes the dashboard single-machine.
- UX modeled on the converged fleet-UI pattern (Conductor, Vibe Kanban, Claude Code Agent View): status board grouped by machine, sort/filter to what needs attention. Read-only.
- Served on the hub's Tailscale IP with the same bearer auth, so it is reachable from a laptop or phone on the tailnet, like the OpenClaw gateway.

## Acceptance
- One page shows live runs across all machines, grouped by node, updating in near-real-time.
- Reachable over Tailscale from another device with the token.
- No regression to the local `center` dashboard when no hub is configured.

## UX polish (level up the page, not just the data)

This evolves the existing `brigade center serve` `/view/agents` page (currently at 127.0.0.1:8876), which already renders the machine-card layout but only with local data. On top of wiring real cross-machine data + Tailscale access:

- **Sort / filter** the run list to what needs attention: by state (running / verifying / needs-attention / failed first), by age (oldest active first), by machine, by repo, by agent/harness. Default view surfaces anything stuck or failed at the top.
- **Claims column** (once Phase 4 lands): show who owns which repo/target right now (owner node + conductor + TTL remaining), so the collision picture is visible at a glance. Before Phase 4, show the local lease owner.
- **Live status affordances**: per-run elapsed timer, a color/badge per state, and a clear "needs attention" flag (paused / awaiting-approval / stale-no-heartbeat).
- **Grouping** by machine card stays the primary layout (it already exists); filters apply within/across cards.

Keep it read-only and dependency-free on the client (server-rendered HTML, matching the current center serve style). No SPA framework.

## Acceptance (additions)
[...]

PR #1145, opened 2026-08-23
feat(fleet): hub-served web Fleet dashboard over Tailscale

solomonneas:
## Summary

Fleet-sync Phase 3 (#1124): the hub serves a read-only web Fleet dashboard so the fleet view is a page you leave open or check from a phone over Tailscale.

- `GET /` and `GET /view/machines` — one card per observed `node_id` (never a hardcoded host list): live run count, when the hub last heard from it, the claims it holds, and a table of its runs (run, repo, seat/harness, state badge, elapsed, last event).
- `GET /view/repos` — per repo: where it is running, its claim (owner node · conductor · TTL remaining, from `/claims`), last outcome, and a `collision` flag when two nodes have live runs on one repo or a live run is not on the claim owner.
- Attention buckets `failed` / `awaiting approval` / `stale` (no event for 30 min) / `running` / `queued` / `interrupted` / `succeeded`; default `sort=attention` puts the first three on top, oldest first. `sort=age|node|repo|state|seat`, substring filters `node= repo= seat= state=`, `attention=1`, `all=1` apply to both boards and carry across the board links.
- Stdlib only, no framework or CDN asset, mirrors the `brigade center serve` page style. Tables/sort/filter are plain HTML + query params (work with JS off); `<meta refresh>` every 10s; the inline script only ticks elapsed timers and adds a client-side text filter.
- Same bearer auth as the JSON endpoints, plus `?token=` once for phone use: 303 to the same URL without the token and an HttpOnly, SameSite=Strict, 30-day `brigade_fleet_view` cookie. The cookie is an HMAC of the token (never the token) and authorizes only the HTML routes — a copied cookie cannot read `/status`/`/claims` or post events/claims; rotating the token invalidates every cookie. Tradeoff documented in `docs/fleet-sync.md` and the hub runbook.
- No token, cookie value, or claim holder token is ever rendered; responses carry the same CSP / `no-store` / `no-referrer` headers as the local dashboard. The local `center` dashboard is untouched.

## Verification

- `PYTHONPATH=src python -m pytest tests/test_fleet_sync.py tests/test_fleet_claims.py tests/test_fleet_dashboard.py -q` → 112 passed
- `ruff format`, `ruff check`, `mypy` clean on `fleet_hub.py`, `fleet_dashboard.py`, the new test file
- Manual: seeded a hub bound to this host's Tailscale IP, exercised `?token=` → 303 + cookie → both boards with curl; confirmed no token/holder in the HTML, cookie refused on `/status`
- No CLI flags added, so `docs/command-inventory.md` is unchanged


solomonneas:
Correctness/security probe (Opus, read-only) on 9228f3a3. Verdict: MERGE, no blocking item. The probe transcript only retained its final line; the one follow-up it named is filed as #1146 (board queries are unbounded and the page refreshes every 10s, so cost grows with the event table).

PR #1147, opened 2026-08-23
fix(fleet): guard heartbeat re-acquire with orphan release (#1142)

### LINKED ISSUE #1142: fleet claims: heartbeat re-acquire lacks the orphan-release guard the initial acquire has
Follow-up from the #1140 (Phase 4, #1125) correctness review. Epic #1121.

## Problem
`_renew_loop` re-acquires on a `missing` renew. If that re-acquire HTTP call outruns `CLAIM_TIMEOUT_SECONDS` (plausible: the hub's `busy_timeout` is 5000ms, longer than the client's deadline), the loop records `hub-unavailable` and exits on stop, `release_claim` deletes nothing, then the abandoned INSERT commits. The row leaks for a full TTL. The initial acquire path already guards this with `_schedule_orphan_release`; the heartbeat path does not.

## Fix
Apply `_schedule_orphan_release` (same holder token) to the heartbeat re-acquire path, and add a test that abandons the re-acquire mid-flight and asserts no claim survives after exit.

Also from the same review, docs-only: the claim key is the bare workspace directory name (matches the event `repo` key), so unrelated repos sharing a basename arbitrate against each other and worktrees of one repo do not; document the tradeoff. And a 401 from the hub degrades to the local lock with one WARNING; `brigade doctor` should surface fleet auth failures distinctly.

solomonneas:
Fixes #1142. Follow-up from the #1140 review.

- `fleet_client._renew_loop`: the heartbeat re-acquire now schedules the same holder-token orphan release the initial acquire has, so a re-acquire abandoned by the client deadline cannot leave a TTL-long claim nobody releases. Regression test forces the delayed re-acquire and asserts no claim survives after the context exits.
- `docs/fleet-sync.md`: documents that the claim key is the bare workspace directory name (unrelated repos sharing a basename arbitrate against each other; worktrees of one repo do not) and that a hub 401 degrades to the local lock with one WARNING.
- `brigade doctor` unchanged: it does not probe the fleet hub.

Verification: authored on the gandalf (Windows) lane; `tests/test_fleet_claims.py` re-run on Linux via `brigade work verify run` at landing (counts in the PR checks). Landed by the conductor.


solomonneas:
Correctness probe (Opus, read-only) on 6ee1b344. Verdict: HOLD on findings 1 and 3 (2 folds into the fix; 4 and 5 are cleanups).

## Findings
**1. BLOCKER — the heartbeat-scheduled orphan release deletes the run's own live claim mid-run.** `src/brigade/fleet_client.py:840-847`
`_schedule_orphan_release` fires unconditionally after `min(ORPHAN_RELEASE_RETRY_SECONDS, remaining)` = **30s by default**, and it releases with the *same* `holder` token the run is still using. At the initial-acquire call site (line 804) that is safe: `repo_claim` yields and the run proceeds with no claim, so any committed row is by definition garbage. At the heartbeat call site the run is still live and `_renew_loop` does **not** return after the failed re-acquire — it hits `if outcome.reason == "hub-unavailable": continue` (line 851-855) and keeps renewing. Nothing cancels the orphan thread when the re-acquire turns out to have committed. `fleet_hub.handle_claim` release is `DELETE FROM claims WHERE target=? AND holder_token=? AND owner_node=?` (fleet_hub.py:370-373) with no liveness check, so the delete lands on the live row.
Failure scenario, stock config (ttl=900, renew every 300s, `CLAIM_TIMEOUT_SECONDS`=2.5, retry=30):
- t=300 renew → 409 `missing` (hub restarted from backup). Re-acquire exceeds the 2.5s deadline → `hub-unavailable` → orphan thread scheduled. The abandoned `_run_with_deadline` worker (fleet_client.py:249, daemon, not cancelled) commits the row at t≈303.
- t=305 next renew succeeds. Row live, run correctly protected.
- **t=330** orphan thread calls `release_claim` with the same holder → row deleted while the run is mid-flight.
- t=330-600 target is free; node B acquires and works the same repo concurrently. At t=600 our renew returns `missing`, the re-acquire gets 409 held, `_renew_loop` logs "lost ownership" and exits; our run finishes unclaimed.
Without this patch that run stays claimed for its whole duration, so this is a regression on the exact invariant claims exist to enforce. Same defect via a second path: hub down for a while, orphan thread still armed, a later heartbeat re-acquire succeeds, orphan release then kills it.
Fix shape: the heartbeat call site needs the `stop` event — do the release only after the block exits (`if not stop.wait(...): return` / defer to the `finally`), or record a pending-orphan flag and re-release after `heartbeat.join()`. Leave line 804 as-is.
**2. Unbounded orphan threads, none cancellable or joined.** `src/brigade/fleet_client.py:841`, `:755`. Every timed-out re-acquire spawns another daemon thread with its own full-TTL deadline. A flapping hub on a long run leaves several concurrent threads, each an independent chance to fire finding #1. They are daemons and never joined, so on process exit a pending cleanup is silently dropped — tolerable for the initial-acquire path (nothing else will release), but here the `finally` at line 875 already issues an identical `release_claim(target, holder=...)`, so the thread is mostly redundant with it.
**3. Test covers only the benign half.** `tests/test_fleet_claims.py:442`. The new test exits the block immediately, so the abandoned commit lands *after* the exit release — the one ordering where releasing is correct. There is no test for the mid-run ordering (commit lands, heartbeat renews it successfully, run still inside the block), which is where the new code deletes a live claim. Add that case; it will fail today. Minor: the test's determinism rests on a ~40ms margin (exit release at ≈t+0.05 vs. hub commit at t+0.1); under load it passes vacuously rather than flaking red, so not a CI risk, just weaker than it looks.
**4. CHANGELOG bullet is orphaned above `### Added`.** `CHANGELOG.md:10-12`. It sits between `## [Unreleased]` and `### Added` with no blank line, while an `### Fixed` section already exists in Unreleased. It is a fix; move it there.
**5. Docs worktree claim has an unstated caveat.** `docs/fleet-sync.md:92-93`. "Separate worktrees ... have different directory names" holds only when the worktree has its own `.brigade/node.toml` or no ancestor has one. `find_workspace_for_path` (fleet_client.py:165-181) walks to the *nearest ancestor* with `node.toml`; if the per-user fallback identity `~/.brigade/node.toml` exists (`_home_identity_target`, line 184), every claimless directory under `$HOME` resolves to the home directory name and they all arbitrate against each other. Low severity, docs-only.
## Explicitly checked, no defect
- **Guard firing on a successful re-acquire:** correctly gated on `not outcome.granted and outcome.reason == "hub-unavailable"` (line 840); a granted re-acquire schedules nothing.
- **Double release:** exit `release_claim` + orphan release is idempotent — the second gets `200 {"released": false}` → reason `missing` → the loop exits. Harmless on its own.
- **Racing a *later* run's acquire:** impossible. Each `repo_claim` mints a fresh `uuid4().hex` holder (line 795) and the hub's DELETE is token-fenced, so a stale orphan thread cannot touch a subsequent run's row.
- **No-hub path:** unchanged. The new call site is inside `_renew_loop`, reachable only after a granted claim, which requires a configured hub. A `release_claim` returning `no-hub` also terminates the loop immediately (reason ≠ `hub-unavailable`).
## HOLD
Blocking: **#1** (orphan release deletes the live claim mid-run) and **#3** (add the mid-run regression test that pins it). #2 should be folded into the #1 fix. #4 and #5 are non-blocking cleanups.

solomonneas:
Correctness probe #3 (Opus, read-only) on f29f4901. Prior HOLDs (a) closed, (b) mostly closed. Verdict: HOLD on finding 1 (the deferred exit-time orphan release is a daemon thread that sleeps 30s before its first attempt, so in the real CLI it dies with the process and never runs); fix finding 4 too (test leaks a live daemon that reaches the operator's real hub).

## Prior HOLDs
**(a) Orphan release deleting the run's own live claim — CLOSED.** Scheduling moved to the `finally` at `src/brigade/fleet_client.py:871`, after the synchronous `release_claim` at :870. The token is per-`repo_claim` (`holder = uuid4().hex`, :812) and the hub deletes only on `WHERE target = ? AND holder_token = ? AND owner_node = ?` (`src/brigade/fleet_hub.py:418`), so the thread cannot touch a later acquire by the same node/process — a later `repo_claim` mints a new token. No double-release hazard: a second DELETE on an already-freed row returns `200 {"released": false}` → reason `missing`, definitive, loop exits.
**(b) Timing flake — mostly closed.** Test 1 (`tests/test_fleet_claims.py:442`) is fully event-gated: no sleeps, every handoff via `Event.wait(30)`, and the heartbeat terminates deterministically (renew #2 returns `held`). Test 2 has a residual load-sensitivity, below.
## Findings
**1. `src/brigade/fleet_client.py:871` (with `:750`) — the deferred cleanup never issues a release in the real CLI. CONFIRMED, blocking.**
`_schedule_orphan_release`'s `_loop` sleeps *before* its first attempt (`time.sleep(min(ORPHAN_RELEASE_RETRY_SECONDS=30, ...))` at :750). At the initial-acquire call site (:800) that is fine — the run body keeps the process alive. At the new exit call site it is not: the only caller is `src/brigade/cli/run.py:515`, entered on the `lifecycle` ExitStack; everything else (artifact collection, patch validation, receipts) runs *inside* that stack, so the claim context exits milliseconds before `cmd_run` returns and the interpreter exits. The daemon thread is killed inside its first 30s sleep, having sent nothing.
Failure scenario: `brigade run`, hub goes down mid-run; heartbeat renew returns `missing`, the re-acquire exceeds `CLAIM_TIMEOUT_SECONDS` but commits on the hub; `pending_orphan_release` is set. At exit, the synchronous `release_claim` (:870) also fails (`hub-unavailable`), the orphan thread is spawned and immediately dies with the process. The committed row survives the full 900s TTL, blocking every other node — the exact #1142 leak the CHANGELOG says is now prevented. The only case actually covered is "hub recovered by exit time", and that is covered by the unconditional :870 release alone, not by the new code. Fix is cheap: attempt the release before the first sleep (move `time.sleep` after a failed attempt), and/or gate scheduling on :870 returning `hub-unavailable` and give it one bounded synchronous attempt or a short non-daemon join.
**2. `src/brigade/fleet_client.py:870-871` — the synchronous release's result is discarded.** The thread is spawned even when :870 returned a definitive `released`/`missing`, producing a redundant DELETE (or, with fix 1 applied, a pointless extra request per run whenever the flag was ever set). `if release_claim(...).reason == "hub-unavailable" and pending_orphan_release.is_set():` is the intended condition. Low.
**3. `src/brigade/fleet_client.py:869-871` — flag read races the bounded join. PLAUSIBLE, low.** `heartbeat.join(timeout=2 * 2.5 + 1 = 6s)`; the worst in-loop path after `stop.set()` is renew (≤2.5s) + re-acquire (≤2.5s) = 5s plus scheduling overhead. On a contended runner the heartbeat can set `pending_orphan_release` after the join expires; the main thread then reads it unset and schedules nothing. Silent, and yields the same TTL-long leak. A `stop`-aware final check (or setting the flag before issuing the re-acquire rather than after) closes it.
**4. `tests/test_fleet_claims.py:524` (`test_timed_out_heartbeat_reacquire_stays_held_until_exit`) — leaks a live daemon thread past the test.** Unlike test 1 (:446), this test does not patch `ORPHAN_RELEASE_RETRY_SECONDS`. Its `tracking_schedule` → `real_schedule` spawns the real 30s-sleeping loop with `ttl_seconds=60`, so it survives ~60s / 2 attempts after the test returns. By then monkeypatch has restored `HOME` and unset `BRIGADE_FLEET_HUB_URL`, so `release_claim` re-reads the operator's real `~/.brigade/fleet.toml` and POSTs a release (bearer token included) to their production hub. Harmless to data (holder-fenced) but it is an outbound request from the test suite to a real host, plus stray threads and sockets during unrelated tests. Add `monkeypatch.setattr(fleet_client, "ORPHAN_RELEASE_RETRY_SECONDS", 0)` as test 1 does.
**5. `tests/test_fleet_claims.py:566-574` — residual CI-under-load flake risk. PLAUSIBLE.** With `_claim_renew_interval` patched to `0` and `controlled_renew` returning `real_renew` for every call ≥2, the heartbeat busy-loops real HTTP renews from `renew_succeeded` until the `with` block exits. The main thread's `fleet_client.fetch_claims()` assertion at :573 runs into that with a 2.0s `REPORT_TIMEOUT_SECONDS` and raises `FleetClientError` on timeout. Not a deadlock (hub is WAL + `busy_timeout=5000`), but it is the same class of load-sensitivity as the prior HOLD. Cheap fix: have `controlled_renew` return `held` (or gate on an event) once `renew_succeeded` is set, so the loop stops like test 1's does.
## Clean
- No-hub / no-identity paths untouched: both return before `stop` and `pending_orphan_release` are created (:797-:806).
- The guard does not fire on a successful re-acquire in a harmful way: the flag is set only on `granted=False and reason == "hub-unavailable"` (:841-842), and a later successful renew leaves it set but the exit-time release is holder-fenced and idempotent.
- Docs: `docs/fleet-sync.md:89-101` matches `resolve_claim_target` (:603) and `find_workspace_for_path` semantics, and the 401 → one WARNING → local-lock claim matches `_claim_op` (401 raises → `hub-unavailable`, retried once, then :798). Minor omission: a 401 hub also arms the initial-acquire orphan loop, which will re-POST every 30s for the whole TTL and never get a definitive answer — pre-existing, not introduced here.
**HOLD.** Blocking: finding 1 (deferred cleanup cannot run before process exit, so the advertised retry does not exist in the CLI). Should-fix before merge: finding 4 (test reaches the operator's real hub). Findings 2, 3, 5 are follow-ups.

solomonneas:
Correctness probe #4 (Opus, read-only) on 9efe300d. Verdict: MERGE. All three prior HOLDs verified closed. Findings 1 and 5 filed as a follow-up; 2-4 are polish.

## Prior HOLDs: all three closed
**HOLD 3 (daemon retry died with the process).** Closed. `src/brigade/fleet_client.py:874-882` is now a synchronous inline loop in the `finally`, not a thread. Join budget widened to `2*T + max(1.0, T)` = 7.5s (`:872`); worst-case in-flight heartbeat work after `stop.set()` is renew + re-acquire = 2×`CLAIM_TIMEOUT_SECONDS` = 5s, since `_run_with_deadline` (`:233-256`) hard-caps each. 7.5 > 5, so
**HOLD 3 (test busy-looped renews).** Closed. With `_claim_renew_interval → 0`, every `controlled_renew` iteration is gated: call 1 blocks on `allow_missing`, call 2 blocks on `reacquire_finished` (test 2) or returns terminal `held` (test 1), call 3+ returns `held` which exits the loop. At most one ungated iteration.
**HOLD 1 (mid-run deletion of the run's own live claim).** Closed. Nothing releases during the run; `pending_orphan_release` only records intent. Test 2 (`tests/test_fleet_claims.py:551-609`) proves the claim survives the late re-acquire and a subsequent successful renew.
**HOLD 2 (test timing flakiness).** Closed. Both tests are event-driven end to end (`reacquire_started` → `timeout_observed` → `allow_reacquire` → `reacquire_finished`), no sleeps, no wall-clock assertions, 30s waits are ceilings not delays.
## Checks you asked for specifically
- **Racing a later legitimate acquire / deleting a live claim:** not possible. `holder` is a fresh `uuid4().hex` per `repo_claim` invocation, and the hub's release is `DELETE ... WHERE target=? AND holder_token=? AND owner_node=?` (`src/brigade/fleet_hub.py:416-419`). A retry can only ever delete this context's own row; a row re-acquired by anyone else (or a later run) returns 409 `held` and is un
- **Double release:** safe. Second DELETE matches 0 rows → 200 `{released: false}` → reason `missing` → loop breaks. Bounded at 3 total calls.
- **Guard firing on a successful re-acquire:** the flag is never cleared, so yes it stays set after a successful re-acquire — but the only consequence is re-issuing a release we already wanted to succeed. No harm.
- **No-hub path:** untouched. `repo_claim` returns before `stop`/`pending_orphan_release` exist when `hub_url` is empty or the node id is unclaimable; `no-hub` / `no-identity` reasons never equal `hub-unavailable`, so the retry loop can't fire.
## Findings (none blocking)
1. `src/brigade/fleet_client.py:874` — **residual leak window the guard doesn't cover.** The retry is gated on `release_outcome.reason == "hub-unavailable"`. If the straggler acquire is still in flight when the exit release lands, the release gets a *definitive* answer (`200 {released:false}` → reason `missing`), the loop is skipped, and the abandoned POST then commits a row that lives the ful
2. `src/brigade/fleet_client.py:874-882` — **silent teardown stall, no operator signal.** Worst case `brigade run` exit is now 7.5s join + 3×2.5s releases ≈ 15s (was 8.5s) with zero logging, including on Ctrl-C unwind through the `ExitStack` at `src/brigade/cli/run.py:516`. A `_LOG.warning` when the retries exhaust would tell the operator a claim may have leaked.
3. `src/brigade/fleet_client.py:825,842` (minor) — the `pending_orphan_release` flag buys nothing in the branch it guards: if the exit release fails with `hub-unavailable`, retrying is correct whether or not a re-acquire happened. Dropping the flag removes a thread-shared object and the "never cleared" wart.
4. `docs/fleet-sync.md:89` — the new "Phase 4 claims" section is inserted *above* "Phase 3 surface" (`:104`), and says nothing about the orphan-release/exit-cleanup behavior this PR implements. Content is accurate (401 → `HTTPError` → `except Exception` → `hub-unavailable` → one warning + local-lock fallback), just misplaced and incomplete.
5. Pre-existing, out of diff, same bug class: `_schedule_orphan_release` (`:745-758`) still sleeps `ORPHAN_RELEASE_RETRY_SECONDS` before its *first* attempt on a daemon thread. On the initial-acquire path that's usually fine (the run outlives 30s), but any `brigade run` that finishes in under 30s after a hub-unavailable acquire hits exactly the HOLD-3 failure this PR fixed elsewhere. One-line fix:
Not run: tests (per instruction).
**MERGE.** No blocking items. Finding 1 is a real residual window worth a follow-up issue (or a two-word guard change now); 2-5 are polish.

PR #1149, opened 2026-08-24
fix(fleet): claim release, same-node reconcile steal, --no-fleet-claim

### LINKED ISSUE #1141: fleet claims: crash self-lockout needs a CLI release and same-node reconcile steal
Follow-up from the #1140 (Phase 4, #1125) correctness review. Epic #1121.

## Problem
A run killed with SIGKILL leaves an unexpired hub claim owned by its own node with a dead holder token. The next `brigade run` in that repo hits acquire's UPSERT guard (unexpired, token mismatch), gets 409, and exits 2 with "claimed by node <itself>". Phase 1's lease reconcile frees the local run.lock for a dead owner immediately, but nothing frees the hub row, so the machine is locked out of its own repo for the residual TTL (600-900s). `brigade fleet claims` can only list; there is no release, no `--force`, and no `--no-claim` escape. The only manual out is editing the hub SQLite.

## Fix
- `brigade fleet claims --release <target>` (bearer-authed, owner_node must match unless `--force`).
- When the local lease reconcile determines the previous owner on this node is dead, allow a same-`owner_node` steal on acquire (new holder token) instead of a 409.
- A `--no-fleet-claim` run flag as the documented escape hatch.

## Acceptance
- SIGKILL a run, start another in the same repo on the same node: proceeds without waiting for TTL.
- Releasing another node's claim without `--force` is refused.

solomonneas:
Fixes #1141 (follow-up to the #1140 / Phase 4 correctness review, epic #1121).

## Problem
A run killed with SIGKILL leaves an unexpired hub claim owned by its own node with a dead holder token. The next `brigade run` in that repo hits acquire's UPSERT guard, gets 409 "claimed by node <itself>", and the machine is locked out of its own repo for the residual TTL (600-900s). The only manual out was editing the hub SQLite.

## Fix
**(a) `brigade fleet claims --release <target> [--force] [--json]`** — bearer-authed. `POST /claims` `release` gains a `scope` field: `"node"` deletes the row without a holder token *only* when the caller's `node_id` owns it (another node's claim → 409 naming its owner); `"force"` deletes whoever holds it. `holder` is optional for those scopes; `renew` is never widened (a token-less renew would let any same-node process keep a claim alive). The deleted row comes back as `claim`.

**(b) same-node supersede on acquire** — `runguard.run_lock` gains an `on_reconcile` callback that fires for each dead-owner lock the acquire reconciled. `brigade run` feeds that verdict into `fleet_client.repo_claim(supersede_same_node=...)`: the acquire is sent with `scope: "node"`, and the hub replaces an unexpired claim owned by the *same* `node_id` under the new token (response carries the replaced row as `superseded`; logged once). Never another node's claim. A same-node claim with no dead lock to vouch for it (sibling run, or a run already `runs recover`ed) is still refused — the error now names `brigade fleet claims --release <target>` and `--no-fleet-claim` instead of "wait for TTL".

**(c) `brigade run --no-fleet-claim`** — skips the hub claim, relies on the local run lock alone, one WARNING on `brigade.fleet`; forwarded to the detached child.

## Tests (`tests/test_fleet_claim_release.py`, 17 new)
- Hub: node-scope acquire supersedes own dead holder only (other node still 409; dead token fenced out of renew; idempotent retry reports nothing superseded); node-scope release needs no token but same owner; force releases any owner; scope validation (force on acquire, node on renew, bogus, token-less holder release, `unknown` node).
- Client: token-less release by node/force; `repo_claim(supersede_same_node=True)` supersedes + logs once; refusal hint names the way out; supersede never steals another node.
- CLI: own-node release text + JSON; **another node's claim refused without `--force`** (claim survives), then `--force` frees it; missing target, `--force` misuse (exit 2), no identity, hub down, no hub.
- Dispatch: **SIGKILL-style** — dead-pid `run.lock` + unexpired own-node claim (default 900s TTL) → `brigade run` proceeds, asserts elapsed < TTL/10, dead run terminalized, claim released on exit; own stale claim *without* a dead lock still refused with hint; dead lock never steals another node's claim; `--no-fleet-claim` runs past another node's claim, leaves it untouched, logs exactly once; detached argv forwarding; `run_lock(on_reconcile=)` fires for a dead owner, not for free/live locks.

## Verification
```
ruff format / ruff check  — clean (touched files)
mypy src/brigade/{runguard,fleet_hub,fleet_client,cli/fleet,cli/run}.py — Success: no issues found in 5 source files
PYTHONPATH=src .venv/bin/python -m pytest tests/test_fleet_claims.py tests/test_fleet_sync.py tests/test_fleet_claim_release.py -q
  103 passed in 46.16s
+ tests/test_runguard.py tests/test_run_detach.py (run_lock / detach argv seams): 257 passed total
BRIGADE_EXTRAS=1 brigade roadmap commands --write — no diff (inventory lists command paths, not flags); --check clean
```

Docs: CHANGELOG (Unreleased / Fixed), `docs/fleet-sync.md` (Phase 4 surface section; Phase 4 table row updated from "not started" to shipped since #1140 merged), `docs/technical-guide.md` (`--no-fleet-claim` in the common run flags).


solomonneas:
Correctness probe (Opus, read-only) on bf030169. Verdict: HOLD on finding 1 (supersede evidence is keyed to the lock path while the steal acts on the claim target; nothing links them). Findings 2-6 are follow-ups unless cheap.

## Findings
**1. BLOCKING - the supersede's "dead" evidence is keyed to the lock path, but the steal acts on the claim target, and nothing links the two.**
`src/brigade/cli/run.py:547` sets `supersede_same_node=bool(reconciled_dead_owners)`, where the evidence comes from `runguard._acquire_lock` (`src/brigade/runguard.py:1123`) reconciling the lock at `lock_path(cwd)` = `git_root(cwd)/.brigade/run.lock` (absolute path). The claim it then supersedes is keyed by `resolve_claim_target` = the workspace **directory basename** (`src/brigade/fleet_client.py:606-621`). The hub's supersede predicate is `owner_node` equality only (`src/brigade/fleet_hub.py:431`). Nothing on the claim row records which run/lock owns it, so the hub cannot verify the reconciled lock is the superseded claim's holder.
Failure scenario: node N has `~/a/api` and `~/b/api` (same basename, both falling back to the per-user `~/.brigade/node.toml` identity, which is the default when a workspace has no `node.toml`). A live run in `~/a/api` holds claim `api`. A new run starts in `~/b/api`; its own lock happens to be dead-owned from an earlier crash, so `on_reconcile` fires and the acquire goes out with `scope: "node"`. The hub replaces the **live** run's row. The victim's next renew gets 409 `held` → `_renew_loop` logs "lost ownership" and returns without re-acquiring (`fleet_client.py:910-916`; only `reason == "missing"` re-acquires). Both runs now proceed on the same claim target with no arbitration, permanently.
Same defect with a nastier blast radius: `node.ensure_identity` deliberately keeps a stored `node_id` "unchanged even when hostname or platform have drifted" (`src/brigade/node.py:146-157`), so a cloned VM image or an rsynced `~/.brigade` gives two physical machines one `node_id`. Machine 2 reconciling its own local dead lock then supersedes machine 1's unexpired claim - the exact invariant the hub is supposed to hold. Local pid liveness proves nothing about the other machine.
Cheap fixes, in order of strength: (a) store the acquiring run's lock `owner_token` (or run_dir) on the claim row and require the client to present the reconciled dead owner's token, so the supersede matches an exact row; (b) failing that, send the reconciled lock's `acquired_at` and refuse to supersede a claim acquired *after* the dead run died - that alone kills the "stole a claim created later by a live sibling" case. No test covers the target-vs-lock-path mismatch today; `TestDispatchRecovery` only exercises one workspace.
**2. `brigade fleet claims --release` has no liveness evidence and no confirmation.**
`src/brigade/cli/fleet.py:135-136` resolves a node id and immediately issues the token-less delete; the hub deletes any row matching `owner_node` (`fleet_hub.py:486`). Nothing checks whether a run is actually alive. Failure scenario: an operator releases a target whose run is hung-but-alive; the live run's heartbeat sees 409/missing and (if another machine acquires first) logs "lost ownership" and continues unarbitrated while the other machine starts on the same repo. Partial mitigation exists by accident: if nobody else claims it within ttl/3, the victim's re-acquire path self-heals. Suggest refusing a non-`--force` release when `runguard` reports a live owner on that workspace's run lock - that is also what would make `--force` mean something.
**3. `owner_node` is caller-asserted; `--force` is gated only in argparse. Acceptable, but the docstrings overstate the guarantee.**
Auth is one shared bearer token, so any fleet member can post `scope: "node"` with someone else's `node_id` and supersede or release their unexpired claim - and `scope: "force"` is available to any token holder regardless of the CLI flag. This is not a *new* trust boundary (the token already authorizes arbitrary claims and events, and `--force` is intentionally offered), so I would not block on it. But `fleet_hub.py:23-33` still reads "an unexpired one is never silently stolen" / "never another node's", and the CHANGELOG repeats it, which is now only true for honest clients. Say so explicitly in the module docstring: `scope` is an intent marker, not an authorization.
**4. Minor - `--release` resolves the node id from cwd, not from the target.**
`cli/fleet.py:135` calls `resolve_node_id()` with no `base_path`, so it uses the cwd's nearest workspace `.brigade/node.toml` (or the home fallback). Releasing a target other than the cwd's workspace, when that other workspace has its own `node.toml`, is refused as "held by node X, not this node", pushing the operator to `--force`, which is deliberately over-broad. A `--node` override or resolving from the target's workspace would fix it.
**5. Minor - version skew message.** A token-less release against a pre-#1141 hub fails validation (`holder` required) → 400 → `_claim_op` maps it to `hub-unavailable` → "fleet hub claim release failed: claim field 'holder' must be a non-empty string", which reads like a client bug rather than "your hub is too old."
**6. Cosmetic - the supersede warning can be lost.** If the first `scope: "node"` acquire commits but the response is lost, the retry (`fleet_client.py:844`) sees `prior[7] == holder` and returns no `superseded`, so the steal happens with no log line.
## Checked and clean
- Hub SQL is correct: `prior[7]` is genuinely `holder_token` (`_fetch_claim` selects it as index 7), placeholder/param counts match (11/11), `renew` is never widened, `CLAIM_ACTION_SCOPES` rejects `acquire`+`force` and `renew`+`node`, and `holder=None` is accepted only for a non-holder release.
- Fenced release still requires the token; a superseded run's own exit-release is a no-op 409 and does not delete the new owner's row.
- Fail-open contract intact: `repo_claim`'s `hub-unavailable`, double-`missing`, `no-hub`, and `no-identity` paths are unchanged; the orphan-release retry still uses holder scope. The no-hub path is untouched.
- `--no-fleet-claim` does log (one `WARNING` on `brigade.fleet`; no `basicConfig`/`NullHandler` anywhere, so `lastResort` puts it on stderr) and is forwarded to the detached child. Not recorded in `run.json`, but neither is a successful claim, so that is consistent rather than a regression.
- The detach path returns before `run_lock`, so a detached child does not spuriously enter supersede mode. `RetainRunLockError` unwinds the claim before retaining the lock, so no claim is held without its lock.
- Tests cover both refusal directions (other node's claim, own claim with no dead lock), idempotent retry, `--force`, scope validation, and the `on_reconcile` negative cases.
**HOLD** - blocking on finding 1: tie the supersede to the claim it claims to be superseding (dead owner's lock token stored on the row, or at minimum an `acquired_at` ordering check), and add a test where the reconciled lock and the superseded claim target belong to different directories. Findings 2-6 are follow-ups. If you judge same-basename workspaces and cloned `node.toml` out of scope for this fleet, finding 1 downgrades to a follow-up and this merges - but say so in the issue rather than leaving the docstring guarantee unqualified.

solomonneas:
Addressed the HOLD in 94cbc4db.

**1 (blocking) — supersede now matches the exact claim row.** Every claim row records the acquiring run's `run.lock` lease (`lock_token`, `lock_acquired_at`, `lock_run_dir`; hub schema v3, never serialized; v2 rows survive via `ALTER TABLE` and can never be superseded). A `scope: "node"` acquire must present the reconciled dead lease as `supersede`; the UPSERT predicate is `owner_node = ? AND lock_token = ? AND (lock_acquired_at IS NULL OR lock_acquired_at <= ?)`. The secondary stamp guard compares the row's recorded lease stamp with the presented one — both are the same node's own ISO-8601 stamps from `owner.json`, so it never crosses the hub/node clock boundary (a hub-clock comparison would refuse legitimate supersedes under ordinary skew). `brigade run` reads its own lease via the new `runguard.read_lock_owner` and passes the reconciled dead owner payload; an unattributable lock (no token) supersedes nothing. The heartbeat re-acquire re-sends the lease and never supersedes.
Tests: hub exact-row match (different lease / older stamp / no lease / another node), `test_dead_lock_elsewhere_never_supersedes_a_live_run` (two same-basename workspaces sharing one `node_id` — the cloned-`node.toml` case is the same shape — ws-b's reconciled dead lock is refused and ws-a's holder still renews), and an end-to-end test that a real run's row carries its lease so its own simulated SIGKILL is superseded on the next run.

**2 — `--release` liveness.** `TARGET` may now be a workspace path (its claim key, node identity, and run lock are used; a bare key equal to the cwd's workspace key is guarded the same way). A non-`--force` release is refused while `runguard.inspect_run_lock_reconcile(workspace) == "live"`.

**3 — docstrings/CHANGELOG** now say: `scope` is an intent marker that keeps honest clients from stealing each other's claims, not an authorization; `node_id`/leases are caller-asserted under the one shared bearer token, which already authorizes `force`.

**4 — `--node NODE_ID`** override added; identity otherwise resolves from the target workspace, not cwd.

**5 (optional) — done:** the client now surfaces 400 bodies verbatim, and the CLI appends "(this fleet hub predates token-less release; upgrade it)" when the hub rejects `holder`/`scope`.

**6 (optional) — not done:** a lost-response retry of a supersede still logs nothing (the hub cannot distinguish it from an idempotent retry). Left as a follow-up.

Verification: ruff format/check clean; mypy clean on the 5 touched modules; `pytest tests/test_fleet_claims.py tests/test_fleet_sync.py tests/test_fleet_claim_release.py` → 112 passed; plus `test_runguard.py` + `test_run_detach.py` → 266 passed. `brigade roadmap commands --write`/`--check`: no diff.

solomonneas:
CI on bf030169 failed 6 tests in `tests/test_run_cli.py` (fixed-signature `run_lock` fakes rejecting the new `on_reconcile` kwarg; 8829 passed otherwise). Fixed in 57545beb: the fakes mirror the real signature and `runguard.read_lock_owner` tolerates a stubbed lock yield. Locally: `tests/test_run_cli.py` + `tests/test_fleet_claim_release.py` → 116 passed.

solomonneas:
Correctness probe #2 (Opus, read-only) on 57545beb. The supersede fencing from the first HOLD is verified correct (per-lease lock_token, never serialized). Verdict: HOLD on two NEW findings in the --release CLI path; 3 and 4 should land with them.

### What holds up
- **Supersede fencing is correct.** `fleet_hub.py:508-511` requires `claims.owner_node = excluded.owner_node AND claims.lock_token = <supersede.token>`. `lock_token` is a uuid4 minted per `_publish_lock` and never serialized (`_claim_payload` and the `GET /claims` query at `fleet_hub.py:608` both omit it), so a caller can only present a lease token it read off a local `owner.json`. A different node cannot steal an unexpired claim even by asserting `node_id`, because it cannot guess the token. A row with `lock_token IS NULL` fails the `=` comparison and is unsupersedable. Correct.
- **`lock_run_dir` is stored but never compared** (`fleet_hub.py:508-511`) — the "dead lock in one directory can never replace a claim taken by a run in another" guarantee rests entirely on token uniqueness. That is sufficient; the column is dead weight, not a bug.
- **Evidence for "dead"** is `_lock_is_stale` → `os.kill(pid, 0)` on both `pid` and `owner.json.pid`, plus `_invoke_before_terminalize`'s reconcileability gate, which raises rather than reconciling an activated-journal lock. Same-machine pid liveness is the right evidence for a same-node supersede; pid reuse fails safe (no reconcile). The container/shared-FS case with a cloned `node.toml` defeats it, but the docstring calls it out and the token match confines the blast radius to a row that identity itself created.
- **Fail-open is unchanged.** 400 now returns a payload instead of raising, but `_claim_op:731` still maps any non-200/409 to `reason="hub-unavailable"`, and `repo_claim` still falls back to the local lock. A v2 hub ignores unknown `scope`/`lock`/`supersede` keys (`_validate_claim_request` has no strict-key check), so a v3 client against a v2 hub degrades to a plain acquire — refused, never stolen.
- **`--no-fleet-claim` is not silent** (`run.py:538`, one `WARNING` on `brigade.fleet`) and is forwarded to the detached child (`run.py:951`). Heartbeat re-acquire never supersedes (`fleet_client.py:951`).
- **No-hub path unchanged**: `reason="no-hub"` short-circuits before any new code.
### Findings
**1. HIGH — `--release <key>` silently retargets when a same-named directory exists. `src/brigade/cli/fleet.py:143`**
`_release_workspace` treats any argument that is an existing directory as a workspace path, then calls `resolve_claim_target(workspace)`, which walks *ancestors* for `.brigade/node.toml` and returns the enclosing workspace's name.
Failure: cwd is `~/repos/brigade` (a workspace). `brigade fleet claims` lists a stuck claim on key `api` (owned by a crashed run in `~/work/api` on this node). Operator runs `brigade fleet claims --release api`. `Path("api").is_dir()` is true because `~/repos/brigade/api` is an ordinary package dir, so `workspace = ~/repos/brigade/api` and `resolve_claim_target` returns `"brigade"`. The command releases the claim on `brigade` — the operator's own repo, possibly held by a live run — and leaves `api` untouched. With `--force` it deletes it regardless of owner. The printed line names the real target, so it is detectable after the fact, but only after the wrong claim is gone. Gate the path branch on the argument looking like a path (contains a separator, or `.`/`~` prefix) or on the resolved key matching the argument.
**2. MEDIUM — the run-lock liveness gate is inert for the primary invocation. `src/brigade/cli/fleet.py:168`**
`if workspace is not None and not force and inspect_run_lock_reconcile(workspace) == "live"`. When `_release_workspace` returns `workspace = None` — a bare claim key that is neither a directory nor the cwd's own key, i.e. exactly the "a claim key as listed by this command" case the `--release` help advertises at `fleet.py:74-77` — the guard is skipped entirely and no error says so. `tests/test_fleet_claim_release.py:319` blesses this path (`--release repo-a` with no workspace, no liveness check).
Failure: node N has a live run on `~/work/api` holding claim `api`. Operator, in `/tmp`, runs `brigade fleet claims --release api` to clear what they believe is a stale row. Guard skipped, `owner_node` matches N, row deleted. The live run does not notice until its next heartbeat (`_claim_renew_interval(ttl)` = ttl/3, 300s at the 900s default). In that window another node acquires `api` and both runs work the repo. When the heartbeat does fire, `fleet_client.py:961` logs "lost ownership … continuing on the local run lock" and the run keeps going anyway. At minimum, refuse a non-`--force` release when the workspace could not be resolved, and say why.
**3. MEDIUM — `--node` bypasses the `--force` gate, and the JSON receipt misreports it. `src/brigade/cli/fleet.py:80-84`, `167`, `196`**
The stated contract is "owner_node must match unless `--force`". `--node NODE_B` makes it match by assertion: `release_claim(target, node_id=NODE_B, force=False)` sends `scope: "node"` with the victim's `node_id`, and the hub deletes it. `tests/test_fleet_claim_release.py:356` asserts exactly this (`released is True … forced is False`). Under one shared bearer token this is not a privilege escalation — `--force` is equally available — but it defeats the CLI-level guardrail without the `--force` prompt-equivalent, and the `--json` payload records `"forced": false` for what is a cross-node release. Either set `forced` true whenever `node_override` differs from the resolved local node, or require `--force` alongside `--node <other>`.
**4. LOW — token-less release reports a row it may not have deleted. `src/brigade/fleet_hub.py:584-588`**
The comment says "One statement, so a concurrent acquire cannot slip a new owner's row under a node-scoped delete." The DELETE is one statement and is correctly guarded, but `prior = _fetch_claim(...)` on line 584 is a separate read outside the write transaction, and the hub is a `ThreadingHTTPServer` with a per-request connection (`fleet_hub.py:815`). Under `--force`, a concurrent acquire between the SELECT and the DELETE makes the 200 response's `claim` describe the previous owner while a different row was deleted. Reporting-only; fold the read into the transaction or drop the claim in the comment.
### Verdict
**HOLD.**
Blocking:
- Finding 1 (`fleet.py:143`) — `--release` can delete a different claim than the one named.
- Finding 2 (`fleet.py:168`) — the advertised liveness refusal does not apply to the bare-key invocation it is advertised for.
Non-blocking but should land with them: finding 3 (`--force`/`--node` consistency and the `forced` field), finding 4 (comment or transaction fix).
The hub-side fencing, the schema v3 migration, `--no-fleet-claim`, and the fail-open/no-hub paths are sound as written.

solomonneas:
Addressed probe #2 in 0d1faafd.

**1 (blocking) — `--release <arg>` is always a claim key.** The directory branch is gone; `--path` opts into directory semantics, and the directory must be the workspace itself — `find_workspace_for_path` returning an ancestor is an error ("… is inside the workspace X; pass that workspace path instead"), never a retarget. Test `test_release_bare_key_is_never_retargeted_by_a_same_named_directory`: cwd is workspace `ws` holding a live claim, `ws/repo-a` is an ordinary subdir, `--release repo-a` (with and without `--force`) releases `repo-a` and `ws`'s holder still renews.

**2 (blocking) — liveness gate now covers the bare-key case.** New hub action `inspect` returns the current claim and, to the owner `node_id` only, the stored `lock_run_dir`. Without `--force` the CLI inspects, maps the run dir to a workspace on this machine (`run.json` `lock_workspace`/`cwd`, else the `<ws>/.brigade/runs/<id>` layout), and refuses while `runguard.inspect_run_lock_reconcile(ws) == "live"`; if the claim records no run dir or it isn't on this machine it refuses and says which ("pass --path <workspace> … or --force"). Test `test_release_bare_key_checks_liveness_through_the_recorded_run_dir` covers alive → refused, dead → released, no run dir → refused, foreign run dir → refused, `--force` → released with `forced: true`.

**3 — `--node` no longer bypasses `--force`.** A `--node` that differs from this machine's resolved identity is refused without `--force`; `forced` in the JSON receipt is now true for every release that was not a same-node verified one (the old test at :356 is replaced by `test_release_node_override_needs_force_when_not_this_node`).

**4 — hub receipt.** The prior-row read and the DELETE share a `BEGIN IMMEDIATE` transaction (same for acquire's prune/read/upsert), so `claim` in the receipt is the row the delete removed.

Verification: ruff/mypy clean; `tests/test_fleet_claims.py tests/test_fleet_sync.py tests/test_fleet_claim_release.py tests/test_fleet_dashboard.py` → 142 passed; `brigade roadmap commands --write` → no diff.

solomonneas:
CI on 0d1faafd: `test (3.11)` passed; `test (3.10)` failed one unrelated timing test (`tests/test_work_cmd_verification.py::test_work_verify_graphtrail_delta_preexisting_db_timeout_uses_stale_baseline`, `KeyError: 'stale_graph_used'` on a 0.05s `graphtrail_timeout`; 8857 passed otherwise) and fail-fast cancelled 3.12. The test passes 5/5 locally and touches nothing in this PR; re-running the failed jobs.

solomonneas:
Correctness probe #3 (Opus, read-only) on 0d1faafd. Prior HOLD items on the bare-key path are closed; the steal-path fencing re-verified sound. Verdict: HOLD on finding 1 (--release --path skips the deadness proof entirely and can delete a live same-node run's claim from any same-named directory) with finding 2 (token-less release not fenced to the inspected row) landing in the same change; 3 and 4 are follow-ups.

## Findings
**1. `--release --path` skips the deadness proof entirely (blocking)**
`src/brigade/cli/fleet.py:236-241` — the `inspect` probe only runs `if workspace is None`. In `--path` mode `workspace` is set at line 220, so the CLI never fetches the claim's recorded `lock_run_dir` and never checks that the claim it is about to delete belongs to the workspace passed. The only gate is line 255: *this* directory's `run.lock` isn't live. Target is derived from `workspace.name` (
Failure: node N has a live run in `~/work/api` holding claim `api` (lock lease recorded, `run.lock` live). Operator is blocked in `~/tmp/api` (or any dir named `api`, even an empty one) and runs `brigade fleet claims --release ~/tmp/api --path`. Lock verdict there is `absent` → passes → `release_claim(scope="node")` → hub deletes the row by `owner_node` alone (`fleet_hub.py:605`) → the liv
**2. Token-less release isn't fenced to the row that was inspected (medium)**
`src/brigade/fleet_hub.py:602-607` — the delete matches on `target` + `owner_node` only, ignoring expiry, holder token, and the lease columns the row now carries. The CLI's proof (`fleet.py:238-261`) is a separate round trip. Failure: operator inspects a stale claim on node N (dead run, verdict passes); a new `brigade run` on N acquires `api` in the interim; the release then deletes that live ru
**3. v2→v3 `ALTER TABLE` is not concurrency-safe (low)**
`src/brigade/fleet_hub.py:210-213` — `init_db` runs per request on a `ThreadingHTTPServer` with a fresh connection each time. Two concurrent claim POSTs against a freshly-upgraded v2 DB both see the column missing; the loser gets `duplicate column name` → `sqlite3.Error` → 500 → client maps it to `hub-unavailable` → that run fails open onto the local lock only. Self-heals after the first
**4. Contract drift in help/docs (low)**
- `--force` help ("or liveness cannot be verified") vs `fleet.py:255`, which refuses only on `"live"`. An `invalid` verdict (malformed/non-directory lock, `runguard.py:917`) sails through without `--force`.
- `docs/fleet-sync.md` and the CHANGELOG both say a non-forced release always proves deadness via `inspect` + recorded run dir; `--path` does not (finding 1). CHANGELOG's "`forced` is true for every release that was not a same-node verified one" is likewise false — `forced` is just the flag.
- `fleet.py:247`: when the run dir exists locally but doesn't map to a workspace (no `run.json`, non-standard layout), the error says "is not on this machine", which is wrong.
## Checked and clean
- Steal path evidence is sound: `on_reconcile` fires only from `_acquire_lock` after `_claim_stale_lock`, which requires every recorded pid dead (`runguard.py:172-185, 266-278`), and `run.py:544` filters to owners carrying `owner_token`. The hub's supersede branch (`fleet_hub.py:525-529`) requires `owner_node` match **and** exact `claims.lock_token` match, so a dead lock in another directory or on
- Renew is never widened (`CLAIM_ACTION_SCOPES["renew"]`), and the heartbeat re-acquire drops `supersede` (`fleet_client.py:964`).
- No-hub path unchanged; `--no-fleet-claim` logs a `warning` on `brigade.fleet` (same logger as `fleet_client._LOG`) and is forwarded to the detached child (`run.py:951`).
- Fail-open unchanged by the new 400 handling: `HTTPError` was already caught as `URLError` → `hub-unavailable`; an older hub ignores unknown `lock`/`scope` keys rather than 400ing.
- Param binding order in the upsert and the `_fetch_claim` index shifts (7/8/9/10) are correct; per-request connections mean the new `BEGIN IMMEDIATE` calls can't nest.
No tests run, per instruction.
**HOLD** — blocking: finding 1 (`--path` release deletes a live same-node run's claim with no proof and no `--force`). Finding 2 should land with it since the fix shares the same code path; 3 and 4 are follow-ups.

solomonneas:
Addressed probe #3 in f272c7f7.

**1 (blocking) — both `--release` modes now run the same proof.** `_release_claim` always inspects (unless `--force`), resolves the claim's *recorded* `lock_run_dir` to a workspace on this machine, requires that workspace to equal the `--path` directory when one is given ("the claim on 'api' was taken by a run in <ws-a>, not <ws-b>"), and takes `runguard.inspect_run_lock_reconcile` on that recorded workspace — `live` refuses, `invalid` (malformed lock) refuses, anything else (absent / dead-pid verdicts) proceeds. Test `test_release_path_proves_the_claim_belongs_to_that_workspace`: live run in `work/api`, operator in an empty `tmp/api` → refused with the live run still renewing; also refused when ws-a's run is dead (wrong directory); alive → refused; malformed lock → refused; dead → released with `forced: false`; `--force` from the wrong dir → released with `forced: true`.

**2 — release fenced to the inspected row.** The CLI sends the inspected claim's `acquired_at`; the hub's DELETE requires `target AND owner_node AND acquired_at` to still match (inside the existing `BEGIN IMMEDIATE`), else 409 with the current row ("was re-acquired since it was inspected"). Tests at hub, client, and CLI level (`test_release_refuses_a_claim_re_acquired_after_inspect` re-acquires between the CLI's inspect and release; the new run's claim survives).

**3 (follow-up, done) — migration race.** The claims-table check and `ALTER TABLE`s run under `BEGIN IMMEDIATE`, and a `duplicate column` `OperationalError` is treated as success. Test: 8 threads calling `init_db` on a v2 database with one lease column already present → no errors, columns present once, rows intact.

**4 (follow-up, done) — contract drift.** `invalid` now refuses (matches the `--force` help); help/docs/CHANGELOG say both modes prove deadness and that `forced` reflects the flag; the run-dir messages distinguish "does not exist on this machine" from "has no readable run.json … and is not under a <workspace>/.brigade/runs layout".

Verification: ruff/mypy clean; `tests/test_fleet_claims.py tests/test_fleet_sync.py tests/test_fleet_claim_release.py tests/test_fleet_dashboard.py` → 146 passed; `brigade roadmap commands --write` → no diff.

solomonneas:
Note: f272c7f7 produced no CI run because the PR had become CONFLICTING with main after #1147 (heartbeat orphan-release guard, same `_renew_loop` hunk as the lease re-acquire here) — GitHub skips `pull_request` runs it cannot build a merge ref for. Merged origin/main in 2ce0014f keeping both: `pending_orphan_release.set()` before the re-acquire, and the re-acquire sending the lease and never superseding. Fleet suites after the merge: 148 passed. (81d795fe is an empty commit from before I found the cause.)

solomonneas:
Correctness probe #4 (Opus, read-only) on 2ce0014f. Rounds 1-3 verified closed; supersede, renew fencing, --force gating, and fail-open all re-confirmed sound. Verdict: HOLD on one narrow item (finding 1: an inspect probe that finds no claim owned by this node still falls through to an unfenced token-less release with acquired_at=None, which the hub's NULL-tolerant WHERE accepts). Findings 2 and 3 are follow-ups.

## Findings
**1. `src/brigade/cli/fleet.py:248` — the token-less release is unfenced whenever the inspect probe doesn't find a claim owned by this node.**
`inspected_acquired_at` is only set inside `if probe.claim is not None and probe.claim.get("owner_node") == node_id:` (fleet.py:248-280). Every other outcome falls through to `release_claim(target, node_id=node_id, force=False, acquired_at=None)` at fleet.py:281, and the hub's token-less delete (`fleet_hub.py:625-628`) is then `DELETE ... WHERE target=? AND owner_node=? AND (NULL IS NULL ...)` —
Failure scenario: operator runs `brigade fleet claims --release api` on node A after a crash; the residual claim has already expired (or was already cleaned), so `inspect` returns `claim: null`. Between the probe and the release, `brigade run` starts in another workspace on node A and legitimately acquires `api`. The release deletes that live run's row. The run keeps executing; node B can now acqu
**2. `src/brigade/fleet_hub.py:204` (non-blocking) — `init_db` now takes a write lock on every request.**
The v2→v3 migration is wrapped in `BEGIN IMMEDIATE` unconditionally, and `init_db` runs per request (fleet_hub.py:777, 817, 859) including for `GET /claims` and every dashboard page. Steady state that serializes all hub requests behind a write lock with `busy_timeout=5000`; under concurrent dashboard + claim traffic a request can take SQLITE_BUSY and 500 (which for claims is fail-open, so it deg
**3. `src/brigade/cli/fleet.py:148-176` (non-blocking) — `_workspace_from_run_dir` duplicates `runguard.resolve_run_lock_workspace` (runguard.py:138-155)** with a slightly different precedence order. Two copies of "which workspace did this run lock" will drift; the CLI's extra needs (existence check, reason string) could wrap the existing helper.
## On the specific questions you asked

solomonneas:
Addressed probe #4 in 6a5dc942.

**1 (blocking)** — closed on both sides. CLI: when the inspect probe does not find a claim owned by this node, a non-force release refuses ("no claim owned by this node (<id>) on '<target>'; use --force to delete another owner's or an unverifiable row", or the held-by-node message when another owner holds it) and never sends a release request (`test_release_other_node_refused_without_force` now asserts zero `release_claim` calls). Hub: a token-less node-scoped release without `acquired_at` is a 400 ("claim field 'acquired_at' is required for a token-less node-scoped release"); only `scope: "force"` deletes unfenced — so even a bypassing client can no longer match-anything. Tests: probe-found-nothing refuses without `--force` (with `--force` it reports "no active claim"); the re-acquired race returns 409 and deletes nothing (the fresh run's claim renews afterward); unfenced release is 400 at hub and client level.

**2 (follow-up, done)** — `init_db` checks the claims columns first and takes `BEGIN IMMEDIATE` only when DDL is actually needed; the steady state never serializes behind the write lock.

**3 (follow-up, open)** — `_workspace_from_run_dir` vs `runguard.resolve_run_lock_workspace` dedup left for a follow-up (the CLI variant adds existence checks and refusal reasons; folding them into runguard touches its public surface).

Verification: ruff/mypy clean; `pytest tests/test_fleet_claim_release.py tests/test_fleet_claims.py` → **71 passed**.

solomonneas:
Same CI-skip cause recurred: main took #1156 (fleet export/Dolt sink) and the PR went CONFLICTING (docs/fleet-sync.md phases table), so 6a5dc942 got no pull_request run. Merged origin/main in 57908b5c (table keeps the phase-4 row from this PR and main's phase-5 row). Fleet suites on the merged tree incl. the new test_fleet_export.py: 178 passed.

solomonneas:
CI 3.10 failure on 57908b5c was my concurrency test hitting `database is locked` under runner load. Fixed in f89f2f27 by making `init_db` robust (test unchanged, still asserts zero errors): same-process callers serialize on a module-level `threading.Lock` keyed by db path with a needs-migration re-check under it, and the `BEGIN IMMEDIATE` DDL retries `database is locked` with bounded backoff, treating 'another connection finished the migration' and 'duplicate column name' as success; the steady state takes no lock. Local: the concurrency test 10/10 passed; `tests/test_fleet_claim_release.py tests/test_fleet_claims.py` → 71 passed.

solomonneas:
Correctness probe #5 (Opus, read-only) on f89f2f27. Verdict: MERGE (merged as b6fd4a07). All four threat questions clean; migration retry verified correct. Findings below filed as follow-ups.

## Findings
1. **`cli/fleet.py:324-330` — a claim with no recorded `run_dir` can only ever be released with `--force`, even in `--path` mode.** The non-force proof is unconditionally routed through `probe.lock_run_dir`. Two ordinary cases produce `NULL` there: runs started with `--no-artifacts` (`run.py:463`, `output_dir=None` → `_lock_owner_payload` sets `run_dir: None`), and every claim that predates th
2. **`fleet_hub.py:713-720` — the 409 refusal payload is read after `commit()`, not from the transaction.** `prior` is already captured inside the write transaction at line 703, but the failure branch re-reads. Failure scenario: a token-less release is fenced out by `acquired_at`; between `commit()` and line 714 a third party deletes or replaces the row, so the 409's `owner` describes a row that
3. **`cli/fleet.py:231-237` — the `no-identity` hint dead-ends.** It tells the operator to "pass `--node`", but `cli/fleet.py:283-289` rejects any `--node` that differs from `local_node`, and when identity resolution failed `local_node` is `"unknown"`, so every `--node` value differs. Failure scenario: on a machine with no `.brigade/node.toml`, `brigade fleet claims --release repo` prints "...or
4. **`run.py:539-543` — the skip is logged but not recorded in the run receipt.** `run.json` carries nothing distinguishing a run that held a hub claim from one that skipped arbitration. Failure scenario: two machines run the same repo concurrently because one used `--no-fleet-claim`; post-hoc the receipts are identical and the only trace is stderr, which for a detached run lands in `detached.lo
## Residual, non-blocking
- Node identity is a per-workspace (or per-home) uuid4, so a workspace shared across machines over a network filesystem collapses two machines into one `node_id` *and* one `run.lock`, at which point machine B's pid check calls machine A's live run dead and the supersede fires. The docs' "cloned node identity ... is never touched" is accurate for a copied `node.toml` with separate workspaces (diffe
- `handle_claim` acquire now takes a RESERVED lock on every acquire, not just contended ones. A `database is locked` after the 5s busy timeout becomes a 500 → `hub-unavailable` → fail-open to the local lock. Same exposure the single-statement upsert had, marginally wider window.
- Accepting `inspect_run_lock_reconcile == "skipped"` as "dead" is correct here (it implies no live pid; it only means `reconcile_run_lock` won't touch the *lock*), but it is worth a comment since the verdict name reads like an abstention.
Migration handling (`_migrate_claims_table`, `fleet_hub.py:220-262`) is correct: retry loop always terminates via `break` or `raise`, `ALTER TABLE` is idempotent against a concurrent winner, `ADD COLUMN` is skipped when the table was just created, and rollback is unconditional.
MERGE. None of the four threat questions produced a defect; items 1-4 are follow-ups, with #1 the one I would actually fix before the feature meets its first real crash.

[exited with code 0]
## Residual, non-blocking

- Node identity is a per-workspace (or per-home) uuid4, so a workspace shared across machines over a network filesystem collapses two machines into one `node_id` *and* one `run.lock`, at which point machine B's pid check calls machine A's live run dead and the supersede fires. The docs' "cloned node identity ... is never touched" is accurate for a copied `node.toml` with separate workspaces (different `lock_token`), just not for a shared mount — where the local lock model is already broken independently of this PR.
- `handle_claim` acquire now takes a RESERVED lock on every acquire, not just contended ones. A `database is locked` after the 5s busy timeout becomes a 500 → `hub-unavailable` → fail-open to the local lock. Same exposure the single-statement upsert had, marginally wider window.
- Accepting `inspect_run_lock_reconcile == "skipped"` as "dead" is correct here (it implies no live pid; it only means `reconcile_run_lock` won't touch the *lock*), but it is worth a comment since the verdict name reads like an abstention.

Migration handling (`_migrate_claims_table`, `fleet_hub.py:220-262`) is correct: retry loop always terminates via `break` or `raise`, `ALTER TABLE` is idempotent against a concurrent winner, `ADD COLUMN` is skipped when the table was just created, and rollback is unconditional.

MERGE. None of the four threat questions produced a defect; items 1-4 are follow-ups, with #1 the one I would actually fix before the feature meets its first real crash.

PR #1156, opened 2026-08-24
feat(fleet): deterministic export and optional Dolt sink (fleet sync phase 5, #1127)

### LINKED ISSUE #1127: fleet sync phase 5 (optional): Dolt sink for versioned fleet history/analytics
Epic: #1121. Optional, after Phases 2-4. Requires the hub.

**Goal:** a versioned, queryable long-term history of the whole fleet for analytics and audit, using Dolt ("Git for data") as a downstream sink, kept off the live hot path.

## Rationale
Dolt is overkill as the live coordination layer (its 3-way SQL merge solves conflicts that never happen when each machine only appends its own rows), but it is a good fit for the one thing it is uniquely good at: a **versioned SQL store you can branch, diff, and time-travel**. Reserve it for fleet history/analytics, not live status.

## Scope
- Periodically export the hub's events + outcomes into a Dolt database (a separate process/box), committed over time.
- Enables queries like "diff this week's outcome/verification rates vs last week across all machines," per-agent throughput history, and an auditable timeline, with `dolt diff` between commits.
- The live hub stays plain SQLite + HTTP; Dolt is a sink, never a dependency of the hot path.

## Acceptance
- A scheduled export lands hub data into Dolt with time-stamped commits.
- At least one useful historical/analytics query demonstrated (e.g. weekly outcome-rate diff across the fleet).
- Removing the Dolt sink does not affect live fleet operation.

solomonneas:
Fixes #1127. Epic #1121.

- `brigade fleet export --since <ts> --format jsonl|csv [--claims] [--db] [--out]`: deterministic stream of hub events or claims (stable column order, primary-key order, digest included; byte-identical on re-export).
- `[fleet.sink]` in `~/.brigade/fleet.toml`, `enabled = false` by default: `brigade fleet sink` runs `dolt table import -u` via subprocess when a `dolt` binary is on PATH; without dolt it is a documented no-op with a clear message. Never a Python dependency, never on the hot path.
- Docs: new section in `docs/fleet-sync.md` with an example weekly outcome-rate diff query; `docs/command-inventory.md` regenerated; CHANGELOG under Unreleased.

Verification: fleet suites (export, sync, claims, dashboard) 128 passed via `brigade work verify run`; ruff format/check and mypy clean; full `./scripts/verify` 8834 passed with one unrelated failure (`test_harness_user_scope.py::test_uninstall_state_mutation_failure_restores_every_profile_surface`, an update-notify cache file appearing during the test's home snapshot).

Ox (ox_deep) draft finished on the codex gpt-5.6-sol lane, landed by the conductor.


solomonneas:
Correctness probe (Opus, read-only) on 371bfb1e. Verdict: HOLD on findings 1 (sqlite3/OS errors escape as tracebacks on the two likely deployment shapes) and 2 (Dolt fleet_claims upsert never drops released claims). 3-6 folded into the same fix.

## Findings
**1. `fleet_export.py:245` — sqlite3/OS errors escape as a traceback, not an operator error.** `dispatch_export` (and `run_sink_pass`) catch only `FleetExportError`. `connect_hub_readonly` validates the open with `SELECT 1`, which never touches a table or the WAL/-shm, so every real failure surfaces later from `iter_event_records`/`iter_claim_records` as an uncaught `sqlite3.OperationalError`. Scenarios:
- `--db` pointed at any non-hub SQLite file, or a phase-2-era hub DB copy that a v2 hub never reopened (`init_db` is what creates `claims`): `no such table: claims` → stack trace. `fleet sink` always exports claims, so a scheduled pass tracebacks instead of exiting 1 cleanly.
- The hub DB is WAL (`fleet_hub.py:148`). A `mode=ro` connection still needs write access to the `-shm`; if the hub runs as a service user and the operator exports as themselves, the first query raises `SQLITE_READONLY_CANTINIT` → traceback, not the promised message. (`timeout=10` does set busy_timeout, so the busy path itself is fine.)
- `--out /missing/dir/x.jsonl` → uncaught `FileNotFoundError`.
Fix: wrap the query/write path and re-raise as `FleetExportError`.
**2. `fleet_export.py:338` — the Dolt `fleet_claims` table never drops released claims.** `table import -u` is upsert-only. Acquire `repo-a` → sink → release → sink: Dolt still shows `repo-a` owned with its old `expires_at`, indistinguishable from a live claim, and any query of "current claims" from the sink is wrong forever. Events are append-only so `-u` is right there; the claims snapshot needs `import -r` (replace) or a truncate-then-import, or the docs must say the claims table is an accumulating log, not a snapshot. `docs/fleet-sync.md` currently implies snapshot semantics.
**3. `fleet_export.py:244` — stdout path is locale/newline dependent; `--out` is not.** `--out` opens `newline="", encoding="utf-8"`; `sys.stdout` uses the ambient encoding and universal-newline translation. On Windows (the repo has a `windows-native-acceptance` job) `brigade fleet export ... > a.csv` writes CRLF while `--out a.csv` writes LF, so the byte-identical guarantee holds only per-sink, and `--format csv` with a non-ASCII `repo`/`seat` under a cp1252/latin-1 stdout raises `UnicodeEncodeError`. JSONL is safe by accident (`ensure_ascii=True`). `test_out_file_matches_stdout_stream:198` can't catch this because capsys bypasses the real stdout wrapper. Fix: `sys.stdout.reconfigure(newline="\n", encoding="utf-8", errors="strict")` (or write bytes via `sys.stdout.buffer`).
**4. `fleet_export.py:171-173` — `--since` silently drops events whose `ts` won't parse.** The hub only validates `ts` as a non-empty string (`fleet_hub.py:_validate_event`), so a node can persist a non-ISO `ts`. Those rows appear in a full export but in no `--since` export, so an incremental nightly archive loses them permanently with no warning and no count in the stderr summary. At minimum count and report them on stderr. Boundary itself is correct (`ts < since` skipped → `>=` inclusive) and the UTC/`Z`/naive handling is right.
**5. `fleet_export.py:197` — CSV values are not neutralized against formula injection.** `repo`, `seat`, `harness`, `run_id` are attacker-controlled strings from any node that can POST events; a value like `=HYPERLINK(...)` or `@SUM(...)` is written unquoted and executes if an operator opens the export in Excel/Sheets. Low severity for the Dolt path, real for the documented "archive/diff" path. Prefix-guard `= + - @ \t \r` or document CSV as machine-only.
**6. `fleet_export.py:334` — `dolt init` will likely fail on a fresh host with no configured identity.** Dolt refuses `init` without `user.name`/`user.email`, which turns the first scheduled sink pass into an exit-1 job failure, contradicting the "never a scheduled-job failure" design. Pass `init --name brigade --email brigade@localhost` (or check `dolt config` first). Flagged as plausible — I did not run `dolt` to confirm the version's behavior.
## Checks that came back clean
- Determinism: `ORDER BY` uses BINARY collation over the exact PK, columns are fixed tuples, JSONL is `sort_keys=True`/`ensure_ascii`, `_epoch_to_iso` is pure UTC, no `now()` anywhere in the row path. Re-export is byte-identical on a quiesced DB.
- Secrets: `holder_token` is never in `CLAIM_COLUMNS`; `[fleet] token`/`token_file` are never read by `load_sink_config`; dolt argv and error messages carry only paths and DDL.
- Subprocess: fixed argv list, no `shell=True`, no config value ever interpolated into a shell string; `dolt_binary` goes through `shutil.which`; timeout and `check=False` handled.
- Sink gating: `dispatch_sink` returns 0 before touching the DB when `enabled` is not literally `true` (`section.get("enabled") is True`, so `"yes"`/`1` stay off), and returns 0 with a clear message when `which` finds no binary.
- No behavior change for existing users: the diff touches only `fleet_export.py`, `cli/fleet.py` (two new subparsers), tests, and docs. `fleet_hub.py` — `/status`, `/claims`, `/events`, dashboard — is untouched.
**HOLD** — blocking: (1) unhandled `sqlite3.Error`/`OSError` producing tracebacks on the two most likely deployment shapes, and (2) `fleet_claims` upsert leaving released claims live in Dolt (fix the import mode or correct the docs). Items 3-6 can follow up, though 3 is cheap and CI covers Windows.

solomonneas:
Correctness probe #2 (Opus, read-only) on 2d636274. All six prior findings verified closed. Verdict: HOLD on F1 (the sink never runs dolt add/commit, so no version history exists); F2 near-blocking (the CSV spreadsheet guard mutates values on the way into Dolt); F3-F6 non-blocking.

### New findings
**F1 (blocking). `fleet_export.py:370-381` - the sink never commits, so it produces no version history.**
`dolt table import` writes to the working set only; `dolt add`/`dolt commit` is what creates a revision. After a week of nightly passes, `dolt log` still shows only the `dolt init` commit, `dolt diff HEAD` shows one ever-growing uncommitted delta, and `AS OF`/`dolt_history_*` queries return nothing. For `fleet_claims` this is worse than cosmetic: `-r` overwrites the working-set snapshot every pass
**F2. `fleet_export.py:152-156` + `:366-369` - the spreadsheet guard mutates data written into Dolt.**
`_csv_value` is applied by `_write_records` for every CSV caller, and the sink is a CSV caller. A field whose value starts with `=`, `+`, `-`, `@`, tab, or CR is stored in `fleet_events`/`fleet_claims` with a literal leading `'`. Concretely: a target or seat named `-worker` lands in Dolt as `'-worker`, and JSONL export of the same row disagrees with CSV export of it. The guard belongs at the human
**F3. `fleet_export.py:188-195` - `--since` filters on node-supplied `ts`, not `received_at`.**
Phase 2 spools events when the hub is unreachable, so an event created 2026-08-20 can be ingested 2026-08-23. An operator doing incremental archival (`--since` = last export time, one file per day) silently never exports it, and the "skipped N with unparseable ts" counter does not cover this case because the ts parses fine. `received_at` is already selected and exported; either offer `--since-rece
**F4. `fleet_export.py:219,223,252` - `brigade fleet export | head` exits 1 with error noise.**
`BrokenPipeError` is an `OSError`, so it becomes `error: could not write fleet export: [Errno 32] Broken pipe`, return code 1, plus the interpreter's "Exception ignored in: <_io.TextIOWrapper name='<stdout>'>" on shutdown. No `BrokenPipeError` handling exists in `cli/__init__.py`. For a command whose docstring says the stream "pipes cleanly", catch `BrokenPipeError` ahead of the generic `OSError`
**F5. `cli/fleet.py:70` - `from .. import fleet_export` runs at parser-build time.**
Every other dispatch in this module defers its heavy import into the dispatch function (`_dispatch_serve` at `:97`). This one imports `sqlite3`, `csv`, `subprocess`, `shutil`, and `tempfile` on every `brigade` invocation, including `brigade --help`. Move it into small `_dispatch_export`/`_dispatch_sink` wrappers.
**F6 (note, not a defect). `fleet_export.py:167` - read-only open of a WAL database needs a writable `-shm`.**
Same-host, same-user is the documented usage and works. But if the hub died leaving a hot WAL and no `-shm`, `mode=ro` cannot create the shared-memory file and the export fails with `unable to open database file`. The error is already operator-facing ("could not open fleet hub database"); consider naming the WAL case in the message.
No behavior change for users without `[fleet.sink]`: `dispatch_sink` returns 0 before touching the DB (`:393-398`), the reporting/claims/status/dashboard paths are untouched, and nothing in `fleet_export.py` is imported by `fleet_hub.py` or `fleet_client.py`.
**HOLD** - blocking: F1 (sink produces no Dolt commits, so "versioned history" is not delivered; either add the commit step or correct the docstring, `docs/fleet-sync.md`, and the phase table). F2 is a near-blocker if the sink ships as-is (data mutated on the way into Dolt). F3-F6 are non-blocking.

solomonneas:
Correctness probe #3 (Opus, read-only) on 5adedbf3. Rounds 1 and 2 verified closed. Verdict: HOLD on finding 1 (--claims exports expired claims as live, and the sink writes them into fleet_claims as current state); 2 and 3 land in the same push; 4 needs a real dolt check.

## New findings
**1. `--claims` exports expired claims as live — blocking.** `fleet_export.py:216` selects every row from `claims` with no expiry predicate and no `expired` column. The hub prunes expired rows only inside `acquire` (`fleet_hub.py:370`), and `list_claims` (`fleet_hub.py:430-442`) filters `expires_at <= now` and tags survivors with `expired`. Failure: node A crashes holding `repo-x` with a 900s TT
**2. Unconfigurable 120s timeout on an unbounded full-table import.** `fleet_export.py:56, 391, 433`. Every pass re-exports the whole `events` table (no watermark, no `--since` for the sink) into `dolt table import -u`, and `DOLT_TIMEOUT_SECONDS` is a module constant with no `[fleet.sink]` key. Failure: a hub that has accumulated ~1M events across the fleet takes longer than 120s in `dolt table im
**3. `--out` truncates the destination before the export can fail.** `fleet_export.py:305` opens `"w"`, which truncates immediately. Failure: `brigade fleet export --claims --out fleet-claims.jsonl` against a schema-v1 hub database (no `claims` table) leaves the existing archive file empty and exits 1 with "could not query fleet hub database". Write to a sibling temp file and `os.replace`.
**4. Doc SQL likely buckets everything into one NULL week (unverified).** `docs/fleet-sync.md:190` runs `DATE_FORMAT(ts, '%x-%v')` against `ts varchar(64)`, which holds `2026-08-23T12:00:00.123456Z` (`run_events.format_recorded_at`). go-mysql-server is stricter than MySQL about the trailing `Z`; the copy-paste example probably returns a single NULL `iso_week`. I did not run `dolt`, so this is unco
Not run: tests (per instruction) and `dolt` itself. `dolt status --porcelain` is assumed supported; I could not verify it against a binary.
No code changed, so no Brigade verify capture or handoff applies to this review.
**HOLD** — blocking: finding 1 (expired claims exported and written into `fleet_claims` as current state). Findings 2 and 3 are small and should land in the same push; 4 needs a check against a real `dolt`.

solomonneas:
Correctness probe #4 (Opus, read-only) on 9a774a59. All three prior rounds verified closed. Verdict: MERGE, no blocking items; findings 1-5 below are filed as a follow-up (1 first).

## New findings (all non-blocking)
**1. `expired` is written as Python `True`/`False` into a Dolt `boolean NOT NULL` column** — `fleet_export.py:98` vs `:253`/`:270`. `csv.writer` stringifies bools (confirmed: `csv.writer(...).writerow([False])` → `False`). Every sink pass with ≥1 live claim hands `dolt table import -r fleet_claims` the literal `False`. Dolt's import goes through GMS type conversion for `boolean` (tinyint); `False` is not a valid integer literal there, so the claims import may fail on every pass. `fake_dolt` is a shell stub that never type-checks, so no test covers this. Emit `0`/`1`. Secondary: with `include_expired=False` the column is constantly `False` in the sink — it carries no information and only adds import risk. Sink-only, so not a HOLD, but I would not advertise the sink until one pass runs against a real `dolt`.
**2. SQL NULL becomes `""` in CSV** — `fleet_export.py:270`. `repo`/`seat`/`harness` (nullable, `fleet_hub.py:100-102`) and `owner_conductor` land in Dolt as empty strings, so `WHERE repo IS NULL` never matches and the same row disagrees between the JSONL export (`null`) and the CSV/Dolt export (`''`). Any `GROUP BY repo` analytic over `fleet_events` silently splits unset from empty.
**3. Watermark only advances when Dolt reports a dirty working set** — `fleet_export.py:601-619`. If a pass dies after `table import -u` but before `_write_event_watermark` (e.g. the claims import fails, exit 1), the next pass re-imports the same rows, Dolt reports no change, `committed` is `False`, and the watermark stays stale — so the incremental window grows without bound until some claim happens to change. Idempotent, so no corruption, but it defeats the point of `events_incremental`.
**4. Sink drops events with unparseable `received_at` silently** — `fleet_export.py:625`. `dispatch_export` reports `stats.unparseable_timestamps` on stderr (`:418-420`); `run_sink_pass` collects the same stats into `event_stats` and never prints them. Once a watermark exists, such an event is skipped this pass and excluded from every later pass by the watermark. Practically unreachable today (hub-generated `isoformat()`), but it is permanent silent loss when it happens.
**5. `--out` inherits the tempfile's 0600 mode** — `fleet_export.py:357-379`. `NamedTemporaryFile` creates 0600, and `os.replace` carries that onto the destination, so re-exporting over an existing 0644 file tightens it. Tighter, not looser, so no leak — but a cron export another service reads will start failing after the first run under the new code. `os.chmod(temporary, 0o666 & ~umask)` before the replace.
No nondeterminism, no injection path (argv is a list, `cwd` is the only config-derived value, `shutil.which` resolves the binary), no secret leakage, and no `/status` / `/claims` / dashboard code path is touched.
MERGE — no blocking items. File 1-5 as follow-ups; #1 first, since it plausibly breaks the sink for its first adopter.

PR #1161, opened 2026-08-24
fix(fleet): per-node hub credentials bind node_id to the caller

### LINKED ISSUE #1150: fleet hub [HIGH]: bind node_id to the caller; a shared bearer lets any fleet member impersonate another node
Source: Daybreak Blue defensive pass 2026-08-24 over the merged fleet hub (#1131, #1140, #1145). Epic #1121. Deployment context: hub bound to a Tailscale IP on an LXC container, one shared bearer token for every fleet machine, plain HTTP on the tailnet.

## Finding
`src/brigade/fleet_hub.py:169`: the bearer authenticates the caller but never binds event or claim `node_id` to that caller, so any fleet member can post events or acquire/release/supersede claims as another node.

## Fix options (pick one, document the trust model either way)
- Per-node credentials: the hub keeps a small `nodes` table (node_id -> token hash); each machine gets its own token; the hub derives node_id from the token and rejects a body `node_id` that does not match.
- Or verified Tailscale identity: read the tailnet peer identity (e.g. via `tailscale whois` / the local API) for the source IP and map it to node_id server-side.
- Interim: keep the shared bearer but log a WARNING on the hub when body node_id changes for a source IP, and state in the docs that scope/owner fields are honest-client intent, not authorization (also noted in #1149 review).

## Acceptance
- A caller cannot post events or claim operations for a node_id it does not own.
- Existing single-token deployments have a documented migration path.

solomonneas:
## Summary

Implements #1150 with the per-node-credentials design: a node token *is* the node's identity on the fleet hub, and the single shared bearer becomes the admin token.

- **Hub** (`fleet_hub.py`): new `nodes` table (`node_id`, `token_sha256`, `label`, `created_at`, `revoked_at`; schema v4, added in place with `IF NOT EXISTS`). `POST /events` and `POST /claims` derive the caller's `node_id` from its token and answer **403** when a body `node_id` differs (whole batch rejected, nothing stored; every claim action incl. force release and inspect). A revoked token answers **401**. `GET /nodes` / `POST /nodes` (`add` / `revoke`) are admin-only; `add` returns the plaintext token once and nowhere else (409 while enrolled and unrevoked; re-add after revoke rotates). Node tokens may also read `/status` and `/claims` so `brigade fleet status` works on every machine.
- **Admin writes off by default**: the admin token may post under any `node_id` only with `brigade fleet serve --allow-admin-writes`; otherwise those posts are 403 with a message naming `brigade fleet nodes add` and the flag. Dashboard/cookie unchanged (admin token).
- **Client** (`fleet_client.py`): `[fleet] node_token_file` (`BRIGADE_FLEET_NODE_TOKEN` overrides). Without it, `load_fleet_config()` falls back to `token_file` and logs **one WARNING per process** naming the deprecated shared-token mode, so existing deployments keep working through the migration. 403 joins 401 as a spool-keeping status; `_post_claim_blocking` now surfaces the hub's 403 message in `ClaimDecision.detail`. New `load_fleet_settings()`, `add_node()`, `revoke_node()`, `fetch_nodes()`.
- **CLI**: `brigade fleet nodes add <node_id> [--label L] [--json]`, `nodes list [--json]`, `nodes revoke <node_id> [--json]`; `brigade fleet serve --allow-admin-writes`. Startup banner names the mode.
- Both token kinds compared with `hmac.compare_digest` (node lookup is by SHA-256 digest, then constant-time compare); tokens never logged or listed.
- **Docs**: trust-model section in `docs/fleet-sync.md` (node token = identity; admin token = control plane; scope/owner fields = honest-client intent), endpoint/config updates; runbook gains enrollment + rotation in §7, a *Migration from the shared token* section, 403 troubleshooting, a node-token smoke test, and the v4 rollback note. `docs/command-inventory.md` regenerated. CHANGELOG under `[Unreleased] › Security`.

Existing hub fixtures in the fleet suites now pass `allow_admin_writes=True` explicitly (they exercise the pre-#1150 shared-token semantics); the new `tests/test_fleet_nodes.py` covers the default mode.

## Tests

`tests/test_fleet_nodes.py` (new): impersonation rejected (403, events and every claim action), matching node accepted, revoked token rejected (401) and rotation, unknown/malformed bearer 401, `/nodes` admin-only, add conflict + field validation, token stored only as SHA-256, constant-time compare for both kinds, v3→v4 in place, binding enforced at the function level, client node-token path (no warning), client shared-token fallback with exactly one WARNING, 403 keeps the spool, env precedence, CLI add/list/revoke round trip and error reporting, `serve --allow-admin-writes` plumbing.

```
PYTHONPATH=src .venv/bin/python -m pytest tests/test_fleet_sync.py tests/test_fleet_claims.py tests/test_fleet_dashboard.py tests/test_fleet_nodes.py -q
137 passed in 56.81s

+ tests/test_fleet_claim_release.py tests/test_fleet_export.py tests/test_fleet_action_dispatch_cmd.py
206 passed in 86.50s
```

`ruff format` / `ruff check` clean on touched files; `mypy src/brigade/fleet_hub.py src/brigade/fleet_client.py src/brigade/cli/fleet.py`: no issues.

## Notes for review

- `force` release remains available to any enrolled node (attributed to its own `node_id`, which is the change); restricting it to the admin token was outside the chosen design and is documented as honest-client intent.
- Hub rollback across the v3→v4 boundary is refused by the existing `user_version` guard; the runbook says to restore the pre-upgrade backup or stay on the new ref with `--allow-admin-writes`.

Fixes #1150


solomonneas:
Security/correctness probe (read-only) on a05630bf. Verdict: HOLD on one rollout-blocking design gap (per-workspace node_id vs one-token-per-machine enrollment); the auth core is verified sound (no bypass, no impersonation path, no token leak). Decision below.

## Review: PR #1161 (fleet hub per-node credentials, #1150)
**BLUF: HOLD on one rollout-blocking gap. The auth core itself is correct and I found no bypass, impersonation path, or token leak.**
### Blocking
**1. `node_id` is per-workspace, not per-machine, so one node token per machine silently breaks every other workspace on that machine.**
`docs/runbooks/fleet-hub-hogwarts.md:250` ("Every machine needs its own node token") plus the single `brigade fleet nodes add <NODE_ID>` in section 7 assume one identity per host. The code does not work that way: `aboyeur.make_run_dir` calls `node.ensure_identity(workspace)` (`src/brigade/node.py:146`), which mints a fresh uuid4 into `<workspace>/.brigade/node.toml` on the first run in each workspace, and `fleet_client.resolve_node_id` (`src/brigade/fleet_client.py:243`) resolves the *nearest ancestor* workspace file. `cli/run.py:549` passes `base_path=run_cwd`, so claims use that same per-workspace id. `docs/fleet-sync.md:143` already states each workspace "resolves its own workspace identity".
Failure scenario: host has `~/repos/foo` and `~/repos/bar`, each with its own `.brigade/node.toml`. Operator runs `brigade node` in `foo`, enrolls that id, installs the token. Runs in `bar` now POST events under bar's id, hub answers 403, and since 403 is in `_RETRYABLE_4XX` (`fleet_client.py:103`) those events spool forever and start dropping oldest at the 16 MiB cap (`_spool_append_locked`, warning only). Every `repo_claim` in `bar` gets 403, retries once, logs "fleet hub unreachable", and falls open to the local lock, so cross-machine arbitration for that repo is gone. Both failures are quiet: the hub sees a healthy fleet, the machine sees warnings in a log nobody reads.
Fix is a decision, not a patch: either enroll every workspace identity (runbook + `nodes add` guidance), let one token cover several `node_id`s, or make node identity per-machine. Whichever, the runbook and the migration section need to match the code before this ships.
### Non-blocking
**2. Unauthenticated `POST` now forces a DB write before rejection.** `fleet_hub.py:1147` opens `init_db` before `_caller` authenticates; `init_db` ends with `PRAGMA user_version=4` + `commit()`, a write transaction. Before this PR `_authorized()` gated the route first. Any peer that reaches the port and sends `Authorization: Bearer x` can drive WAL write-lock churn against real nodes. Bounded by the Tailscale-only bind, so low, but the inline comment presents the reordering as pure hardening and it is a tradeoff.
### Checked and clean
- Route matrix: `/nodes` GET+POST admin-only (403 for node tokens, body not read first); `/events` and `/claims` reject the admin token with 403 unless `--allow-admin-writes`; dashboard and cookie enrollment still require the exact admin token, not a node token. No route accepts the wrong token kind.
- Impersonation: `store_events` (`fleet_hub.py:390`) rejects the whole batch before any insert; `handle_claim` (`fleet_hub.py:647`) checks before any read or write, covering acquire/renew/release/inspect including `scope=node`/`force`. `force` by an enrolled node is still possible and is correctly documented as such in both the module docstring and `docs/fleet-sync.md`.
- Hashing: `secrets.token_urlsafe(32)` (256 bits) with unsalted SHA-256 is the right call for a high-entropy random token; the indexed lookup is by digest, then `hmac.compare_digest`, and the admin token uses `compare_digest` on the whole header. No timing or rainbow-table concern.
- Revocation takes effect on the next request (`init_db` and `lookup_node_token` run per request); the digest is retained so the old token gets 401 "revoked" rather than falling through to unknown. Residual claims held by a revoked node expire at TTL since renew now 401s.
- `add_node`'s single-statement upsert with `WHERE nodes.revoked_at IS NOT NULL` plus the `rowcount` check makes concurrent adds safe and re-add the only rotation path.
- Token leakage: hub `log_message` is a no-op, `token_sha256` is excluded from `_NODE_COLUMNS`, the 409 and 400 paths return no token, error strings carry `node_id` only, and the client's `FleetClientError` messages relay the hub's message without the credential.
- Client migration: `load_fleet_settings` keeps the two token kinds separate, `load_fleet_config` falls back to the admin token with exactly one process-level WARNING, `_admin_request` refuses to substitute a node token, and 401/403 on claims still yields `hub-unavailable` so the fail-open contract holds.
- Schema v4 is one additive `CREATE TABLE IF NOT EXISTS`; the newer-version refusal and the rollback consequence are documented accurately.
- Docs match the code on every trust-model claim I checked, including the honest-client framing of `scope`, `lock`, and `supersede`.
Tests not run per your instruction; `tests/test_fleet_nodes.py` reads as genuine coverage of the above (impersonation, revocation, 409 on re-add, hash-only storage, v3 to v4 in place, the shared-token fallback both ways) and resets `_SHARED_TOKEN_WARNED` in its fixtures. No code changes made, so no Brigade verify capture applies.
**HOLD** — blocking item: finding 1 (per-workspace node identity versus the one-token-per-machine enrollment model and runbook).

PR #1181, opened 2026-08-25
fix(fleet): claim-loss abort, orphan-release leak windows, proxy-free opener and private spool

### LINKED ISSUE #1152: fleet claims [MEDIUM]: a run continues after the hub says another owner holds the claim; abort or fence protected work on lost ownership
Source: Daybreak Blue defensive pass 2026-08-24 over the merged fleet hub (#1131, #1140, #1145). Epic #1121. Deployment context: hub bound to a Tailscale IP on an LXC container, one shared bearer token for every fleet machine, plain HTTP on the tailnet.

## Finding
`src/brigade/fleet_client.py:797`: claim acquisition and heartbeat loss fail open. When a renew returns 409 held-by-another, `_renew_loop` logs "lost ownership" and the run keeps going unarbitrated.

## Fix
- Make the reaction to lost ownership configurable: default `abort` (cancel the run's dispatch phase and mark the run interrupted with reason `claim-lost`), with `continue` as the documented opt-out for solo machines.
- Or enforce the fencing token at the mutation sink (the journal / PR push step refuses to proceed if the claim holder token is no longer the hub's current holder).

## Acceptance
- Two machines cannot both keep working the same repo after a supersede or expiry; the loser stops within one heartbeat interval and says why.

### LINKED ISSUE #1154: fleet client [MEDIUM]: proxy-disabled opener for tailnet traffic; spool directory and file hardening (0700/0600, no-follow opens, exclusive temp files)
Source: Daybreak Blue defensive pass 2026-08-24 over the merged fleet hub (#1131, #1140, #1145). Epic #1121. Deployment context: hub bound to a Tailscale IP on an LXC container, one shared bearer token for every fleet machine, plain HTTP on the tailnet.

## Findings
- `src/brigade/fleet_client.py:216`: the default `urllib` opener honors `HTTP_PROXY`/`HTTPS_PROXY`, so the plain-HTTP bearer can be sent to a configured proxy before it reaches the tailnet.
- `src/brigade/fleet_client.py:284`: spool directories and files inherit umask; append, temp, and lock opens follow symlinks, exposing event metadata or enabling redirected writes on permissive home directories.

## Fix
- Build the opener with `ProxyHandler({})` for hub traffic (and document NO_PROXY as the interim control); prefer HTTPS via Tailscale Serve in the runbook.
- Create the spool dir 0700 and files 0600; open with `O_NOFOLLOW` where available and refuse non-regular files; exclusive, unpredictable temp names for the atomic rewrite; reuse the repo's existing private-mode helpers and the Windows-portable test helper from #1133.

## Acceptance
- A proxy env var never receives hub traffic; a symlinked spool path is refused with a logged error; mode tests pass on POSIX and skip cleanly on Windows.

### LINKED ISSUE #1157: fleet claims: close the straggler-acquire leak window at exit and the initial-acquire orphan thread first-sleep
Follow-ups from the #1147 (issue #1142) final review. Epic #1121.

1. `fleet_client.repo_claim` exit path: the synchronous release retry is gated on the sync release returning `hub-unavailable`. If the abandoned re-acquire is still in flight when the exit release lands, the release gets a definitive `released:false` (reason `missing`), the retry loop is skipped, and the straggler POST then commits a row that lives the full TTL. Fix: also retry once when the reason is `missing` and `pending_orphan_release` is set, or have the hub reject an acquire whose holder token already released (tombstone the holder for one TTL).
2. `_schedule_orphan_release` (initial-acquire path) still sleeps `ORPHAN_RELEASE_RETRY_SECONDS` before its first attempt on a daemon thread, so a `brigade run` that finishes within 30s of a hub-unavailable acquire never releases. Attempt immediately, then back off.
3. Polish: log a WARNING when exit retries exhaust (teardown can now take ~15s silently, including on Ctrl-C); drop the never-cleared `pending_orphan_release` flag if it buys nothing; move the docs/fleet-sync.md "Phase 4 claims" section below Phase 3 and describe the exit cleanup behavior.

solomonneas:
Fixes #1152. Fixes #1154. Fixes #1157.

All changes are in `src/brigade/fleet_client.py`, its tests, `docs/fleet-sync.md`, and the CHANGELOG. `fleet_hub.py`, `fleet_dashboard.py`, and `cli/fleet.py` are untouched so this does not collide with the in-flight dashboard work.

## #1152: fail closed on lost ownership

When the heartbeat renew is answered 409 held-by-another, `repo_claim` no longer logs and lets the run continue unarbitrated. Default: one WARNING, then abort the guarded run through the same main-thread interrupt path as Ctrl-C. `on_claim_lost` callback for a different reaction; `BRIGADE_FLEET_CLAIM_LOSS=continue` / `claim_loss_policy="continue"` is the documented opt-out for solo machines.

## #1154: proxy-free opener and private spool

The hub opener is built with an empty `ProxyHandler`, so `HTTP_PROXY`/`HTTPS_PROXY` can neither intercept the bearer token en route to a tailnet hub nor blackhole requests. Spool directory 0700; spool, lock, and temp files 0600 regardless of umask; opens are no-follow (`O_NOFOLLOW` where available) and refuse non-regular files; atomic rewrites use exclusive unpredictable temp names. Mode assertions are POSIX-only in tests.

## #1157: orphan-release leak windows

The background release scheduled after a lost acquire response attempts immediately before backing off, so a short `brigade run` still frees the row instead of leaving it for the TTL. At exit, the inline release retry also runs after a definitive `missing` answer, not only on `hub-unavailable`, because the straggler POST can land right after the exit release was told the claim was gone. Exhausted exit retries log a WARNING naming the residual TTL exposure.

## Verification

Conductor, through Brigade on this tree: `ruff check` on `fleet_client.py` and the six fleet test files (`+1`); `pytest -q` over `tests/test_fleet_claim_release.py tests/test_fleet_claims.py tests/test_fleet_client_hardening.py tests/test_fleet_nodes.py tests/test_fleet_sync.py tests/test_node.py`: 193 passed, 1 failed, 1 deselected.

The 1 failed and 1 deselected are `test_fleet_claims.py::TestClientClaims::test_resolve_claim_target_uses_workspace_name` and `test_fleet_claim_release.py::TestReleaseCli::test_release_path_proves_the_claim_belongs_to_that_workspace`. Both fail identically on clean `origin/main` (b72cfc77): 2 failed, 80 passed on the same two files. They are environment-bound (pytest tmp under `$HOME` plus a home `~/.brigade/node.toml` makes the resolver return the home directory name), tracked separately. CI has no home node identity and runs the full `./scripts/verify` gate.

## Provenance

Implemented by the `ox_worker` seat (`opencode/x-preview-f-free`) under `brigade run` `19faf9b8.20260825-205724-6d530229`; the worker committed and pushed, the conductor verified and opened this PR because the worker sandbox has no `gh` auth.


solomonneas:
Read-only security review by the `daybreak` seat (codex `gpt-daybreak-blue-latest`, brigade run `19faf9b8.20260825-221052-fabaf7e7`, native read-only sandbox). Relayed by the conductor; findings are the reviewer's, not yet independently reproduced. Converting to draft until 1-4 are addressed.

BLUF: PR #1181 is not ready to merge. Likely, confidence High - static tracing found six gaps, including claim-loss suppression and orphan-release races.

## Findings

1. **High - `on_claim_lost` suppresses fail-closed abort**
   - **Where:** [fleet_client.py:1295](src/brigade/fleet_client.py:1295)
   - **Scenario:** A no-op or failing callback prevents `interrupt_main()`. The heartbeat exits while the guarded work continues beside the new owner.
   - **Fix:** Always perform the default interrupt after notification unless `claim_loss_policy="continue"` was explicitly selected. Callback failure must fall back to interrupting.

2. **Medium - Immediate orphan release can race ahead of the lost acquire**
   - **Where:** [fleet_client.py:1076](src/brigade/fleet_client.py:1076)
   - **Scenario:** A timed-out acquire remains in flight. The immediate release receives `missing` and exits, then the acquire commits, leaving the claim until TTL expiry.
   - **Fix:** Treat `missing` as retryable during the uncertainty window, or track and join the abandoned request before accepting it as definitive.

3. **Medium - Exit retries can all finish before a late re-acquire commits**
   - **Where:** [fleet_client.py:1315](src/brigade/fleet_client.py:1315)
   - **Scenario:** Three back-to-back releases receive `missing`; the abandoned re-acquire commits immediately afterward.
   - **Fix:** Synchronize with the request worker or space retries across at least the outstanding request timeout.

4. **Medium - Proxy-free opener still forwards bearer tokens across redirects**
   - **Where:** [fleet_client.py:282](src/brigade/fleet_client.py:282)
   - **Scenario:** A hub endpoint returns a cross-origin 302. Python’s default redirect handler copies `Authorization` to the redirected request.
   - **Fix:** Disable redirects or allow only same-origin redirects while stripping credentials otherwise.

5. **Low - Spool directory traversal and mode enforcement remain incomplete**
   - **Where:** [fleet_client.py:359](src/brigade/fleet_client.py:359), [fleet_client.py:394](src/brigade/fleet_client.py:394)
   - **Scenario:** A symlinked `fleet-spool` directory is followed. `chmod` failure is swallowed, and existing permissive spool or lock files are not changed to 0600.
   - **Fix:** Anchor operations to a verified `O_DIRECTORY|O_NOFOLLOW` directory descriptor, use `dir_fd`, and `fchmod` existing files.

6. **Low - Claim-loss abort is recorded as user cancellation**
   - **Where:** [run.py:638](src/brigade/cli/run.py:638), [run.py:762](src/brigade/cli/run.py:762)
   - **Scenario:** The abort works, but receipts and stderr report “canceled by user,” losing the actual claim-loss cause.
   - **Fix:** Record a claim-loss flag and dedicated failure kind before interrupting.

## Clean areas

- Exact `BRIGADE_FLEET_CLAIM_LOSS=continue` handling is deliberate; explicit `"abort"` wins.
- Every fleet HTTP request routes through `_hub_open`; none uses proxy-aware `urlopen`.
- Leaf spool opens use `O_NOFOLLOW` and reject non-regular files. Temp names are exclusive and unpredictable.
- Releases are fenced by target, node, and holder token at [fleet_hub.py:793](src/brigade/fleet_hub.py:793). Repeated releases cannot delete another owner’s claim.
- The CLI propagates the default `KeyboardInterrupt` and cancels active work.

No tests were run because the task was explicitly read-only. `git diff --check origin/main...HEAD` and `git status --short` produced no output; no files were changed.

Next: add delayed-commit race tests before revising the cleanup synchronization.

solomonneas:
Sendback outcome (ox_worker, brigade run `19faf9b8.20260825-221632-9b6df536`, verify-then-fix): all six findings reproduced, fixed in 9a2c0747.

1. High, callback suppression: notify is best-effort, then `_thread.interrupt_main()` always runs unless `claim_loss_policy="continue"`; a raising callback still interrupts. Tests: `test_raising_claim_lost_callback_still_interrupts`, `test_lost_ownership_invokes_the_abort_callback_once` now expects KeyboardInterrupt.
2. Medium, orphan release vs in-flight acquire: `ORPHAN_RELEASE_UNCERTAINTY_SECONDS = 2 * CLAIM_TIMEOUT_SECONDS + 1`; `missing` inside the window is retried. Race test `test_orphan_release_retries_missing_within_uncertainty_window`.
3. Medium, exit retries: each retry sleeps one `CLAIM_TIMEOUT_SECONDS` first. Race test `test_exit_retry_is_spaced_across_a_late_reacquire_commit`.
4. Medium, redirects: `_HubRedirectHandler` follows only same scheme+netloc; cross-origin 3xx is a transport failure.
5. Low, spool dir: symlinked spool directory refused, chmod failure logged, `fchmod(0o600)` on every open of an existing file.
6. Low, claim-loss cause: `abort_on_fleet_claim_lost` records `failure_kind=fleet-claim-lost`, rc=2, distinct message. Test `test_claim_loss_midrun_is_not_recorded_as_user_cancel`.

Docs: CHANGELOG entries amended, `docs/fleet-sync.md` updated. Conductor verification through Brigade: ruff check and format (`+1`, `+1`); `pytest -q` over the fleet claim/client/sync tests plus `test_read_only_enforcement.py` and `test_outcome_infrastructure_attribution.py`: 236 passed, 2 deselected (#1182) (`+1`). Re-dispatching the daybreak review against 9a2c0747; stays draft until it clears.

solomonneas:
Daybreak re-review against 9a2c0747 (brigade run `19faf9b8.20260825-223146-4ec7affb`). Relayed by the conductor.

Conductor disposition: round 2 dispatched for the client-side items (1, the default-port netloc nit in 4, 5, 6, and the new non-main-thread finding). Findings 2 and 3 are timing-bounded on the client by design; the protocol fencing the reviewer recommends is hub-side and out of this PR's scope (this branch deliberately does not touch `fleet_hub.py`), and the residual is a TTL-bounded stale claim rather than an integrity break, since hub release is already fenced by holder token. Filed as a follow-up for the hub bundle.

BLUF: NOT READY. Almost certainly, confidence High - three residual failure paths reproduced in-memory.

### Prior findings

1. **[HIGH] Not closed: callback can still suppress abort.** [fleet_client.py:1341](src/brigade/fleet_client.py:1341) calls the callback before `interrupt_main` and catches only `Exception`. A blocking callback or `SystemExit` prevents interruption. Probe: `callback_baseexception_interrupts 0`. Fix: trigger cancellation independently of callback execution.

2. **[MEDIUM] Partially closed: immediate orphan cleanup.** [fleet_client.py:1122](src/brigade/fleet_client.py:1122) retries `missing` for six seconds, but the abandoned worker at [fleet_client.py:343](src/brigade/fleet_client.py:343) can outlive that window, particularly during DNS resolution. Fix: server-side tombstone/generation fencing or cancellable request completion.

3. **[MEDIUM] Partially closed: exit retry race.** [fleet_client.py:1374](src/brigade/fleet_client.py:1374) performs only two finite retries. Probe committed after the final retry: `late_commit_orphan True release_calls 3`. The test at [test_fleet_claims.py:1190](tests/test_fleet_claims.py:1190) commits before the second retry, so it misses this case. Fix: use protocol fencing rather than timing.

4. **Closed: cross-origin Authorization replay.** [fleet_client.py:291](src/brigade/fleet_client.py:291) permits same-origin redirects and rejects cross-origin redirects. Minor compatibility issue: raw `netloc` comparison rejects equivalent default-port redirects such as `http://hub` to `http://hub:80`.

5. **[LOW] Partially closed: spool directory hardening.** File descriptors are regular-file checked before `fchmod`, so no unexpected-fd issue was found. However, [fleet_client.py:383](src/brigade/fleet_client.py:383) has a symlink-check/chmod TOCTOU, and chmod failure merely warns before continuing. Fix: open and validate the directory via no-follow descriptors, and refuse writes when privacy cannot be enforced.

6. **[LOW] Partially closed: receipt classification.** [run.py:501](src/brigade/cli/run.py:501) records claim loss from the heartbeat thread but swallows failures. If that write fails, [run.py:665](src/brigade/cli/run.py:665) records a user keyboard interrupt. The regression test checks stderr, not `run.json`. Fix: classify and record from the main-thread handler using `fleet_claim_loss`.

### Additional regression

**[MEDIUM] Non-main-thread use does not abort the guarded block.** `_thread.interrupt_main()` targets the process main thread regardless of which thread entered `repo_claim`. Probe: `interrupt_called 1 guarded_worker_continued True`. Restrict abort-policy use to the main thread or add owner-thread cooperative cancellation.

No deadlock was found. Exit cleanup is bounded, but a blackholed hub can add roughly 20 seconds. Read-only checks passed:

- Ruff: `All checks passed!`
- `git diff --check`: exit 0
- `git status --short`: clean

Full pytest and Brigade receipt capture were not run because they write files, contrary to the report-only sandbox.

NOT READY

solomonneas:
Round-2 outcome (ox_deep, brigade run `19faf9b8.20260825-224008-c783ef31`): all five client-side items reproduced with failing tests first, fixed in f4bbadcd.

1. High: `_abort_claim_owner` sets `cancel_event`, runs the callback under `except BaseException`, and a watchdog daemon fires the interrupt after `CLAIM_LOST_CALLBACK_GRACE_SECONDS` if the callback blocks. Tests: `test_systemexit_claim_lost_callback_still_interrupts`, `test_blocking_claim_lost_callback_does_not_delay_the_abort`.
2. Medium (new): `repo_claim` captures the owner thread; non-main owners get cooperative cancel via `ClaimDecision.cancel_event` plus a WARNING naming the requirement, never a wrong-thread `interrupt_main`. Test: `test_claim_loss_off_main_thread_aborts_that_thread_cooperatively`.
3. Low: `_ensure_private_dir` opens `O_DIRECTORY|O_NOFOLLOW|O_CLOEXEC`, `fchmod` 0700, re-verifies via `fstat`, refuses writes when unenforceable; lock/spool files open with `dir_fd` against the validated descriptor. Tests: `test_unenforceable_dir_privacy_refuses_the_write`, `test_unverifiable_dir_mode_refuses_the_write`.
4. Low: heartbeat callbacks only print and append markers; classification moved to the main thread (`_fleet_interrupt_failure`), receipts assert `run.json` `failure.kind == "fleet-claim-lost"`. Test: `test_claim_loss_kind_survives_heartbeat_receipt_failure`.
5. Nit: `_origin()` / `_same_origin()` normalize scheme-default ports.

Original findings 2 and 3 keep the timing approach by disposition; `test_exit_logs_when_orphan_cleanup_retries_are_exhausted` now pins the post-final-retry late commit as the documented TTL-bounded residual (hub-side fencing is #1189).

Conductor verification through Brigade: ruff check and format (`+1`, `+1`); `pytest -q` over the fleet claim/client/release/sync tests plus `test_read_only_enforcement.py` and `test_outcome_infrastructure_attribution.py`: 248 passed, 2 deselected (#1182) (`+1`). Re-dispatching daybreak against f4bbadcd.

solomonneas:
Daybreak round-2 re-review against f4bbadcd (brigade run `19faf9b8.20260826-001902-0446da2b`). Relayed by the conductor. All five round-2 items confirmed closed; four narrow residuals below. Round 3 dispatched (last Ox round for this PR).

BLUF: NOT READY. Very likely, confidence High - three security gaps remain in fail-closed handling and spool anchoring.

Findings:

- **Medium:** [`run.py:557`](src/brigade/cli/run.py:557), [`fleet_client.py:1502`](src/brigade/fleet_client.py:1502). If writing the credential-error message fails, execution never reaches the marker or `interrupt_main()`. The callback exception is swallowed, allowing the run to continue after credential rejection. Move the marker before diagnostics and invoke the interrupt from `finally`.

- **Medium:** [`run.py:542`](src/brigade/cli/run.py:542). Claim-loss classification is appended after a fallible `print()`. A closed stderr still triggers the independent abort, but records it as user cancellation with exit 130. Append the marker first and test stderr failure.

- **Medium:** [`fleet_client.py:580`](src/brigade/fleet_client.py:580), [`fleet_client.py:587`](src/brigade/fleet_client.py:587). Atomic rewrites ignore the validated `dir_fd`; both temporary creation and replacement resolve `path.parent` again. A concurrent directory swap can redirect the rewrite. Create and replace using names relative to `dir_fd`.

- **Low:** [`fleet_client.py:1283`](src/brigade/fleet_client.py:1283). The watchdog and callback-finally paths use a non-atomic check/set on `fired`. At the grace boundary both can interrupt. Guard the one-shot interrupt with a lock and test the boundary race.

Clean:

- Callback `BaseException` no longer suppresses claim-loss abort.
- Off-main ownership uses cooperative cancellation without interrupting the unrelated main thread.
- Directory mode refusal and `dir_fd` cleanup are correct.
- Ordinary callback completion prevents a late watchdog interrupt.
- Duplicate receipt classification is harmless because finished receipts are ignored.
- Default-port origin normalization is correct.
- The accepted orphan-release residual remains TTL-bounded.

`git status --short` produced no output; no files changed.

NOT READY

solomonneas:
Round-3 outcome (ox_deep, brigade run `19faf9b8.20260826-003940-892610ca`): all four residuals reproduced with failing tests first, fixed in bd76e603 (rebased onto current main).

1. Credential-refusal path: marker appended before any print, `interrupt_main` in `finally`. Test: `test_dead_stderr_credential_refusal_still_aborts_as_fleet_credentials_rejected`.
2. Claim-loss classification: marker first, then print. Test: `test_dead_stderr_claim_loss_is_recorded_fleet_claim_lost_not_user_cancel`.
3. Spool rewrite: temp created via `os.open(O_CREAT|O_EXCL|O_NOFOLLOW, 0o600, dir_fd=fd)` with a random name, replaced via `os.replace(..., src_dir_fd=fd, dst_dir_fd=fd)`; `mkstemp` only on the no-descriptor fallback. Test: `test_swapped_spool_directory_never_receives_the_rewrite`.
4. One-shot watchdog gate under a lock; completed callback reaps the watchdog. Test: `test_grace_boundary_race_fires_the_interrupt_exactly_once`.

Conductor verification through Brigade: ruff check and format (`+1`, `+1`); six-file pytest with the two #1182 deselects: 252 passed (`+1`). Re-dispatching daybreak against bd76e603.

solomonneas:
Daybreak round-3 re-review against bd76e603 (brigade run `19faf9b8.20260826-021245-db6f0e27`). All four round-2 residuals confirmed closed on POSIX; one regression the fix introduced on Windows. Relayed by the conductor; fix lane dispatched.

BLUF: One medium-severity regression remains. The four round-2 residuals are closed on POSIX.

Finding:

- **Medium, Windows spool failure:** [`_ensure_private_dir()`](src/brigade/fleet_client.py:407) promises `None` when `dir_fd` is unsupported but always returns an fd at line 456. `_spool_lock()` then performs unsupported descriptor-relative operations at line 508. Python documents that `dir_fd` does not work on Windows. The exception is swallowed at lines 815-816, silently losing undelivered fleet events. [Python documentation](https://docs.python.org/3.12/library/os.html#files-and-directories)
- **Fix:** On unsupported platforms, use path-based privacy handling and return `None`, activating the existing `mkstemp`/`os.replace` fallback. Add a native or simulated Windows spool-and-rewrite test.

Confirmed closed:

- Credential marker precedes diagnostics; interrupt runs in `finally`.
- Claim-loss marker precedes diagnostics.
- POSIX temp creation and replacement use the validated `dir_fd`.
- Watchdog and callback share a locked one-shot gate.
- Retry collisions leak no fds; successful temp fds close through `fdopen`.
- Claim-loss interrupt fires once.

NOT READY

solomonneas:
Coordination: the Fleet Cloud Workers / Command Deck orchestrator (branch `t3code/fleet-command-deck-spec`, specs `docs/phase-fleet-cloud-workers*.md`) has reserved `src/brigade/fleet_client.py`, `src/brigade/cli/fleet.py`, and the fleet tests. This PR lives entirely in those files, so the conductor stops driving it here.

State at handoff: rounds 1 to 3 of the daybreak review are closed on POSIX (see the comments above). One item is open from the round-3 re-review: on Windows `_ensure_private_dir` returns a descriptor although `dir_fd` is unsupported there, so spool writes fail silently; a fix lane was already running when the reservation arrived and its result, if it verifies, will be pushed to this branch only (no merge). After that, no further changes from the conductor. Whoever owns `fleet_client.py` next can rebase or absorb this branch; the review trail and the residual list are complete.

solomonneas:
Final conductor change, 727ee1bc (rebased onto current main): the Windows regression from the round-3 re-review is closed. `_dir_fd_supported()` gates the descriptor-relative spool path (platform check plus `os.supports_dir_fd` membership of `os.open`, `os.rename`, `os.unlink`, `os.stat`; note `os.replace` registers under `os.rename`), `_ensure_private_dir` returns `None` where unsupported so the `mkstemp`/`os.replace` fallback runs, and failed spool writes log at WARNING instead of being swallowed. Five regression tests in `TestSpoolPathFallbackWithoutDirFd` simulate the no-dir_fd platform end to end through `report_event`.

Conductor verification through Brigade: ruff check and format (`+1`, `+1`); five-file pytest with the two #1182 deselects: 177 passed (`+1`).

No further changes from the conductor; ownership of `fleet_client.py` and this branch passes to the Command Deck orchestrator per the note above. Suggested next step for them: one daybreak pass on 727ee1bc, then rebase or absorb.

solomonneas:
Final daybreak review (read-only, hub-tracked run in conductor-1157) on the merged tree: NOT READY on two design-level liveness gaps and one test-hygiene item, none of which this PR regresses; per the wave's last-round rule they are filed as #1219 and noted in the fleet_client.py docstring (419e6857):

- HIGH a node whose claim renewals are selectively blocked retries hub-unavailable without a local lease-expiry deadline (fleet_client.py:1643).
- HIGH a worker blocked in a silent child wait outlives a lost claim until the queued KeyboardInterrupt lands (fleet_client.py:1418, cli/run.py:536).
- LOW the proxy test can pass under NO_PROXY covering 127.0.0.1 (test_fleet_client_hardening.py:40).

Cleared: holder fencing, external-session derivation after #1216, descriptor-relative spool fallback, TLS context. Branch history was squashed onto origin/main with the merged tree (the fleet_client import union with #1216) so the pushed diff carries only this change; gates after the merge (Brigade-captured): ruff check, ruff format --check, mypy, eight fleet suites plus test_run_cli 380 passed. Marking ready; merging on green CI. Command Deck: the same commits are already in your t3code/fleet-command-deck-spec branch via origin/pr-1181, so a rebase onto main after this merge should drop them as duplicates.

solomonneas:
CI note: no pull_request workflow run was created for 419e6857 or 8fb98150 because GitHub Actions entered a major outage at 15:11 UTC today (githubstatus: 'Incident with Actions', investigating). Reruns of existing runs still execute, new runs are not being created. The branch is mergeable against current main (merge-tree clean). Merging as soon as Actions recovers and a green run exists for 8fb98150.

Issue #1093, opened 2026-08-22
scanner authority store: true same-uid boundary (OS keyring / separate-uid service) - hardening track


solomonneas:
Follow-up to #957. The operator chose Option A (user-level sticky signing marker + confirmed-downgrade command, implemented in #1078). Option A raises the bar for the same-uid scanner authority-store bypass but is defense in depth, not a hard boundary: an attacker who also owns `~/.brigade` can still remove the marker. This issue tracks the true-boundary options for operators who need one:

- **(B) OS-mediated secret / keyring**: store the HMAC key in an OS keyring (Keychain / Secret Service / DPAPI) so a same-uid file write cannot read it. Adds a platform-specific dependency and code paths; must degrade cleanly (WARN, fall back to Option A) where no keyring is available so zero-dependency quickstart still works.
- **(C) Separate-uid authority service**: the authority store is owned by a different uid and mediated by a small local service; the scanner cannot write it directly. The only fully-real boundary against a same-uid attacker, but the largest build and it breaks the zero-dependency quickstart promise.

Neither is on the 0.28.0 path. File as a hardening track; pick B or C (or both, keyring-first) when an operator has a real same-uid-adversary threat model. Doctor should keep honestly naming the residual class until one lands.

solomonneas:
Burn disposition: skipped for this run. The routing brief classifies this as a tracking issue for the same-uid boundary and supplies no bounded implementation slice. No implementation attempt or branch was started; future work should begin from a separately scoped issue with its own acceptance test.

PR #1185, opened 2026-08-25
fix(work): bind legacy receipt and proof bytes to verifier authority

### LINKED ISSUE #881: fix(work): bind scanner receipt and proof bytes to verifier authority

solomonneas:
Fixes #881.

## Root cause

The legacy import migration path trusted scanner-reproducible receipt and persisted-proof bindings recorded in an unsigned external authority store. A same-uid writer could forge a post-run `receipt.json` (and matching sidecar proof) and re-bind the unsigned store to the new bytes, granting the forged row a legacy source/content identity during untrusted-identity migration.

## Change

`_legacy_import_source_content_identity` in `src/brigade/work_cmd/ledger.py` now requires the workspace's authority record to be a verifier-signed HMAC envelope; unsigned stores no longer confer legacy identity. Regression tests in `tests/test_authority_store_hmac.py`, `tests/test_work_cmd_sweeps.py`, and `tests/test_work_import_provenance_stamp.py` cover the forged post-run receipt being rejected. CHANGELOG entry under Unreleased.

## Verification

Conductor, through Brigade on this tree: `ruff check` and `ruff format --check` on the changed files (`+1`, `+1`); `pytest -q tests/test_authority_store_hmac.py tests/test_work_cmd_sweeps.py tests/test_work_import_provenance_stamp.py`: 197 passed (`+1`). Full `./scripts/verify` runs in CI.

Security-flavoured change: per roster policy the Ox seat is not the primary security reviewer, so this wants a `daybreak` or Claude review pass before merge.

## Provenance

Implemented by the `ox_worker` seat (`opencode/x-preview-f-free`) under `brigade run` `2904d3ea.20260825-213535-7cf0ef8c`; committed, rebased onto main, and filed by the conductor.


solomonneas:
Read-only security review by the `daybreak` seat (codex `gpt-daybreak-blue-latest`, brigade run `2904d3ea.20260825-221053-276836d8`, native read-only sandbox). Relayed by the conductor; the reviewer reports in-memory probes for the High and the first Medium, not yet independently reproduced. Converting to draft until all three are addressed.

BLUF: PR #1185 still has one high-severity bypass and two medium integrity flaws. Very likely, confidence High - control flow was traced and confirmed with in-memory probes.

## Findings

### High: alternate legacy identity path bypasses signed authority

- **Where:** [ledger.py:4183](src/brigade/work_cmd/ledger.py:4183)
- **Scenario:** An unsigned-workspace attacker uses `source_item_id` matching the incoming canonical key, forges a successful receipt, and rebinds the unsigned store. `existing_migration_proof` checks only `_has_locally_stamped_import_proof`, bypassing the new signed-store helper and persisted sidecar requirement. The incoming record is skipped or remains dismissed.
- **Probe:** `legacy_helper_none True`, `unsigned_receipt_predicate True`, `result 0 1 0 0`.
- **Fix:** Use one shared authenticated-legacy predicate requiring the receipt, persisted sidecar, and signed authority record from a single verified snapshot.

### Medium: signedness is checked against a different store read

- **Where:** [ledger.py:3272](src/brigade/work_cmd/ledger.py:3272)
- **Scenario:** The helper HMAC-verifies one read, then reopens the path and only checks `envelope_version`. A concurrent same-UID writer can present unrelated bytes on the second read.
- **Probe:** `accepts_unrelated_second_read True`.
- **Fix:** Return signedness and the verified record from the same descriptor read. Validate receipt, proof, and workspace against that single authenticated payload.

### Medium: canonical dedupe still trusts unsigned sidecar bindings

- **Where:** [ledger.py:4180](src/brigade/work_cmd/ledger.py:4180)
- **Scenario:** Without external-key isolation, a writer can plant a canonical existing row plus reproducible sidecar and unsigned binding. `canonical_existing_proof` then suppresses the genuine incoming record without reaching the signed-store guard.
- **Fix:** Require signed authority for every proof used to suppress imports, including canonical dedupe.

## Clean areas

- HMAC covers the canonical full record, including target, workspace identity, receipt/proof file bindings, and sequence.
- Key ID validation and MAC comparison use `hmac.compare_digest`.
- The direct legacy helper rejects fresh unsigned and intentionally downgraded stores.
- Signed honest legacy migration and workspace relocation remain represented in regression tests.

Verification: `git diff --check origin/main...HEAD` exited 0; `git status --short` was empty. Pytest and Brigade receipt verification could not run because the read-only sandbox has no writable temporary directory. No files were changed.

Next: gate all migration and dedupe paths on one authenticated authority snapshot, then add alternate-source-key and concurrent-store-swap tests.

solomonneas:
Sendback outcome (ox_worker, brigade run `2904d3ea.20260825-221740-a78268ef`, verify-then-fix): all three findings reproduced with a failing test first, then fixed in 6d8a429f.

- High, alternate legacy identity path: `test_alternate_legacy_identity_path_requires_authenticated_proof` failed (`0 == 1` imported); fix is a shared `_authenticated_legacy_import_proof` predicate (receipt + persisted sidecar + signed authority snapshot) used by both `_legacy_import_source_content_identity` and the `existing_migration_proof` path.
- Medium, second-read swap: `test_authority_signedness_comes_from_one_verified_read` failed; `_read_verified_authority_snapshot` reads once, requires the raw payload to be an HMAC envelope, and MAC-verifies that exact payload; `_authority_store_binding_is_verifier_signed` delegates to it.
- Medium, unsigned canonical dedupe: `test_unsigned_canonical_dedupe_cannot_suppress_incoming` failed; `canonical_existing_proof` now requires `_has_persisted_import_proof` and a signed snapshot from the same single-read helper.

Conductor verification through Brigade on the tree: ruff check and format (`+1`, `+1`), `pytest -q` on the three ledger test files: 200 passed (`+1`). Re-dispatching the daybreak review against 6d8a429f; stays draft until it clears.

solomonneas:
Daybreak re-review against 6d8a429f (brigade run `2904d3ea.20260825-222824-cae4105b`): the three original findings are confirmed closed; two remain. Relayed by the conductor. Round-2 fix lane dispatched; PR stays draft.

BLUF: NOT READY. Very likely, confidence Moderate - static tracing found two unresolved paths; runtime tests were excluded by the read-only constraint.

### Findings

- **HIGH - Authority snapshot is not bound to validated artifacts.**  
  **Where:** [ledger.py:3314](src/brigade/work_cmd/ledger.py:3314), [ledger.py:2703](src/brigade/work_cmd/ledger.py:2703), [ledger.py:4227](src/brigade/work_cmd/ledger.py:4227)  
  **Scenario:** Receipt and sidecar checks reread authority state independently. The verified snapshot’s record is discarded. A concurrent same-UID writer can alternate forged unsigned bindings and a saved valid signed envelope, subject to clearing the documented defense-in-depth marker or catching its publication gap. Legacy identity or canonical suppression can then combine evidence that never existed in one authenticated snapshot.  
  **Fix:** Reanchor first, read one signed snapshot, and pass that exact record into receipt, directory, and sidecar validators. Add a regression test that swaps state between artifact validation and the final predicate.

- **MEDIUM - First canonical dedupe after relocation imports a duplicate.**  
  **Where:** [ledger.py:4227](src/brigade/work_cmd/ledger.py:4227)  
  **Scenario:** After renaming a workspace, the new authority-store path does not exist. Signedness is checked before `_has_persisted_import_proof()` triggers reanchoring, so `authority_is_signed` remains false and the matching canonical import is accepted once. Existing relocation tests pre-open the proof directory, masking this ordering bug at [test_authority_store_hmac.py:1066](tests/test_authority_store_hmac.py:1066).  
  **Fix:** Reanchor before acquiring the snapshot and add an end-to-end canonical dedupe test that renames the workspace without pre-opening authority directories.

### Clean

- The direct `source_item_id` / `existing_migration_proof` bypass is routed through the shared predicate.
- `_read_verified_authority_snapshot()` itself verifies HMAC over the exact bytes read.
- Static unsigned-sidecar dedupe is rejected.
- Honest signed legacy imports and the legacy relocation path remain covered.
- `git status --short` and `git diff --check origin/main...HEAD` produced no output. No files were changed.

Next: bind all proof validation to one post-reanchor snapshot and add the two missing regression tests.

NOT READY

solomonneas:
Round-2 outcome (ox_deep, brigade run `2904d3ea.20260825-223355-6572d343`): both remaining findings reproduced with failing tests first, fixed in 7eeb05c3.

- High, validators re-reading authority: `test_store_swap_between_validators_cannot_combine_snapshots` (swaps two validly signed envelopes between artifact validation and the final predicate) failed before, passes after. `_acquire_authenticated_authority_snapshot` reanchors then reads one verified record; the receipt reader, proof predicates, and directory/file validators gained a `record=` mode that validates against that snapshot and never reopens the store.
- Medium, relocation ordering: `test_canonical_dedupe_after_rename_without_preopening_authority` failed before (duplicate re-imported), passes after. The snapshot helper reanchors before its single read and `_append_import_records` acquires it once up front.

Conductor verification through Brigade: ruff check and format (`+1`, `+1`); `pytest -q` on the three ledger test files: 202 passed (`+1`).

Open question for the maintainer before this leaves draft: the round-1 rule that every suppression needs a signed authority store makes 9 existing tests fail (`tests/test_work_cmd_imports.py` x8, `tests/test_work_cmd_scanners.py::test_work_scanners_ingest_output_sanitizes_hostile_records_and_owns_identity`), all green at the parent `75f40f93`. Their fixtures are default workspaces without `authority_store.isolation = "external-key"`, where no signed store exists, so canonical dedupe now fails closed and stops suppressing duplicates for default-config installs. Either the fixtures model external-key isolation (accepting that default workspaces lose dedupe) or a documented unsigned-dedupe trust path is defined for non-isolated workspaces. Not changing those tests without agreement.

solomonneas:
Posture split landed in 2d2b9292 per maintainer decision (ox_deep, brigade run `2904d3ea.20260826-001518-6fb9b019`).

- Legacy identity grants (A): unchanged, signed-only via `_authenticated_legacy_import_proof` and the single verified snapshot, in every workspace.
- Canonical dedupe suppression (B): new named predicate `_unsigned_dedupe_proof_for_non_isolated_workspace`; returns false whenever `authority_key.hmac_enabled(target)` (isolation `external-key`), so isolated workspaces keep the signed rule. With isolation off and no verified snapshot it admits the pre-#881 unsigned path and logs one warning per process naming the downgrade. Documented in `docs/security.md` ("Canonical import dedupe and the isolation posture") and the CHANGELOG.
- The 9 default-workspace tests pass unchanged. Round-2's `test_unsigned_canonical_dedupe_cannot_suppress_incoming` was a non-isolated fixture asserting rejection; it is converted to external-key isolation with the outcome assertions intact (signed store plus same-uid-forged sidecar bytes). New tests: unsigned dedupe works in a non-isolated workspace (with the once-per-process warning), and is refused in an isolated one.

Conductor verification through Brigade: ruff check and format (`+1`, `+1`); `pytest -q` over the three ledger files plus `test_work_cmd_imports.py` and `test_work_cmd_scanners.py`: 417 passed (`+1`). Re-dispatching the daybreak review against 2d2b9292.

solomonneas:
Daybreak round-3 review against 2d2b9292 (brigade run `2904d3ea.20260826-003924-33f32491`). Relayed by the conductor. Legacy identity, single-snapshot binding, reanchoring, and the converted fixture confirmed clean; one Medium on the implementation of the posture split (the isolation toggle must be anchored outside the workspace), plus a Low and an Info. Follow-up lane dispatched; stays draft.

BLUF: NOT READY. Very likely, confidence High - the isolation selector trusts scanner-writable workspace configuration.

### Findings

**[MEDIUM] Anchor isolation posture outside the workspace**

- **Where:** [ledger.py:3474](src/brigade/work_cmd/ledger.py:3474), [ledger.py:4369](src/brigade/work_cmd/ledger.py:4369), [config.py:255](src/brigade/security_cmd/config.py:255)
- **Scenario:** When the signed snapshot is missing, malformed, or mid-rotation, a same-uid scanner can change `.brigade/security.toml` to `off` or malformed. Invalid configuration is normalized to `off`, selecting the unsigned path. Before an external sticky marker exists, attacker-controlled raw authority and sidecar records can suppress a canonical import.
- **Fix:** Persist external-key posture per target outside the workspace when isolation is enabled. Refuse unsigned fallback when that marker exists or configuration is unreadable/invalid; only confirmed downgrade may clear it.

**[LOW] Cover the actual snapshot-failure branch**

- **Where:** [test_authority_store_hmac.py:927](tests/test_authority_store_hmac.py:927)
- **Scenario:** The test deletes a sidecar but retains a valid signed snapshot at line 950. It does not cover missing, malformed, or rotating stores, config mutation, or its claimed absence of a warning.
- **Fix:** Parameterize those store states, assert the recordless proof validator is never called, and capture stderr.

**[INFO] Correct the downgrade warning**

- **Where:** [ledger.py:3463](src/brigade/work_cmd/ledger.py:3463)
- **Issue:** The docstring and warning claim receipt validation, but the fallback at line 4406 validates only the persisted sidecar. `docs/security.md` describes the code correctly.
- **Fix:** Remove the receipt claim.

### Clean

- Legacy identity grants remain signed-only.
- Validators consume one post-reanchor snapshot; store-swap rejection holds.
- Rename reanchoring remains intact.
- The converted #881 fixture still performs a genuine post-run sidecar forgery.
- Worktree and `git diff --check` were clean. No files changed; tests were not executed under the report-only constraint.

NOT READY

solomonneas:
Posture-anchor outcome (ox_deep, brigade run `2904d3ea.20260826-004438-c0efad16`): the Medium reproduced and is fixed in 9477af86 (rebased onto current main).

- Marker: per-target file beside the existing `authority-signed` sticky state under the user directory (`authority_marker.py`), keyed by the same target fingerprint the signed marker and store use; recorded the first time genuine external-key isolation is observed (`authority_store_isolation_state` in `security_cmd/config.py`).
- `_unsigned_dedupe_proof_for_non_isolated_workspace` refuses when the marker exists or when `security.toml` exists but cannot be parsed; the unsigned path is allowed only for a healthy, genuinely-off workspace with no marker. Clearing is wired into the existing audited `brigade security authority downgrade` (removes both markers with backup/restore; no new CLI verb), documented in `docs/security.md`.
- Tests: `test_unsigned_dedupe_refused_when_posture_marker_survives_config_flip` (plus the corrupted-config variant); the isolated-refusal test is parameterized over valid, missing, malformed, and rotating stores, asserts the recordless validator is never called, and asserts no downgrade warning on stderr. Warning and docstring no longer claim receipt validation.

Conductor verification through Brigade: ruff check and format (`+1`, `+1`); `pytest -q` over the five ledger/import/scanner files plus the center tests touching config: 438 passed (`+1`). Re-dispatching daybreak against 9477af86.

solomonneas:
Daybreak round-4 review against 9477af86 (brigade run `2904d3ea.20260826-021250-fc2937e9`). Marker placement, scanner sandboxing, signed-only identity, HMAC reanchor, and the audited downgrade are confirmed clean; three residuals in the marker contract. Relayed by the conductor. One final Ox round dispatched; if it does not clear, this PR freezes as draft with the residual list for a design pass.

BLUF: NOT READY. Very likely, confidence High - two fail-open paths and one snapshot-contract defect remain.

### [MEDIUM] Marker-store absence enables unsigned fallback

- **Where:** [config.py:255](src/brigade/security_cmd/config.py:255), [authority_marker.py:177](src/brigade/authority_marker.py:177), [ledger.py:3504](src/brigade/work_cmd/ledger.py:3504)
- **Scenario:** marker persistence errors are ignored, while a missing or unreadable marker root is treated as “marker absent.” After the workspace config is flipped off and the signed store is replaced, forged unsigned authority and sidecar evidence can suppress a canonical import.
- **Fix:** make marker status tri-state. Permit fallback only after positively reading a healthy marker store and confirming absence. External-key observation must fail closed if marker persistence fails.

### [MEDIUM] Path-only fingerprints lose posture across rename

- **Where:** [authority_marker.py:62](src/brigade/authority_marker.py:62), [ledger.py:2370](src/brigade/work_cmd/ledger.py:2370)
- **Scenario:** moving workspace A to B leaves both markers keyed to A. Before B observes external-key mode, an attacker can flip its config off and plant raw authority at B. The old marker does not block B’s unsigned fallback. Reusing path A also applies stale posture to an unrelated workspace.
- **Fix:** bind posture to stable workspace identity, or transfer and verify posture during authenticated reanchor before evaluating unsigned state.

### [LOW] “No snapshot” is mistaken for “acquire another”

- **Where:** [ledger.py:3465](src/brigade/work_cmd/ledger.py:3465), [ledger.py:4394](src/brigade/work_cmd/ledger.py:4394), [ledger.py:4416](src/brigade/work_cmd/ledger.py:4416)
- **Scenario:** when the append-level acquisition returns `None`, the legacy helper interprets that value as “unset” and reopens authority state. The append therefore does not always consume one snapshot.
- **Fix:** use an explicit sentinel, or a helper that never reacquires when called from the append path.

Clean: workspace and symlink marker placement is rejected; scanner-child HOME and authority state are sandboxed; legacy identity remains signed-only; signed reanchor candidates require a valid HMAC; the downgrade is the only code clearer and performs audited rollback.


**NOT READY**

solomonneas:
Final-round outcome (ox_deep, brigade run `2904d3ea.20260826-023140-594e022a`): all three round-4 findings reproduced with failing tests first, fixed in a5712440 (rebased onto current main).

1. Medium, marker absence: `isolation_marker_status()` is tri-state (present / confirmed-absent / unknown; unreadable root or entry, malformed bytes, non-regular entries are unknown). Only confirmed-absent permits the unsigned path. A posture-persistence OSError during external-key observation now refuses to proceed (doctor reports FAIL; the ledger predicate refuses). Tests: `test_unsigned_dedupe_refused_when_isolation_marker_root_unreadable`, `test_external_key_observation_fails_closed_when_posture_persistence_fails`.
2. Medium, path-keyed posture: markers are schema v2 bound to the workspace `{device, inode}` identity the signed authority record carries; `_reanchor_external_directory_authority` transfers and re-verifies posture via `transfer_isolation_marker` before store adoption and before any unsigned evaluation; lookup matches by identity, so a moved workspace stays governed and an unrelated workspace at a recycled path is not blocked. Tests: `test_rename_without_preopening_carries_posture_and_refuses_unsigned_fallback`, `test_stale_posture_marker_does_not_govern_unrelated_workspace_at_old_path`.
3. Low: `_APPEND_SNAPSHOT_ABSENT` sentinel from `_append_import_records`; the legacy helper never re-acquires. Test: `test_append_consumes_exactly_one_authority_snapshot`.

Conductor verification through Brigade: ruff check and format (`+1`, `+1`); `pytest -q` over the six ledger/import/scanner/security files: 538 passed (`+1`). Re-dispatching daybreak against a5712440; flips to ready on READY.

solomonneas:
Daybreak round-5 review against a5712440 (brigade run `2904d3ea.20260826-030114-8c16d5c7`). Relayed by the conductor.

Status: FROZEN as draft at a5712440 with the residual below. Five review rounds are done; the original #881 forgery (signed-only legacy identity), the single-snapshot binding, reanchor transfer, tri-state marker handling, persistence refusal, and the audited downgrade are all confirmed clean. What remains is one Medium in the marker identity contract (identity churn on non-stable filesystems) and one Low (v1 markers not rewritten in place). The maintainer decides whether this gets one more narrow round or a design pass on stable workspace identity.

BLUF: NOT READY. Very likely, confidence High - one identity-mismatch path can restore unsigned fallback.

Findings:

- **[MEDIUM] Identity churn can reopen unsigned dedupe.** [authority_marker.py:266](src/brigade/authority_marker.py:266), [authority_marker.py:385](src/brigade/authority_marker.py:385), [ledger.py:3544](src/brigade/work_cmd/ledger.py:3544)
  - Scenario: a v2 marker exists, but the root device/inode changes due to a network filesystem, cross-filesystem copy, or recycled path. Existing-marker observation returns without rebinding it. After config is flipped off and signed authority becomes unavailable, the identity mismatch is treated as non-governing, producing `confirmed-absent` and admitting unsigned suppression.
  - Fix: refresh a mismatched marker whenever genuine `external-key` is observed. Treat same-path identity mismatch as `unknown` unless an authenticated transfer or explicit downgrade released the old binding.

- **[LOW] Schema v1 markers are not migrated in place.** [authority_marker.py:385](src/brigade/authority_marker.py:385)
  - Existing regular markers return immediately without schema validation or v2 rewrite. V1 remains fail-closed and transfers to v2 during rename, but stays path-keyed otherwise.
  - Fix: validate and atomically rewrite v1 markers to v2 during external-key observation; add migration and stale-destination tests.

Clean:

- Tri-state unreadable/malformed handling and persistence-failure refusal are closed.
- Same-filesystem rename transfer occurs before store adoption and is reverified.
- The append sentinel prevents `None` snapshot reacquisition.
- Legacy identity grants remain signed-only and consume one authenticated snapshot.
- Inode reuse currently fails closed when identities coincide.
- HMAC verification, target binding, explicit downgrade, and rollback guarantees remain intact.

NOT READY

solomonneas:
Final daybreak review (read-only, hub-tracked run in conductor-881) on 3c576390 with origin/main merged in: READY. No HIGH or MEDIUM in-model finding. Cleared: unsigned receipts or proofs cannot grant legacy identity; ledger content cannot spoof isolation selection; legacy and raw authority records cannot bypass verifier binding; #1207's canonical lock verification remains in place at ledger.py:4698; nothing new touches the #1211 skills-state writers. One LOW in-model cleanup item (relocation-aware downgrade leaves the source marker governing) filed as #1217; the same-uid state-root rewrite stays a #1093 residual as documented on this PR.

Posture as decided by the operator: signed-only for identity grants; unsigned dedupe allowed only in non-isolated workspaces. Gates after the merge (Brigade-captured): ruff check, ruff format --check, mypy, ten authority/imports/sweeps/provenance/security/scanners/ledger/signing/redaction suites 728 passed. Marking ready; merging on green CI.

PR #1193, opened 2026-08-26
fix(harness): report pending, not current, on conflict-free dry runs

### LINKED ISSUE #1170: harness sync dry-run reports OpenCode current with pending writes
## Summary

`brigade harness sync --dry-run` can report an OpenCode profile as `status: current` and `ready: true` while the same result contains 125 missing files and a missing profile receipt. `--check` then exits 1 and reports the managed instruction as missing.

Tested with Brigade 0.27.0 on Linux.

## Reproduction

Run against a Brigade workspace whose user-scope OpenCode profile has not been applied:

```bash
brigade harness sync \
  --target opencode \
  --scope user \
  --workspace "$WORKSPACE" \
  --dry-run \
  --allow-global-stdio \
  --json
```

The OpenCode result included:

```json
{
  "harness": "opencode",
  "status": "current",
  "ready": true,
  "receipt_state": "missing"
}
```

The same result's items contained 125 entries with `status: missing` and 2 with `status: stale`, including the managed instruction, skill files, ownership state, and profile receipt.

Then run:

```bash
brigade harness sync \
  --target opencode \
  --scope user \
  --workspace "$WORKSPACE" \
  --check \
  --profile full \
  --allow-global-stdio
```

This exits 1 and reports:

```text
opencode: conflict ...
  instruction missing: fix with: brigade harness sync --target opencode --scope user --write
```

## Expected behavior

A dry-run with pending writes should not label the installed profile `current`. The result should distinguish "safe to apply" from "already applied and current." `ready` may remain an apply-readiness field if that is the intended contract, but `status` and `receipt_state` should describe installed state consistently with `--check`.

## Likely source

The non-write return path in `harness_profile_cmd.py` derives `status` from whether conflicts exist. Missing planned writes are not conflicts, so the result becomes `current` even though no files were written and the receipt is absent.

solomonneas:
## Summary

`brigade harness sync --dry-run` could return `status: current` and `ready: true` while the same payload contained 125 missing items and `receipt_state: missing`, and while `sync --check` exited 1 calling the managed instruction missing. The non-write return path in `_sync_profile` derived status purely from conflicts, and pending writes are not conflicts.

The result now distinguishes safe-to-apply from already-applied:

- `conflict`: unsafe to apply (unchanged)
- `pending`: conflict-free but items plan create/update/remove (new)
- `current`: fully applied (now means it)
- `updated`: a write run applied something (unchanged)

`ready` remains the apply-readiness field (`not conflicts`) and still drives exit codes, per the contract discussion on the issue. `_uninstall_profile`'s non-write path had the identical flaw and gets the same fix.

Fixes #1170

## Test plan

- [x] New regression tests: dry-run reports `pending`/`receipt_state missing` before apply and `current`/`present` after, across codex (slice1) and all five slice2 harnesses including opencode; uninstall dry-run parity test
- [x] Targeted harness suites green via Brigade verify receipt (120 tests, exit 0)
- [x] ruff check / ruff format / mypy / version-sync / snapshot checks green
- [x] Full suite raw run: 8983 passed, coverage 83.69% (floor 78%); the 8 failures are environment-only on this VM (TMPDIR under home + enrolled fleet node identity) and reproduce identically on unmodified HEAD

### DIFF
--- a/src/brigade/harness_profile_cmd.py
+++ b/src/brigade/harness_profile_cmd.py
@@ -1985,9 +1985,12 @@ def _sync_profile(
             migration="legacy-v1" if migrated_from_legacy else None,
         ), bool(files_written or files_removed)
     receipt_state = "present" if profile.receipt_path.exists() else "missing"
+    # A conflict-free plan with planned writes is "safe to apply", not
+    # "already applied".  Only a fully applied profile is "current".
+    pending = any(item["action"] in {"create", "update", "remove"} for item in items)
     return _result(
         profile,
-        status="current" if ready else "conflict",
+        status="conflict" if conflicts else ("pending" if pending else "current"),
         ready=ready,
         items=items,
         conflicts=conflicts,
@@ -2268,9 +2271,12 @@ def _uninstall_profile(
                 _prune_created_directories(instruction_dirs + generated_plan["prune"], profile.user_root)
             )
     changed = bool(files_removed or files_written)
+    # Dry-run parity with sync: planned removals mean the profile is pending
+    # removal, not already gone.
+    pending = any(item["action"] in {"create", "update", "remove"} for item in items)
     return _result(
         profile,
-        status="conflict" if conflicts else ("updated" if changed else "current"),
+        status="conflict" if conflicts else ("updated" if changed else ("pending" if pending else "current")),
         ready=not conflicts,
         items=items,
         conflicts=conflicts,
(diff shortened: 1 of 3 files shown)


solomonneas:
Daybreak review (read-only, hub-tracked) on 5d655bb5: the fix is confirmed (pending covers every mutation kind plus state and receipt items, no other consumer compares status to current, uninstall symmetry holds, the new tests fail against origin/main). One MEDIUM adjacent gap outside this PR's claim: _doctor_profile, which --check and doctor use, skips ownership-state and receipt validation and reports current on that drift; filed as #1218 rather than blocking this fix. Merging on green CI.

Issue #1144, opened 2026-08-23
bug: preserve worker result when app-server output exceeds 1 MiB


solomonneas:
## Problem

`brigade run` can lose a completed worker result when the app-server stream crosses the 1 MiB combined-output cap.

The worker may finish its edits and verification before the harness reports:

```text
combined output exceeded 1048576 byte limit; same-seat retry skipped because re-probe reported unhealthy [transport-unavailable]
```

This happened in three Grok Bot queue runs:

- `8e7b7f3a.20260823-204713-2fd4e96c`
- `8e7b7f3a.20260823-205616-c58ec38c`
- `8e7b7f3a.20260823-210720-877f89ce`

The third worker had already passed 69 focused tests, Ruff, mypy, and command-inventory validation. Its edits remained in the worktree, but the run was classified as failed and produced no normal synthesis result.

## Expected behavior

- Preserve a bounded final worker result when earlier transport output is truncated.
- Classify the failure as an output-limit condition, not a seat-health failure.
- Avoid re-probing the seat as unavailable solely because the stream exceeded its byte cap.
- Retain enough metadata to recover completed edits and verification receipts.

## Actual behavior

- The run ends as failed after the output cap is crossed.
- The seat is re-probed and reported as transport unavailable.
- The orchestrator receives no normal worker result even when the worker finished its task.

## Acceptance criteria

- A regression test covers a worker stream larger than 1 MiB followed by a small final result.
- The final result remains available, or the run reports a recoverable truncation state with the worktree and receipts attached.
- Output truncation does not poison seat health.
- The run record includes the original byte count and configured cap.

Brigade evidence: `3e98b3d7d34398472b7a7499`.

solomonneas:
Additional reproduction: `4b5ea550.20260823-230111-e61e7cb4` used `--codex-transport exec` and failed the same way before producing an edit.

Both captured stderr logs are exactly 1,048,576 bytes:

```text
1048576 logs/worker-001-coder.stderr.log
1048576 logs/worker-001-coder-attempt-001-initial.stderr.log
```

The log contains normal Codex CLI session and tool output, not a repeated warning flood. This confirms the failure is not limited to the app-server transport. The cap is also consuming the worker's normal execution stream before Brigade can retain a final result.

solomonneas:
Regression reproduced twice on current `main` (`c19edf5b`) with two different OpenAI seats through the app-server transport on 2026-08-31.

[...]

Issue #1015, opened 2026-08-19
security: second full Daybreak-class pass over ledger.py, scanners.py, run_transport.py, app.go


solomonneas:
Part of #1006 (Daybreak Blue defensive pass 2026-08-19, coverage note).

The Daybreak Blue pass sampled the biggest files rather than reading them in full, so the confirmed findings are a floor, not a ceiling. Schedule a second full defensive pass at the same bar over `ledger.py`, `scanners.py`, `run_transport.py`, and `app.go`, reading each end to end.

Coverage paragraph from #1006, verbatim:

**Coverage (from the model):** vault-search surface read fully; `trust_gate.py`, `message_envelope.py`, `evidence_redaction.py`, `memory_vault.py`, `obsidian_vault.py`, `receipt_signing.py`, Go `integrity.go`, and the projection kernel read fully; the large ledger, scanner, transport, `app.go`, hook-runtime, and subprocess surfaces traced selectively. So this is not exhaustive on the biggest files - a second pass over `ledger.py`/`scanners.py`/`run_transport.py` in full is worth scheduling.

## Fix bar

Any finding this pass produces gets its own issue in the same shape as the #1006 split: verbatim attack, files with line refs, fix sketch, and a fix lane that ships a mutation test reproducing the attack plus independent executed-probe review before merge.

solomonneas:
Second-pass audit, part 1 of 3 (scanners.py and run_transport.py on main 06e61919), by the `daybreak` seat (codex `gpt-daybreak-blue-latest`, brigade run `4c2cb171.20260826-030528-ffc4ceb8`, read-only sandbox). Relayed by the conductor. `ledger.py` is deferred until PR #1185 settles; the Go engine files are a separate run. The four new findings are filed as their own issues (linked below).

BLUF: Audit of `06e61919` found five actionable paths: four in `scanners.py`, one in `run_transport.py`. No files were modified.

### `src/brigade/work_cmd/scanners.py`

- **HIGH, #1006-1 remains open by default:** HOME/XDG isolation is only environmental (`126-175`); scanner isolation defaults off (`1727-1733`) and is conditional (`1845-1854`). A same-UID scanner can retain a background child and access verifier state after return. Make isolation mandatory, use a separate UID/mount namespace, and terminate the entire process group.
[...]

PR #1206, opened 2026-08-26
fix(run): salvage an over-cap app-server result only when the turn completed

### LINKED ISSUE #1200: run transport: over-cap app-server result salvaged as ok=True fails open for DAG prerequisites
From the #1015 second-pass audit (part 1, main 06e61919). Severity: Medium. Follow-up to #1144 / PR #1188.

`src/brigade/run_transport.py` (around L847-874) converts any non-empty result carrying `failure_kind="output-limit"` into `ok=True`, and the retry/quarantine gate (around L918-923) then excludes it. A compromised or misbehaving seat can emit a plausible prefix, flood the stream past the cap, and satisfy DAG prerequisites without a completed turn.

Fix: keep the salvaged text as a diagnostic and record `output_truncated`, but keep `ok=False` (and the seat un-poisoned, which was the intent of #1144) unless an authenticated completion event for the turn was observed; downstream stages should treat a truncated result as not satisfying prerequisites.

Refs #1015, #1006, #1144.

solomonneas:
Fixes #1200. Follow-up to #1144 / #1188, from the #1015 second-pass audit.

## Root cause

`run_transport.finish` set `preserved = _is_appserver_output_limit(result) and nonempty text`, so any over-cap app-server result with a plausible prefix became `ok=True`, and the retry/quarantine gate excluded it. Nothing bound `ok` to an authenticated `turn/completed` event, so a flooding seat could satisfy DAG prerequisites without a completed turn.

## Change

- `codex_appserver.TurnResult.completed_observed`: True only when `_finish` processed a real `turn/completed` notification for the turn id; a cap hit mid-stream leaves it False.
- `agents.AgentResult.turn_completed`, propagated by `run_codex_appserver`.
- `run_transport.finish`: `preserved` additionally requires `result.turn_completed`. The truncated text stays as a diagnostic with `output_truncated` in `worker-results.json`; the seat stays un-poisoned and excluded from same-seat retry in both cases (the intent of #1144). DAG prerequisites key off `WorkerResult.ok`, so a truncated, uncompleted result no longer satisfies them.
- CHANGELOG: the #1144 entry reworded to say salvage is successful only with an observed completion event.
- Tests: `test_over_cap_worker_result_with_completed_turn_stays_ok`, `test_over_cap_worker_result_without_completed_turn_is_failure` (replaces #1144's test that asserted the buggy `ok=True`), `test_dag_dependent_skipped_when_only_coverer_is_truncated_uncompleted_result`, `test_over_cap_after_turn_completed_observed_keeps_completion_signal`.

## Verification

Worker (ox_worker, brigade run `8e846886.20260826-031317-802e743a`): ruff check, ruff format, and the transport/aboyeur/envelope/receipt tests at exit 0; also `test_codex_appserver.py` and `test_agents.py` (183 passed).
Conductor re-check through Brigade: same three gates, `+1` each, 314 passed. Full `./scripts/verify` runs in CI. A daybreak review is dispatched against this head.

## Provenance

Implemented by the `ox_worker` seat under `brigade run` (hub-tracked as `conductor-1200`); committed, rebased, and filed by the conductor.


solomonneas:
Daybreak review against 630b832d (brigade run `8e846886.20260826-032802-04feebdf`, read-only sandbox). Relayed by the conductor; converting to draft and dispatching a verify-then-fix lane for the two findings.

BLUF: NOT READY. Very likely, confidence High - two fail-closed defects remain.

- **Medium - completion success path is unreachable:** [codex_appserver.py:308](src/brigade/codex_appserver.py:308) rejects over-cap JSON before routing it. If completion was queued before a later overflow, lines 480-481 discard it. Thus the positive test at [test_codex_appserver.py:309](tests/test_codex_appserver.py:309) bypasses real ingestion, and `completed_observed=True` at line 629 cannot occur through the production stream. Fix: preserve matching completion metadata during bounded ingestion, without retaining oversized text, and add an end-to-end ingestion test.

- **Medium - failed terminal events can qualify as success:** [codex_appserver.py:621](src/brigade/codex_appserver.py:621) sets `completed_observed=True` before checking `turn.status`. A matching `turn/completed` with status `failed` or `interrupted` and oversized text can reach [run_transport.py:873](src/brigade/run_transport.py:873) as `ok=True`, satisfying DAG prerequisites. Fix: require `status == "completed"` for the salvage signal and test failed/interrupted statuses.

Clean: exact turn-ID binding rejects other-turn notifications; post-cap routing fails closed; state is per-turn; resumed over-cap runs remain `ok=False`; app-server overflow returns before same-seat retry; failed DAG prerequisites become explicit `skipped` results.

NOT READY

solomonneas:
Sendback outcome (ox_worker, brigade run `8e846886.20260826-033252-51f65a7c`): both findings reproduced with failing tests first, fixed in 5eeae7f7.

1. Completion unreachable through real ingestion: bounded ingestion now records `(thread_id, turn_id) -> status` for every observed `turn/completed`, including records that overflow the shared budget after parsing and oversized raw records (metadata via a bounded marker and brace-window byte scan, never the text; `_note_overflow_metadata`). `AppServer.observed_completion_status()` feeds `CodexThread._output_limit_result(..., turn_id)`. End-to-end test feeds `_read_loop` an over-cap `turn/completed` and asserts `completed_observed`; a second cap-trip site in `_read_loop`'s leftover-overflow branch was found and wired.
2. Status gate: the over-cap branch sets `completed_observed = (status == "completed")`; failed and interrupted stay `ok=False`. Tests: `test_over_cap_completion_requires_completed_status[failed|interrupted]`.

Conductor verification through Brigade: ruff check and format (`+1`, `+1`); `pytest -q` over app-server, transport, agents, aboyeur: 483 passed (`+1`). Re-dispatching daybreak against 5eeae7f7.

solomonneas:
Daybreak round-2 review against 5eeae7f7 (brigade run `8e846886.20260826-035252-8bb1585c`). Relayed by the conductor. Finding 1 is a functional regression (double byte charge) and must be fixed regardless; round 3 dispatched for all three.

BLUF: NOT READY. Almost certainly, confidence High - three confirmed blockers exist in the reviewed diff.

### Findings

1. **HIGH - Parsed records are charged twice**
   - **Where:** `src/brigade/codex_appserver.py:405,412`
   - **Scenario:** Every valid parsed record consumes twice its byte length. Normal output between half the cap and the full cap now fails as over-limit.
   - **Fix:** Remove the second `_charge_record` call. Add a test asserting one record increases `used` exactly once and near-cap normal traffic succeeds.

2. **HIGH - Oversized-record scan can misattribute completion**
   - **Where:** `src/brigade/codex_appserver.py:51-94,358-363`
   - **Scenario:** The scan accepts `"turn/completed"` anywhere, selects the first `"turn"` occurrence, ignores braces inside strings, and accepts the first nested or duplicate `id`, `status`, and `threadId`. A malformed or attacker-influenced structured record can record `completed` for another turn. Duplicate `status: completed` followed by `status: failed` also yields a false success signal, which downstream can promote to `ok=True`.
   - **Fix:** Use a bounded structural JSON tokenizer that validates exact paths: top-level `method`, `params.threadId`, and direct `params.turn.id/status`. Reject duplicate keys and incomplete structures. Add marker-in-string, nested-object, brace-in-string, duplicate-key, and wrong-thread tests.

3. **MEDIUM - Completion status storage is unbounded and stale**
   - **Where:** `src/brigade/codex_appserver.py:188,293-307`
   - **Scenario:** `_observed_completions` retains every turn indefinitely. `reset_capture()` neither clears nor consumes entries, so long-lived threads accumulate metadata and reused turn IDs can inherit stale completion state.
   - **Fix:** Keep one active completion entry per thread, clear it before each turn, and consume it after result construction.

### Clean

- Salvage now requires status exactly `"completed"` at `src/brigade/codex_appserver.py:675-691`.
- The stream reader stops at `MAX_CAPTURE_BYTES + 1`; oversized data is not fully read or passed to `json.loads`.
- The brace walk itself is capped at 65,536 bytes.
- Existing tests cover completed versus failed status and bounded stream reads, but not the adversarial or single-charge cases above.
- No files were changed and no tests were run under the report-only constraint.

NOT READY

solomonneas:
Round-3 outcome (ox_deep, brigade run `8e846886.20260826-035725-87655d69`): all three round-2 findings reproduced with failing tests first, fixed in 51bd7ffb (rebased onto current main).

1. High, double charge: the second `_charge_record` call in `_ingest_line` is gone; each record charges exactly once. Tests: `test_ingest_line_charges_parsed_record_once`, `test_near_cap_normal_records_ingest_without_overflow`.
2. High, forgeable metadata scan: replaced by `_BoundedRecordScanner`, a structural byte-level tokenizer validating exact paths only (`method == "turn/completed"`, `params.threadId`, `params.turn.id`, `params.turn.status`), rejecting duplicate keys wherever it walks, string-aware for braces and escapes, bounded to the 64 KiB window, never materializing text; input ending inside a value is tolerated only after all four path fields are captured. Eight regression tests (marker in string, nested decoy, brace in value, duplicate status and method keys, wrong path, missing threadId, truncated record, window bound).
3. Medium, unbounded retention: one `(turn_id, status)` entry per thread, cleared in `reset_capture`, consumed by `consume_observed_completion` after `_output_limit_result`. Tests: `test_reused_turn_id_cannot_inherit_stale_completion`, `test_output_limit_result_consumes_observed_completion`.

Conductor verification through Brigade: ruff check and format (`+1`, `+1`); `pytest -q` over app-server, transport, agents: 219 passed (`+1`). Re-dispatching daybreak against 51bd7ffb.

solomonneas:
Daybreak round-4 review against 51bd7ffb (brigade run `8e846886.20260826-042355-108f6eb1`). Relayed by the conductor.

Conductor disposition: the bounded byte tokenizer is removed rather than patched again. An oversized raw record never signals completion (fail closed); completion is honored only for records that `json.loads` parsed within the cap, including the post-parse budget-overflow case; the completion map is keyed by `(thread_id, turn_id)` with an atomic read-and-remove under the lock. Fix lane dispatched.

BLUF: Not ready. Almost certainly, confidence High - read-only probes reproduced two defects.

### [HIGH] Malformed JSON can forge completion

- **Where:** [codex_appserver.py:135](src/brigade/codex_appserver.py:135), [codex_appserver.py:306](src/brigade/codex_appserver.py:306)
- **Scenario:** The scanner accepted a second top-level object, invalid UTF-8, invalid `01` number syntax, and missing closing delimiters after `"status":"completed"`. These can set `completed_observed`, allowing an over-cap worker result to be preserved as successful.
- **Fix:** Use strict UTF-8 and JSON-number validation, reject trailing non-whitespace after the root object, and permit truncation only while scanning a non-target value after all fields were captured. Add regressions for each case.

### [MEDIUM] Interleaved turns lose completion state

- **Where:** [codex_appserver.py:528](src/brigade/codex_appserver.py:528), [codex_appserver.py:551](src/brigade/codex_appserver.py:551), [codex_appserver.py:929](src/brigade/codex_appserver.py:929)
- **Scenario:** Turn B’s reset can erase turn A’s completion. A completion arriving between A’s lookup and thread-wide consume is also deleted. Both were reproduced.
- **Fix:** Key entries by `(thread_id, turn_id)` and atomically read-and-remove the exact entry under the lock, or serialize turns per thread.

Clean: escapes including `\u` and escaped quotes, nested arrays, keys preceding `method`, duplicate-key rejection, 64 KiB bounding, mid-string status truncation rejection, and single charging of nonempty malformed records.

NOT READY

solomonneas:
Round-4 outcome (ox_deep, brigade run `8e846886.20260826-043148-c42f220b`): both findings reproduced with failing tests first, fixed in f891dcc8 (rebased onto current main), net −356 lines.

1. High, forgeable metadata scan: `_BoundedRecordScanner`, `_salvage_completion_metadata`, and their helpers are removed along with both call sites; an oversized raw record only charges the budget and trips the cap. Completion is recorded only for records `json.loads` parsed, keeping the round-2 post-parse overflow path. Tests: trailing second object, invalid UTF-8, malformed number, missing closing delimiters, and a well-formed oversized completed record all leave `completed_observed` false.
2. Medium, interleaved turns: `_observed_completions` keyed by `(thread_id, turn_id)`, `consume_observed_completion` is an atomic read-and-remove under `_state_lock`, `reset_capture` clears only that thread, bounded at 256 with FIFO eviction. Tests: interleaved turns both retain completions, a sibling completion survives the other turn's consume, the map stays bounded.

Conductor verification through Brigade: ruff check and format (`+1`, `+1`); `pytest -q` over app-server, transport, agents: 219 passed (`+1`). Re-dispatching daybreak against f891dcc8.

solomonneas:
Daybreak round-5 review against f891dcc8 (brigade run `8e846886.20260826-045047-daafaef5`). Relayed by the conductor. Down to one ordering race that fails closed (false negative, never a false positive); fix lane dispatched.

BLUF: NOT READY. Almost certainly, confidence High - a concurrency race can lose a valid parsed completion.

### [MEDIUM] Record completion before signaling overflow

- **Where:** `src/brigade/codex_appserver.py:339-344,374-377,641-655`
- **Scenario:** `_charge_record()` detects cumulative overflow and calls `_signal_output_limit()` before `_note_turn_completed()`. A waiting turn can wake, consume the exact-turn entry while it is still absent, and return `completed_observed=False`. The completion is recorded afterward, too late.
- **Fix:** Record parsed completion metadata before publishing `_output_limit_exceeded` or queueing `_OUTPUT_LIMIT`. Add a deterministic ordering regression test. The current test at `tests/test_codex_appserver.py:376-420` completes ingestion first and reinserts completion synchronously, so it does not exercise this race.

### Clean

- Unparsed records cannot create completion state. Oversized records stop before `json.loads`; malformed and non-object records only charge the budget (`347-380`).
- The 256-entry FIFO can evict a live completion after 256 newer distinct keys (`264-275`). This is fail-closed: exact-key consumption returns `None`, yielding `completed_observed=False`. It can cause a false negative, never a false positive.
- Raw record buffering remains capped at `MAX_CAPTURE_BYTES + 1`; post-parse overflow terminates ingestion (`389-413`).
- Each JSONL record is charged once through `_charge_record`; the completion bookkeeping adds no second record charge (`339-380`).
- `_finish` still requires `status == "completed"` before setting the over-cap completion signal (`658-685`).

NOT READY

solomonneas:
Round-5 outcome (ox_worker, brigade run `8e846886.20260826-045413-0222f925`): the ordering race reproduced (`test_over_cap_completed_record_is_noted_before_limit_signal` failed with the completion absent at signal time) and is fixed in 837bce01: on budget overflow of a parsed record, `_ingest_line` checks the budget directly and calls `_note_turn_completed` before `_signal_output_limit`. A `_signal_output_limit` spy asserts the entry already exists at signal time. Clean invariants untouched.

Conductor verification through Brigade: ruff check and format (`+1`, `+1`); `pytest -q` over app-server, transport, agents: 220 passed (`+1`). Re-dispatching daybreak against 837bce01; flips to ready on READY.

solomonneas:
Daybreak round-6 review against 837bce01 (brigade run `8e846886.20260826-045823-d12d5ee7`): READY. Relayed by the conductor; marking ready for review, merges when CI is green. Six review rounds in total; the findings trail above is complete.

BLUF: No findings. READY. Very likely, confidence Moderate - all requested paths are correct; runtime verification was blocked by the read-only sandbox.

- Post-parse overflow records completion before signaling: [codex_appserver.py:374](src/brigade/codex_appserver.py:374).
- Reader leftover overflow remains unparsed and cannot spoof completion: [codex_appserver.py:401](src/brigade/codex_appserver.py:401).
- `_finish` requires `status == "completed"`: [codex_appserver.py:678](src/brigade/codex_appserver.py:678).
- Direct `try_add` charges once, covered at [test_codex_appserver.py:735](tests/test_codex_appserver.py:735).
- Round-4/5 controls remain clean: malformed/oversized rejection, status gating, per-turn isolation, bounded state, atomic consumption, and `ok=False`.

READY

PR #1194, opened 2026-08-26
fix: restore verification throughput locally and in CI

solomonneas:
## Summary

- add a focused development gate and serialize full verification across linked worktrees
- remove fixed subprocess polling latency and isolate host-dependent test discovery
- shard Python CI across four workers per version, combine coverage once, cache pip downloads, and cancel superseded runs
- require Brigade receipt capture for both focused and full verification commands

## Why

Agents were running the complete coverage suite after ordinary changes, and linked worktrees could start overlapping copies. CI also ran the full test collection serially for each Python version. This change gives development work a smaller gate while keeping the full coverage run at PR, merge, release, and verifier-infrastructure boundaries.

## Verification

- `20260826-020836-work-verify-6beac2`: 9,055 passed, 8 skipped, 68 subtests passed, 83.71% coverage
- `20260826-012943-work-verify-8b2400`: `actionlint .github/workflows/ci.yml` passed
- `20260826-013755-work-verify-d04410`: 114 focused hook tests passed
- independent reviews found receipt-enforcement, path-matching, and release-note gaps. All are fixed, with regression coverage for the hook behavior

## Notes

- protected check names remain `test (3.10)`, `test (3.11)`, and `test (3.12)`
- the 78% coverage threshold runs once after combining Python 3.12 shard data
- the first GitHub Actions run will provide the production CI wall-time measurement

## Summary by CodeRabbit

* **New Features**
  * Added focused verification for selected tests, linting, formatting, type checks, and consistency checks.
  * Added deterministic parallel test execution across supported Python versions.
  * Added coverage aggregation with a required 78% threshold.
  * Added safeguards to prevent overlapping verification runs.

* **Bug Fixes**
  * Improved process output collection and termination responsiveness.
  * Expanded detection of verification commands across supported path formats.

* **Documentation**
  * Updated development guidance for focused and full verification workflows.

PR #1207, opened 2026-08-26
fix(scanners): bound scanner runs and inputs, reject work from failed runs

### LINKED ISSUE #1197: scanners: bound subprocess capture and own the process group for scanner runs
From the #1015 second-pass audit (part 1, main 06e61919). Severity: Medium.

`src/brigade/work_cmd/scanners.py` runs scanners with `subprocess.run(..., capture_output=True)` (around L1836-1858) with no byte cap on stdout/stderr and no process-group ownership. A scanner can exhaust memory by streaming unbounded output, and a descendant that keeps the pipe descriptors open can outlive the timeout, which also feeds the #1006-1 same-UID concern (a background child retaining access to verifier state after return).

Fix: replace with bounded streaming capture (the delimited/streaming helpers in `proc.py` from #1178 fit) and terminate the whole process group on overflow or timeout, the way #1190 (PR #1195) does for runbook steps.

Refs #1015, #1006.

### LINKED ISSUE #1198: scanners: enforce byte limits on scanner-controlled inbox, import JSONL, and snapshot reads
From the #1015 second-pass audit (part 1, main 06e61919). Severity: Medium.

In `src/brigade/work_cmd/scanners.py`, inbox reads use unrestricted `read()` (around L1255-1274), import JSONL ingestion accumulates every chunk then joins and decodes everything (around L1708-1717), and existing run artifacts are snapshotted for rollback without a limit (around L417-445). A scanner can create a multi-GB file and force multiple full allocations.

Fix: enforce a total and a per-record byte limit before parsing or rollback snapshotting, rejecting oversized input with a typed failure instead of loading it.

Refs #1015, #1006.

### LINKED ISSUE #1199: scanners: failed scanner runs can still create eligible imports through _scanner_stamp_new_imports
From the #1015 second-pass audit (part 1, main 06e61919). Severity: Medium.

Run proof correctly requires a successful run (`src/brigade/work_cmd/scanners.py` around L590-600), but `_scanner_stamp_new_imports` (around L1452-1523) accepts and rebuilds ordinary rows without that proof and is invoked regardless of run status (around L2658-2685). A scanner can append a valid pending import, exit nonzero, and still produce eligible work. The file-based `--ingest-output` path already skips failed runs (around L2691-2705); the inline path does not.

Fix: on a failed scanner run, restore the pre-run inbox and reject all new rows; stamp only rows covered by a successful run proof.

Refs #1015, #1006.

solomonneas:
Fixes #1197. Fixes #1198. Fixes #1199. From the #1015 second-pass audit (part 1).

## #1197: bounded scanner runs with process-group ownership

`subprocess.run(capture_output=True)` had no byte cap and no process-group ownership, and a pipe-holding descendant blocked `communicate()` past the child's exit. Scanner runs now go through `brigade.proc.run` (bounded streaming capture, process-group kwargs, group reap on overflow or timeout) and record a typed failure (`output_limit_exceeded: true`, error naming the 1 MiB cap). `proc._collect_process_output` stops draining about 0.5 s after the direct child exits and reaps remaining group members instead of burning the timeout. Tests: `test_scanner_run_bounded_output_reaps_child_and_records_typed_failure`, `test_scanner_run_does_not_wait_for_descendant_holding_pipes`.

## #1198: byte limits on scanner-controlled reads

Unrestricted `handle.read()` in `_scanner_inbox_bytes`, unbounded chunk-join in `_scanner_read_import_jsonl`, and full-file rollback snapshots in `_write_scanner_run_file`. Now: 4 MiB total and 256 KiB per-record caps enforced before parsing, a 4 MiB snapshot cap that fails closed without touching the unsnapshotted artifact, and typed failures `ScannerInputLimitExceeded` / `ScannerSnapshotLimitExceeded` (`OSError` subclasses so existing fail-closed paths surface them). Descriptor-relative no-follow validation untouched. Five regression tests.

## #1199: failed runs create no eligible work

`_scanner_stamp_new_imports` accepted and rebuilt new rows regardless of run status. On a failed, timed-out, or nonzero run it now restores the exact pre-run inbox bytes via the retained-snapshot path and rejects all appended rows with `self_import.rejection_reasons.run_not_completed`; stamping still requires the exact in-memory successful-run proof. Tests: `test_failed_scanner_run_restores_pre_run_inbox_and_rejects_new_rows`, `test_failed_scanner_run_leaves_no_eligible_work_end_to_end`.

Note: `tests/test_authority_broker.py` asserted the literal `"capture_output"` in the scanners source as a proxy for "child output stays captured"; it now asserts `"proc_mod.run("` (bounded capture). The secret/env assertions and the captured-output property are unchanged.

## Verification

Worker (ox_worker on shadowfax, brigade run `8d27fc81.20260826-031243-9d39e007`): ruff check, ruff format, and eight scanner/proc/authority test files at exit 0; mypy clean; proc-adjacent suites 274 passed.
Conductor re-check through Brigade on rocinante after applying the diff onto current main: same gates, `+1` each, 341 passed. Full `./scripts/verify` runs in CI. A daybreak review is dispatched against this head. May need a small `proc.py` merge once #1195 lands.

## Provenance

Implemented by the `ox_worker` seat under `brigade run` on shadowfax (hub-tracked as `conductor-1197`); diff applied, verified, committed, and filed by the conductor.


solomonneas:
Daybreak review against 72f7a54b (brigade run `1334e217.20260826-033718-7a0152ba`, read-only sandbox). Relayed by the conductor. Finding 2 is a data-loss regression introduced by this branch, so this converts to draft now; a verify-then-fix lane is dispatched for all three.

BLUF: NOT READY. Very likely, confidence High - three confirmed guardrail gaps remain.

### Findings

1. **HIGH - Forced descendant reap is reported as success**
   - **Where:** [proc.py:525](src/brigade/proc.py:525), [proc.py:546](src/brigade/proc.py:546), [scanners.py:1962](src/brigade/work_cmd/scanners.py:1962)
   - **Scenario:** After the direct child exits, live reader threads trigger the cutoff and process-group termination, but neither timeout nor incomplete-output state is recorded. The direct child’s zero exit code remains, so the scanner is marked completed and receives a run proof. Legitimate delayed output can be discarded. Other `proc.run` callers can also accept partial nonempty output as successful at [agents.py:1354](src/brigade/agents.py:1354).
   - **Windows:** `taskkill` targets the already-exited parent PID at [proc.py:133](src/brigade/proc.py:133); `CREATE_NEW_PROCESS_GROUP` provides no kill-on-close guarantee, so a descendant can survive.
   - **Fix:** Record a typed incomplete-process-group failure whenever the cutoff fires, force a nonzero result, and use a Windows Job Object or equivalent owned-tree primitive.

2. **HIGH - Failed-run rollback can use a fabricated empty baseline**
   - **Where:** [scanners.py:2729](src/brigade/work_cmd/scanners.py:2729), [scanners.py:1477](src/brigade/work_cmd/scanners.py:1477)
   - **Scenario:** Any pre-run snapshot `OSError`, including the new 4 MiB limit, is swallowed and replaced with `before_raw=b""`. If the scanner then truncates the inbox below the cap and fails, rollback atomically publishes that empty value, destroying the pre-run inbox. A missing inbox is also restored as an existing empty file.
   - **Stability:** `_scanner_inbox_bytes` performs no before/after content metadata check, so a same-UID writer can produce a torn snapshot during the read.
   - **Fix:** Abort before launching when the snapshot cannot be captured. Return raw bytes, parsed rows, existence, and stable before/after `fstat` metadata from one bounded read.

3. **MEDIUM - Rollback snapshots still bypass the memory cap**
   - **Where:** [scanners.py:1190](src/brigade/work_cmd/scanners.py:1190), [ledger.py:3427](src/brigade/work_cmd/ledger.py:3427)
   - **Scenario:** Both helpers append chunks without a byte counter. A same-UID writer can grow the inbox after a bounded guard read but before either snapshot, causing unbounded allocation or an indefinitely growing read.
   - **Fix:** Enforce the cap while reading each descriptor and raise the typed limit exception. Do not depend on an earlier size/read check.

### Clean

- Normal failure, timeout, and output-overflow paths do not register a successful run proof.
- Symlinked leaves and parent swaps are rejected with no-follow, held-descriptor validation.
- Snapshot publication itself uses fsync plus atomic replacement.
- Runbooks and Codex app-server do not use `proc.run`, so this change does not alter them.
- New tests cover basic caps and failed-row rejection, but not concurrent growth, snapshot-read failure, partial-success classification, or Windows descendant survival.
- No tests were run due the report-only, read-only constraint. Worktree remains clean at `72f7a54b`.

NOT READY

solomonneas:
Sendback outcome (ox_deep, brigade run `1334e217.20260826-034245-31c4803f`): all three findings reproduced with failing tests first, fixed in c34264af (rebased onto current main).

1. High, cutoff reported as success: `Result.incomplete_process_group` with a forced nonzero exit (125) from `_collect_process_output`; the scanner receipt is marked failed with a typed error so `_register_scanner_run_proof` never fires; `run_agent` returns `failure_kind="incomplete-process-group"`. Windows: an owned kill-on-close Job Object (`_WindowsChildJob`) assigned from the child handle at launch, `_stop_child` prefers `job.terminate()`, taskkill only as a fallback. The earlier `test_scanner_run_does_not_wait_for_descendant_holding_pipes` asserted `status == "completed"`, which encoded the defect; rewritten.
2. High, fabricated empty baseline: `_snapshot_scanner_inbox` is one bounded read returning raw bytes, parsed rows, existence, and before/after fstat identity (dev, ino, mode, nlink, size, mtime_ns); `_scanners_run_payload` aborts before launch on any snapshot OSError (rc 2); rollback refuses an inode mismatch and restores missing-as-missing. Five regression tests.
3. Medium, uncapped snapshot helpers: byte counter inside each chunk loop, `ScannerInputLimitExceeded` in scanners and `ImportInboxSnapshotLimitExceeded` in the ledger snapshot helper.

Conductor verification through Brigade: ruff check and format (`+1`, `+1`); seven-file pytest: 517 passed (`+1`). Re-dispatching daybreak against c34264af; the Windows Job Object path is also exercised by CI's windows-native-acceptance.

solomonneas:
Daybreak round-2 review against c34264af (brigade run `1334e217.20260826-040800-8dc3fbf1`). Relayed by the conductor; round-1 items confirmed closed except the two containment races below. Round 2 fix lane dispatched.

BLUF: NOT READY. Very likely, confidence High - two containment races remain.

## Findings

### HIGH - Windows descendants can escape the Job Object

- **Where:** `src/brigade/proc.py:160-241, 861-878`
- **Scenario:** `Popen` starts the child before the Job Object is created and assigned. The child can spawn a grandchild during that interval. Creation or assignment failure silently returns `None`, restoring the unreliable `taskkill` fallback.
- **Fix:** Create the Job Object first, launch suspended, assign, then resume. Treat creation or assignment failure as a typed launch failure and terminate the suspended child.
- **Test gap:** `tests/test_proc.py:737-780` mocks an already-created process and only tests successful assignment.

### HIGH - Snapshot identity is not bound through launch

- **Where:** `src/brigade/work_cmd/scanners.py:1343-1380, 1416-1456, 2856-2874`
- **Scenario:** The descriptor is closed after snapshotting. A same-UID writer can replace or modify the inbox before scanner launch. On success, substituted rows may be treated as scanner-created. On failure, same-inode concurrent writes can be overwritten because rollback checks only device and inode.
- **Fix:** Hold a run-wide inbox lock from snapshot through stamping or rollback, require all inbox writers to honor it, and revalidate full metadata under that lock before launch.
- **Test gap:** Tests cover mutation during reading and inode replacement during rollback, but not replacement between snapshot and launch.

### MEDIUM - Cutoff cleanup exceptions can strand a running receipt

- **Where:** `src/brigade/proc.py:662-702`; `src/brigade/work_cmd/scanners.py:1997, 2096-2100, 2832-2835`
- **Scenario:** After `incomplete_group=True`, a Job termination or wait exception can bypass `Result` construction. The scanner releases descriptors but leaves the persisted receipt as `running`, blocking later runs.
- **Fix:** Convert cleanup failures into a typed nonzero result and finalize the scanner receipt as failed in its exception path.

### MEDIUM - Ledger snapshot can miss same-inode mutation

- **Where:** `src/brigade/work_cmd/ledger.py:3449-3467`
- **Scenario:** Its before/after comparison omits size, mtime, and the bytes-read length. An in-place concurrent rewrite can produce a torn rollback snapshot.
- **Fix:** Compare size and `mtime_ns`, and require `len(snapshot) == before.st_size`.

## Clean

- No successful `proc.run` return loses `incomplete_process_group`; it is copied into reconstructed results at `proc.py:716-726`.
- Pre-launch snapshot failure creates no run directory, receipt, or authority: snapshot occurs at `scanners.py:2856`, before `_scanner_run_one` at line 2874.
- Missing inbox restoration and byte counters inside read loops are correct.
- `agents.py:1156-1172` maps cutoff to a typed harness failure. Harnesses that daemonize descendants inheriting output pipes will now fail after 0.5 seconds; they must keep the parent alive or redirect descendant output.
VERDICT: NOT READY

solomonneas:
Round-2 outcome (ox_deep, brigade run `1334e217.20260826-041313-fd212b72`): all four findings reproduced with failing tests first, fixed in 0e8b6dea (rebased onto current main).

1. High, Job Object race: the job is created before `Popen`; on Windows the child launches with `CREATE_NEW_PROCESS_GROUP | CREATE_SUSPENDED`, is assigned while frozen, then resumed; creation, assignment, or resume failure is a typed nonzero result (126) that kills the suspended child, with no taskkill fallback. Tests: `test_windows_run_without_job_object_fails_closed_before_launch`, `test_windows_run_terminates_suspended_child_when_assignment_fails`.
2. High, snapshot-to-launch TOCTOU: a run-wide `_scanner_inbox_run_lock` (`inbox.jsonl.lock` beside the inbox, flock/msvcrt, reentrant per process) is held from snapshot through stamping or rollback; every inbox writer honours it; `_revalidate_scanner_inbox_snapshot` compares dev, ino, size, mtime_ns, and bytes-read length under the lock immediately before launch, aborting rc 2. Tests: replacement and same-inode write between snapshot and launch abort; lock reentrancy and cross-process blocking. Disclosed: an attempted extra refusal of same-inode drift at rollback broke the round-1 contract test that a failed run restores over the child's own same-inode writes, and was reverted; the lock plus pre-launch revalidation covers the gap, residual documented.
3. Medium, cleanup exceptions: post-loop `_stop_child` calls run under `bounded_stop()`, exceptions become a typed nonzero result with `incomplete_process_group=True`; `_scanner_run_one`'s except path finalizes the receipt as failed before releasing authority. Tests: `test_cleanup_failure_becomes_typed_nonzero_result`, `test_run_one_exception_finalizes_scanner_receipt_as_failed`.
4. Medium, ledger snapshot compare: includes `st_size`, `st_mtime_ns`, and `len(raw) == before.st_size`. Test: `test_ledger_import_inbox_snapshot_detects_mid_read_append`.

Conductor verification through Brigade: ruff check and format (`+1`, `+1`); seven-file pytest: 527 passed (`+1`). Re-dispatching daybreak against 0e8b6dea.

solomonneas:
Daybreak round-3 review against 0e8b6dea (brigade run `1334e217.20260826-045229-6f1f5311`). Relayed by the conductor. Round-2 items confirmed closed; the lock introduced in round 2 has to cover the ledger's own inbox writers and be opened safely. Round-3 fix lane dispatched.

BLUF: The PR is very likely not ready, confidence High. Three integrity defects and one bounded availability defect remain.

Findings:

- HIGH, `src/brigade/work_cmd/ledger.py:541-554,4147-4259,4734`: canonical `_write_imports` and `_append_import_records` writers never acquire `_scanner_inbox_run_lock`. A concurrent Brigade import, promote, or dismiss can race scanner stamping or rollback, losing or misattributing trusted rows. Move the lock into a shared module and acquire it around every inbox read-modify-write transaction.

- HIGH, `src/brigade/work_cmd/scanners.py:1409-1447`: the workspace lock uses a path-following `os.open` without `O_NOFOLLOW`, nonblocking open, or regular-file/single-link validation. A scanner can plant a symlink/FIFO for the next run or replace the locked inode during this run. Later writers can lock a different inode and bypass exclusion. Use an externally verifier-owned lock, or descriptor-relative no-follow opening with persistent inode authority and replacement detection.

- MEDIUM, `src/brigade/work_cmd/ledger.py:474-497`: `_read_import_inbox_raw` remains unbounded and compares only device/inode/mode/link count. Concurrent growth can consume unbounded memory; an in-place mutation can produce a torn read. The new test at `tests/test_work_cmd_scanners.py:4155-4190` covers `_snapshot_import_inbox`, not this reviewed-snapshot reader. Apply the same byte cap, size/mtime comparison, and `len(raw) == st_size` check here.

- MEDIUM, `src/brigade/proc.py:282-294`: `OpenThread` failure correctly returns false and triggers child termination, but `ResumeThread == 0xFFFFFFFF` breaks the loop and then returns true. A suspended child can hold the inbox lock until timeout. Return false on that value and add direct OpenThread/ResumeThread failure tests.

Clean:

- Job creation precedes suspended `Popen`; assignment precedes resume.
- Pre-launch aborts release the lock through `finally`.
- `msvcrt` lock/unlock uses the same byte at offset zero for callers using this helper.
- Cleanup exceptions become typed results, and receipts are finalized.
- Scanner snapshot revalidation covers identity, size, mtime, length, and bytes.
NOT READY

solomonneas:
Round-3 fixes landed at 825e58bc (Ox deep lane, conductor-1197, hub-tracked). All four round-2 findings reproduced with a failing test first:

- HIGH canonical inbox writers never took the inbox run lock: the lock now lives in work_cmd/inbox_lock.py (reentrant per process) and _write_imports, _append_import_records, _backfill_import_provenance, and _promote_matching_imports hold it across their whole read-modify-write (task-ledger lock before inbox lock, fixed order). Root discovery: self-importing scanner children deadlocked under the new lock, so the launcher passes its held descriptor through BRIGADE_INBOX_LOCK_FD with a new proc.run(pass_fds=...) and children adopt it (nonblocking flock on the shared open file description). Windows cannot pass the descriptor; documented residual.
- HIGH lock file opened path-following: POSIX acquisition walks a held no-follow dirfd, opens O_NOFOLLOW|O_NONBLOCK, requires a regular file with nlink == 1 by fstat, and verify_inbox_lock() re-checks dev/ino against the held descriptor before every protected write in both modules.
- MEDIUM _read_import_inbox_raw unbounded with a weak compare: same 4 MiB cap, size+mtime_ns compare, and len(raw) == st_size check as the snapshot reader.
- MEDIUM ResumeThread (DWORD)-1 reported success: proc.py returns False on 0xFFFFFFFF and the typed cleanup terminates the still-suspended child.

Verification (Brigade-captured): ruff check, ruff format --check, mypy, and the 88 test suites that import proc, work_cmd, scanners, or the inbox lock: 3343 passed, 2 skipped, 2 failed in tests/test_care_cmd.py crontab syntax checks that fail identically on a worktree without this change (crontab -n cannot mkstemp in /var/spool/cron on the conductor host; environmental, passes in CI). Daybreak round 4 next, with an explicit probe of the inherited-fd adoption path.

solomonneas:
Daybreak round 4 (read-only, hub-tracked run in conductor-1197) on 825e58bc: NOT READY, 4 findings, all against the round-3 fd-inheritance design, which I am having removed rather than patched:

- HIGH a scanner child or any descendant sharing the launcher's open file description can flock(LOCK_UN) the run-wide lock and let outside writers into the stamping/rollback window (inbox_lock.py:197, scanners.py:2112).
- HIGH two self-importing children adopting the same open file description both pass their nonblocking flock, so siblings have no mutual exclusion (scanners.py:2121).
- MEDIUM an inheriting child with ambient inbox ownership that takes the task-ledger lock inverts the task-then-inbox order and can strand the task lock for the run window (ledger.py:4755, scanners.py:2942).
- MEDIUM on Windows a hostile parent can hand over an unlocked descriptor to the real lock file and the msvcrt branch never acquires (inbox_lock.py:162-203).

Round-5 lane dispatched with a two-lock design: the run lock stays launcher-owned and is never passed to children; a new writer lock, opened independently by each process, serializes every canonical read-modify-write (launcher sections included); a capability-free environment marker tells a child to skip the run lock and take the writer lock; order is task, run, writer everywhere; Windows uses validated acquisition only, no adoption; BRIGADE_INBOX_LOCK_FD and pass_fds plumbing are deleted. Cleared by the reviewer: closed/malformed/different-file/FIFO/multi-link inherited fds fall back safely, closing the child fd does not release the parent lock, all canonical writer entry points revalidate; low-level rollback helpers (ledger.py:3403,3520; scanners.py:1226,1250) rely on the caller's check with the same-UID race as the documented residual.

solomonneas:
Round-5 redesign landed at 92e29a3f (Ox deep lane, conductor-1197, hub-tracked; the lane hit its 2 h budget during an extra sweep after the work was complete, so the conductor verified and landed its terminal state).

- Inherited lock descriptors are gone: BRIGADE_INBOX_LOCK_FD, child_lock_env, child_lock_pass_fds, _adopt_inherited_lock, and proc.run(pass_fds=...) removed (test_run_exposes_no_descriptor_passing_plumbing, test_scanner_children_receive_capability_free_marker_and_no_descriptors).
- Two locks: the run lock (inbox.jsonl.lock) is launcher-owned for the whole window; a new writer lock (inbox.jsonl.writer.lock), opened and validated independently by each process, serializes every canonical read-modify-write including the launcher's stamping and rollback sections (test_launcher_stamping_waits_for_child_writer_lock, test_two_concurrent_self_importing_children_serialize_on_writer_lock).
- A child that unlocks anything it inherits cannot release the launcher's run lock; an outside writer blocks under the run lock and proceeds after release (test_unlocking_any_inherited_descriptor_cannot_release_launcher_run_lock, test_outside_writer_blocks_under_run_lock_then_proceeds_after_release).
- BRIGADE_SCANNER_RUN_ID is a capability-free marker: forging it takes only the writer lock, stripping it waits on the run lock (test_scanner_run_marker_is_capability_free_and_writer_lock_is_adjacent).
- Lock order task ledger, run, writer everywhere, asserted across all canonical writers; a child holding the writer lock then taking the task lock neither deadlocks nor fails (test_every_canonical_writer_follows_task_then_run_then_writer_order, test_child_writer_lock_then_task_ledger_lock_does_not_deadlock_or_fail).
- Windows: validated acquisition only, no adoption (test_inbox_lock_module_has_no_bare_windows_open_flags).

Verification (Brigade-captured): ruff check, ruff format --check, mypy, scanners/imports/proc/inbox-windows suites 330 passed 1 skipped, ledger/integrity/runs-dir/sweeps/facade/session suites 176 passed. Daybreak round 5 next.

solomonneas:
Merged origin/main into the branch at 5bd14356 (merge commit rather than rebase: the conflict sat in the branch's first commit). Resolution in src/brigade/proc.py keeps #1194's wake-event completion loop (no 50 ms polling) and folds in this PR's descendant-drain cutoff: after the direct child exits, the wait is bounded by both the remaining timeout and the drain window, and incomplete_group is set when the window elapses. Verified through Brigade: ruff check, ruff format --check, mypy, tests/test_proc.py + test_subprocess_guards.py + test_work_cmd_scanners.py 209 passed. Daybreak round 5 next.

solomonneas:
Daybreak round 5 (read-only, hub-tracked run in conductor-1197) on 5bd14356: NOT READY, 3 MEDIUM findings; the redesign itself passed (no lost-wake or spin in the merged proc.run loop, lock order task/run/writer confirmed, marker cannot reach unrelated proc.run children, Windows validates before locking). Round-6 lane dispatched for:

- MEDIUM stale in-process registry: _held_inbox_lock() closes the outermost fd but leaves its _ACTIVE_LOCKS entry (inbox_lock.py:216), so verify_inbox_lock()/verify_inbox_writer_lock() pass after release and a marked child can write unexcluded. Fix: remove the entry atomically on outermost exit, verify under the module lock with an fstat of the live descriptor.
- MEDIUM reads before the writer lock: scanner stamping reads at scanners.py:1615 and locks only for publication at :1795; output ingestion captures its dedup/rollback snapshot at :3048-3069 before _append_import_records() locks, so a forced receipt failure rolls back a concurrent commit. Fix: take the writer lock before each canonical snapshot and hold it through publication or rollback capture.
- MEDIUM unbounded acquisition (inbox_lock.py:291, Windows :310-315): an escaped descendant holding writer.lock hangs launcher stamping and, via the run lock, every outside writer. Fix: deadline-bound acquisition with a typed failure, scanner run fails with bounded cleanup.

solomonneas:
Round-6 fixes landed at 33b90677 (Ox deep lane, conductor-1197, hub-tracked). All three round-5 findings reproduced with failing tests first:

- Stale registry: outermost release pops _ACTIVE_LOCKS under _PROCESS_LOCK before closing the fd; verify_inbox_lock/verify_inbox_writer_lock look up under the same lock and fstat the live descriptor (regular file, nlink 1, dev/ino equal to the held identity) before the path re-check (test_released_inbox_lock_registry_entry_is_removed_and_writers_refuse, test_nested_inbox_lock_entry_persists_until_outermost_exit, test_verify_inbox_lock_fstats_live_descriptor_and_refuses_closed_fd).
- Reads before the writer lock: _scanner_stamp_new_imports wraps read, publish, and rollback capture inside the writer lock; ingestion's dedup read, inbox snapshot, append, and publication-snapshot registration sit in one writer-lock span (test_stamping_takes_writer_lock_before_reading_and_keeps_child_rows, test_ingestion_pre_reads_under_writer_lock_and_concurrent_commit_survives_rollback).
- Unbounded acquisition: nonblocking flock / msvcrt LK_NBLCK retry bounded by DEFAULT_LOCK_DEADLINE_SECONDS = 30 (per-call deadline_seconds), typed InboxLockTimeout; sweep stamping marks the run failed with the reason and runs normal cleanup, ingestion records an inbox_lock_busy rejection, outside writers get the same typed failure (test_busy_writer_lock_deadline_bounds_launcher_stamping_and_releases_run_lock, test_outside_canonical_writer_gets_typed_failure_when_writer_lock_busy).

Verification (Brigade-captured): ruff check, ruff format --check, mypy, nine scanner/import/proc/ledger/sweeps/guards suites 443 passed 1 skipped. Daybreak round 6 next.

solomonneas:
Daybreak round 6 (read-only, hub-tracked run in conductor-1197) on 33b90677: NOT READY, 1 HIGH / 3 MEDIUM / 1 LOW. Round-7 lane dispatched for all five:

- HIGH escaped descendant outlives the run: a scanner forks a descendant holding writer.lock, closes the capture pipes, exits 0; collection completes without terminating the process group, so the descendant forces every scanner and outside writer into InboxLockTimeout indefinitely (proc.py:729-843, scanners.py:2154). Fix: scanner-mode supervision terminates the owned process group after the direct child exits regardless of pipe state.
- MEDIUM lock timeout during ingestion leaves status completed and rc 0 with every import rejected (scanners.py:3140-3195). Fix: run marked failed, nonzero rc.
- MEDIUM add, context, handoff/single promotion, both dismiss modes, and archive read the inbox before taking the run/writer locks (imports.py:58-62, :125-128, :847-890, :968-990, :1029-1050, :1060-1073; services.py:1081-1097). Fix: lock before the initial read and hold through publication, task ledger first where needed.
- MEDIUM ingestion drops the writer lock between the publication snapshot and receipt publication (scanners.py:577-627, :3130-3176); a trusted commit in the gap is deleted by rollback on receipt failure. Fix: hold exclusion through receipt publication and rollback, or refuse a stale rollback by generation.
- LOW thread-shared depth accounting in inbox_lock.py:299-324. Fix: ownership and refcount in one registry entry, serialized initialization.

Cleared: a timeout before a write body leaves no partial publication, context unwinding releases the run lock, Windows participants lock the same byte range.

Process note: round 7 is the last fix lane for this PR in the current wave. Whatever daybreak round 7 still reports becomes follow-up issues plus a residual list in the module docstring, and the PR merges as a net hardening over main.

solomonneas:
Round-7 fixes landed at 18c998d5 (Ox deep lane, conductor-1197, hub-tracked), with origin/main merged in. All five round-6 findings reproduced failing-first:

- HIGH escaped descendant: proc.run(..., supervise_group=True) reaps the owned process group once the direct child exits regardless of pipe state and records Result.descendants_reaped; the scanner receipt carries descendants_reaped (test_scanner_run_reaps_silent_descendant_holding_writer_lock; the earlier deadline test now probes release while the unrelated holder is still alive).
- MEDIUM ingestion lock timeout: run status failed, rc 1 (test_ingestion_lock_timeout_marks_run_failed_and_returns_nonzero).
- MEDIUM writers reading before locking: add, context, handoff promote, single promote, dismiss --all, dismiss single, archive now read inside _canonical_inbox_write windows; single promote takes the task-ledger lock first (seven injected-commit tests; the order test also asserts every inbox read happens under the run/writer locks).
- MEDIUM rollback deleting a gap commit: the writer lock spans append, receipt publication, and rollback; the rewritten test drives the real rollback path with a separate-process peer commit.
- LOW thread depth: ownership and depth in one _ActiveInboxLock entry with serialized init; fd closes exactly once, exclusion holds (test_overlapping_thread_acquisitions_share_one_owner_close_once_and_stay_exclusive).

Verification (Brigade-captured): ruff check, ruff format --check, mypy, eleven scanner/import/proc/services/ledger/guards/sweeps suites 654 passed 1 skipped; scanners + proc suites re-run after the main merge. Daybreak round 7 next (final): its findings go to a follow-up issue and the docstring residuals, then the PR flips ready.

solomonneas:
Daybreak round 7, final (read-only, hub-tracked run in conductor-1197) on 18c998d5: 1 HIGH / 2 MEDIUM / 1 LOW, no new class in the two-lock design. Per the process note these are filed as #1215 with line references and fix shapes and recorded in the inbox_lock.py docstring residuals:

- HIGH a setsid() or double-forked descendant escapes killpg supervision; needs cgroup/PID-namespace containment (descendants_reaped currently means the process group was reaped).
- MEDIUM reentrancy is process-wide, not thread-owned; target is per-thread ownership with other threads serializing.
- MEDIUM five canonical writers outside work_cmd (migration, trust_gate, session_ops, issue_ops, actions_dispatch) still read before locking.
- LOW InboxLockTimeout escapes as a traceback on four CLI paths instead of rc 1.

Closed over seven rounds, each with failing-first tests: run-wide exclusion for every work_cmd canonical writer; hardened lock-file acquisition (no-follow dirfd, regular file, nlink 1, dev/ino re-check); inherited-descriptor design replaced by a launcher-owned run lock plus per-process writer lock and a capability-free run marker; deadline-bound acquisition with typed failure; process-group reaping after the direct child exits; bounded and mutation-checked inbox reads; Windows ResumeThread failure handling; the merged proc.run wake loop. One follow-up commit is landing to make the round-7 thread test deterministic, then the PR flips ready and merges on green CI as a net hardening over main.

solomonneas:
Follow-up at c0539fb3: the round-7 overlapping-thread test now holds all racers inside the acquired context on a second barrier before any exits, so the single-owner assertion is deterministic (it failed once under load in a combined run; five isolated and three combined reruns green, lock code unchanged). The inbox_lock.py docstring records the #1215 residuals. Gates (Brigade-captured): ruff check, ruff format --check, mypy, scanners + proc + imports suites, scanners + proc rerun, all exit 0. Marking ready; merging on green CI.

solomonneas:
CI on c0539fb3: shard 0 failed on all three Python versions (all other shards green): tests/test_work_cmd_edges.py::test_concurrent_edge_writes_do_not_lose_dependency_edges (RunLockError on tasks.json.lock) and tests/test_work_store_measurement.py::test_measure_matrix_json_and_sqlite_report_shape. Reproduced locally. Cause: the round-5 change removed the bounded wait around the task-ledger lock while fixing the lock-order inversion, so trusted concurrent writers now fail fast instead of serializing. A fix lane is restoring deadline-bound waiting for the task-ledger lock (same style as the inbox locks, lock order unchanged) and gating on the full set of suites that import proc, work_cmd, or scanners before this PR merges.

solomonneas:
CI regression fixed at 535b29b9. Root cause: the round-4 sendback commit removed the bounded retry around the task-ledger lock along with the inherited-lock inversion, so trusted concurrent writers of tasks.json failed immediately with RunLockError (the edges test) and the work-store measurement claim race exited rc 2 instead of the CAS rc 13. Fix in work_cmd/ledger.py: outside writers wait up to 30 s (TASK_LEDGER_LOCK_DEADLINE_SECONDS, typed TaskLedgerLockTimeout on expiry); marked scanner children keep a 5 s fail-fast bound so a child holding the writer lock still cannot pin the task lock for a run window; lock order task/run/writer and runguard validation untouched. New test test_concurrent_task_ledger_writers_serialize_instead_of_failing, verified red under the old fail-fast behaviour.

Verification (Brigade-captured): ruff check, ruff format --check, mypy, focused set edges + measurement + scanners + imports + ledger + services + proc, and the lane's full run of the 88 suites importing proc, work_cmd, or scanners: 3379 passed, 2 skipped, 2 failed only on the known host-environmental crontab syntax tests. Merging on green CI.

Issue #1230, opened 2026-08-26
docs: deprecation policy for command paths


solomonneas:
## Problem

The command-surface review (see the parser/inventory contract issue) found 231 parser paths with no direct test and 69 inventory-only paths. The temptation is to delete them. Zero tests is not zero users: the daily runbooks, care entries, and other machines' cron jobs invoke paths that no pytest references. There is currently no written rule for how a command path leaves the CLI.

## Proposal

Add a short deprecation policy to `CONTRIBUTING.md` (or `docs/`) and enforce the mechanical parts:

- a path is removed only after one stable release as a deprecated alias that still works and prints a one-line notice naming the replacement
- the PR that deprecates a path must cite usage evidence: receipts, runbook references, `brigade evidence search` results, or an explicit "no evidence found" line
- the command inventory marks deprecated paths and the release notes list them
- a test asserts that every path marked deprecated in the inventory still parses

Candidates to review first under that policy, not to remove now: `budgets`, `untrusted`, the four ActionQueue trees, and the release namespaces.

## Evidence

- run `c5d68257.20260826-214607-b7de55c8`, lane C and synthesis

Issue #1228, opened 2026-08-26
types: mypy override ratchet, starting with ledger, aboyeur, skills_cmd


solomonneas:
## Problem

`src/brigade/py.typed` ships and CI runs mypy (`.github/workflows/ci.yml:265`), but the single override at `pyproject.toml:88` sets `ignore_errors = true` for 38 module patterns, including `brigade.aboyeur`, `brigade.work_cmd.ledger`, `brigade.skills_cmd`, `brigade.doctor`, `brigade.mcp_*`, `brigade.release_cmd.*`, `brigade.security_cmd.*`. The modules an auditor would read are the ones the gate skips.

Measured with the venv mypy and no override (`--config-file /dev/null --ignore-missing-imports`, receipt `20260826-214545-work-verify-02aa8c`): 2086 errors across 74 files. Two distinct populations:

1. The 39 submodules produced by the #203 split all contain `globals().update(vars(_family_base))` (example: `src/brigade/daily_cmd/telemetry_ops.py:40`). mypy cannot see names that arrive that way, so they produce 1262 `name-defined` errors (`WARN` 84, `SCHEMA_VERSION` 67, `_field` 55, `OK` 35). Every one of those 39 files is in the override. The split created this debt.
2. The three real god files carry only 122 errors: `work_cmd/ledger.py` 46, `aboyeur.py` 10, `skills_cmd.py` 66. Almost all are `union-attr` and `arg-type` on `Optional` values (`Item "None" of "Any | dict | None" has no attribute "get"`), for example `ledger.py:734`, `aboyeur.py:1337` (`is_cli_allowed` receives `str | None`), `skills_cmd.py:430`.

## Proposal

- Phase 1: remove `brigade.work_cmd.ledger`, `brigade.aboyeur`, `brigade.skills_cmd` from the override and fix the 122 errors. These are the cheapest and the highest-value modules.
- Phase 2: for each #203 family, replace `globals().update(...)` with explicit imports from the family base, then drop that family's patterns from the override. One family per PR.
- Guard: a test that asserts the override list only shrinks (compare against a checked-in baseline), so a new module cannot be added to it without editing the baseline in the same PR.

## Evidence

- run `c5d68257.20260826-214607-b7de55c8` (lane B could not run mypy in the worker; numbers above are from the operator's verify receipt)

solomonneas:
[...]

Issue #1226, opened 2026-08-26
test: repo-wide 2,000-line module ratchet with a shrinking allowlist


solomonneas:
## Problem

`tests/test_phase257_facades.py::test_phase257_production_family_files_stay_under_2000_lines` (added in #203) only scans the seven roots that PR split: `daily_cmd`, `center_cmd`, `security_cmd`, `handoff_cmd`, `release_cmd`, `work_cmd/session`, `cli/work`. Nothing guards the rest of `src/brigade`, so the ceiling passes today while these modules sit above it:

| file | lines |
|---|---:|
| `src/brigade/work_cmd/ledger.py` | 6850 |
| `src/brigade/aboyeur.py` | 5362 |
| `src/brigade/skills_cmd.py` | 5305 |
| `src/brigade/runs_cmd.py` | 3584 |
| `src/brigade/claude_hooks/runtime.py` | 3389 |
| `src/brigade/work_cmd/scanners.py` | 3332 |
| `src/brigade/research_cmd.py` | 3088 |
| `src/brigade/outcome_cmd.py` | 2903 |
| `src/brigade/run_redaction.py` | 2802 |
| `src/brigade/harness_profile_cmd.py` | 2570 |

`ledger.py` was 2036 lines at v0.26.1 (2026-08-09) and reached 6850 seventeen days later, almost entirely from #948, #949, #950, #1047, #1057, #1058, #1060, #1070, #1185, and #1207. A scoped ceiling cannot catch that.

## Proposal

Add `tests/test_module_size_ratchet.py`:

- scan every `src/brigade/**/*.py`, ceiling 2000 lines
- `LEGACY_OVERSIZED: dict[str, int]` maps each current offender to its line count at the time the test lands; that number is the file's cap
- assert: files not in the allowlist are at or under 2000; allowlisted files are at or under their cap; every allowlist entry exists and is still above 2000 (otherwise the entry must be deleted)
- the docstring states the rule: entries may be removed or lowered, never added or raised

Retire or fold the phase-257 test once the repo-wide one covers its roots.

## Evidence

- adversarial review run `c5d68257.20260826-214607-b7de55c8`, lane A and synthesis
- verify receipt `20260826-214803-work-verify-8aab89` (current ceiling test is green)

solomonneas:
Complete allowlist as of `2b55678a` (every `src/brigade/**/*.py` over 2000 lines, engines excluded); the table above stopped at ten:

```
[...]

PR #1234, opened 2026-08-27
types: un-ignore aboyeur, skills_cmd, and the ledger facade in mypy (#1228 phase 1)

solomonneas:
## What

Phase 1 of #1228: drop `brigade.aboyeur`, `brigade.skills_cmd`, and `brigade.work_cmd.ledger` from the mypy `ignore_errors` override and fix what `python -m mypy` then reports, plus a guard so the override list can only shrink.

- `pyproject.toml`: three patterns removed, nothing else touched. 35 patterns remain.
- `src/brigade/aboyeur.py`: 10 errors fixed. `Optional` guards on `output_dir`, `orchestrator.cli`, `params`, `stored_status`; one annotation; `ControlServer` now takes `output_dir / "control.sock"` directly, which is the only value `control_socket` can hold at that site (set at the app-server branch under the same `output_dir is not None` condition).
- `src/brigade/skills_cmd.py`: 66 errors fixed. Almost all are `x = rec.get("k") if isinstance(rec.get("k"), dict) else {}` rewritten as a bind-once walrus so the isinstance narrows the value actually assigned; two `cast(Literal[...])` on projection mutation kinds; `getattr(st, "st_reparse_tag", False)` (same truthiness as `0`); `installed_raw is not None` in place of `installed_present`, which is defined as exactly that; one `assert dir_fd is not None` after `_anchor_open_chain` without `missing_ok`, which already guarantees a descriptor.
- `src/brigade/work_cmd/ledger/__init__.py`: 0 errors; already clean after #1232.
- No `# type: ignore` added anywhere.
- `tests/test_mypy_override_ratchet.py` + `tests/fixtures/mypy_override_baseline.txt`: the current `ignore_errors` pattern set must be a subset of the frozen 35-pattern baseline (removals pass, additions fail), and every pattern must still match a module or package under `src/brigade` (stale entries fail). Works on 3.10 via a text parser since `tomllib` is 3.11+.

## Why

#1228, from the review in #1226 through #1230: `py.typed` ships and CI runs mypy, but the modules an auditor would read were the ones the override skipped. These three were the cheapest to reclaim (122 errors measured, 76 after the ledger split). Phase 2 (the 39 `globals().update(vars(...))` facades from #203) is separate.

## Verify

```
.venv/bin/python -m mypy                       # Success: no issues found in 445 source files
.venv/bin/ruff check . && .venv/bin/ruff format --check .
.venv/bin/python -m pytest tests/test_mypy_override_ratchet.py tests/test_aboyeur.py tests/test_skills_cmd.py tests/test_ledger_facade.py tests/test_run_cli.py tests/test_run_lifecycle.py tests/test_run_checkpoint.py tests/test_module_size_ratchet.py -q
```

Brigade receipts: mypy `20260827-014022-work-verify-87f82d`; ruff check and format exit 0 in the same session; pytest receipt below.

Pytest (receipt `20260827-014036-work-verify-7acc37`): 753 passed in 73s across the eight files above.

## Look hardest at

The four hunks that touch control flow rather than annotations: `aboyeur.py` lines 1337 and 1450 (`orchestrator.cli is None` early returns, previously would have passed `None` into `is_cli_allowed` / `hides_write_tools`), `aboyeur.py` line 4638 (`ControlServer` path), and `skills_cmd.py` `selector is not None` (unreachable for `None`, which already exits on `{"valid": False}`). Everything else is a rename or a bind-once narrowing.

Refs #1228


solomonneas:
Daybreak (bounded defensive pass, run `c5d68257.20260827-014249-c7fd650f`) on `c4fa56c3`, focused on the hunks that change control flow:

- `aboyeur.py:1337` and `:1450` (`orchestrator.cli is None`): before, `is_cli_allowed(None, ...)` accepted `None` until a later `startswith` or raised in `fnmatch`, and `hides_write_tools(None, ...)` raised on `None.split()`. Now the orchestrator returns a failed result and reports write tools as not hidden. Exceptional failure becomes an explicit fail-closed refusal; nothing launches.
- `aboyeur.py:4638` (`ControlServer(output_dir / "control.sock", ...)`): `control_socket` is `None` at 4299, set to exactly that expression at 4544, and only reset after app-server or control-server startup failure (4635, 4646). At the call site, a started app-server plus non-null `output_dir` means the old value was the new expression.
- `skills_cmd.py` `assert dir_fd is not None`: `_anchor_open_chain` (1470) returns `None` only with `missing_ok=True`, which the call at 1824 omits. If the contract were violated, the assert aborts before `_collect_fd_tree` would resolve descriptor operations against the process cwd.
- `getattr(st, "st_reparse_tag", False)`: all three sites (1396, 1738, 5043) use boolean context; no comparison with zero exists.
- One intentional malformed-input change: a non-string `run.json` status now defaults to phase `"run"` instead of potentially raising on an unhashable value (`aboyeur.py:2856`).

Verdict: approve.

solomonneas:
@coderabbitai review

solomonneas:
CI status: `test shard (3.10, 3)` has failed three times on the same test, `tests/test_claude_hooks_user_target.py::test_hook_run_latches_after_repeated_timeouts`, and the aggregate `test (3.x)` jobs cascade from it. The same shard passed on main at `41265dc7` and on #1232; this PR does not touch `claude_hooks` and nothing there imports `aboyeur` or `skills_cmd`; the new ratchet tests hash to shards 0 and 1, so shard 3's contents are unchanged from main's passing run.

Root cause is a pre-existing runtime race, reproduced locally and filed as #1236: the timeout-latch state write runs in a best-effort thread capped at 0.5 s, and when `write_session_state` takes longer (a contended runner), the parent moves on and a stale writer clobbers the latch. Slowing `write_session_state` to 0.6 s makes the test fail deterministically (receipt `20260827-021427-work-verify-ba48c7`); twelve unmodified repeats pass. Re-running the failed jobs; if the runner stays contended, the fix in #1236 is the way through rather than more re-runs.

solomonneas:
The 3.10 shard failure is fixed at the source in #1237 (per-session lock around the hook latch state, with a deterministic regression test that fails on main). Once #1237 merges, this branch gets rebased onto main and CI re-run; no re-run lottery.

solomonneas:
#1237 merged as 4f9452ee with the latch fix; this branch is rebased onto it as 99b20dec (same single commit, no conflicts). CI re-running; merging when green.

PR #1243, opened 2026-08-27
types(release_cmd): replace the namespace copy with explicit imports (#1228 phase 2, 2/7)

solomonneas:
## What

Family 2 of 7 for #1228 phase 2. Every `release_cmd` submodule that copied `paths`'s globals at import time (`globals().update(vars(_family_base))`) now imports what it uses explicitly. Leaf modules (`paths`, `version_guard`) are imported by name; the cyclic group around the `evidence` hub (`candidate`, `candidate_audit_ops`, `ci`, `install_smoke`, `schema_ops`, `commands`, and `evidence` itself) uses module aliases (`from . import evidence as _evidence`, and the reverse in `evidence.py`) so partially initialized modules resolve at call time instead of raising `ImportError` during the facade's import sequence. The facade `__init__.py` is unchanged, so `_sync_module_globals` and `_CommandFamilyFacade.__setattr__` still propagate monkeypatches to every submodule.

`brigade.release_cmd` and `brigade.release_cmd.*` leave the mypy override and `tests/fixtures/mypy_override_baseline.txt`. The errors that surfaced (106, almost all `union-attr` on Optional `.get()` chains in `candidate.py`, `evidence.py`, `candidate_audit_ops.py`) are fixed with bind-once narrowing; no `type: ignore` added.

## Verify

mypy exit 0 (446 files), ruff check and format clean, size-ratchet baseline ok; release, facade, ratchet, and adjacent suites: 172 passed (receipt `20260827-132734-work-verify-d64db4`; release, version-sync, phase-257 facades, override and size ratchets, ci-workflow, component-manifest-provenance, cli-setup).

Workers: agy_flash (Gemini 3.7 Flash via Antigravity) for the conversion and narrowing, cursor_grok for the import-cycle fix.

## Look hardest at

The module-alias call sites in `evidence.py` and `candidate.py`: a name that was previously a direct import is now `_module.name`, which is the behavior the old namespace copy had for monkeypatching but not for import order. `tests/test_phase257_facades.py` covers dotted-path reachability and facade patching.

Refs #1228

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,8 +109,6 @@ module = [
     "brigade.phases_cmd.constants",
     "brigade.phases_cmd.session_lifecycle",
     "brigade.phases_cmd.session_ops",
-    "brigade.release_cmd",
-    "brigade.release_cmd.*",
     "brigade.security_cmd",
     "brigade.security_cmd.*",
     "brigade.work_cmd.backup",
--- a/src/brigade/release_cmd/candidate.py
+++ b/src/brigade/release_cmd/candidate.py
@@ -10,8 +10,8 @@ from .. import reportstore, roadmap_cmd, work_cmd
 from ..guard.engine import scan_text
 from ..guard.policy import Policy, default_policy, load_policy
 from ..localio import utc_now as _now
-from . import evidence as _evidence
-from . import schema_ops as _schema_ops
+from . import evidence as _evidence_mod
+from . import schema_ops as _schema_ops_mod
 from .paths import (
     RELEASE_CANDIDATE_STALE_HOURS,
     RELEASE_PRIVATE_PATH_RE,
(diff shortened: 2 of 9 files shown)


solomonneas:
CI: test shard 3 fails on every Python at tests/test_daily_driver_cmd.py::test_daily_repair_unblock_telemetry_context_and_release_evidence with TypeError: 'module' object is not callable. Cause: the sibling modules alias the hub as `from . import evidence as _evidence`, which collides with the real `_evidence()` in evidence.py:257; the facade's `_sync_module_globals()` exports the first-seen binding (install_smoke's module alias) under `release_cmd._evidence`. Fix incoming: module aliases become `_<module>_mod` everywhere in the family, and the remaining family templates now require that suffix plus a collision check.

solomonneas:
Fixed: sibling aliases are now `_<module>_mod` throughout release_cmd; `release_cmd._evidence` is the function again (asserted with inspect.isfunction). Locally: mypy exit 0, ruff clean, daily-driver + release + facade + ratchet suites 138 passed. CI re-running.

PR #1360, opened 2026-09-01
fix(run): charge the capture cap against retained bytes, not stream volume

### LINKED ISSUE #1144: bug: preserve worker result when app-server output exceeds 1 MiB

solomonneas:
Fixes #1144.

## Root cause

The 1 MiB capture cap was charged against **every byte read off the transport**, not against **what Brigade retains**.

`AppServer._ingest_line` charged every JSONL record to one per-thread `ByteBudget`: reasoning deltas, tool-call notifications, command output. Nearly all of it is routed to a queue, consumed and discarded, so it never accumulates, but it still spent the budget. A long agent session emits several hundred KB of session chatter before its final answer.

Measured against a 4 KB cap, 11 `commandExecution` records (4200 bytes) tripped it with **zero bytes of agent text retained**.

The cascade: trip -> `_signal_output_limit()` terminates the app-server child -> `turn/completed` never arrives -> the salvage added in #1188/#1206/#1200 requires `completed_observed=True` -> salvage is refused -> the run fails while the worker's edits sit finished in the worktree.

That is why the earlier fixes did not hold. They guard "the final answer itself is over the cap" and cannot fire for "chatter ate the budget first", which is the shape that actually occurs. Reported reproductions on `c19edf5b` had persisted event streams of 874,509 and 838,303 bytes, both under the cap on their own.

## What changed

The #1108 invariant is bounded retained memory. That is preserved; the cap is not raised.

- `proc.MAX_STREAM_BYTES` bounds the untrusted stream Brigade reads and is the only thing that terminates the child. Reading and discarding does not grow memory.
- `proc.MAX_CAPTURE_BYTES` bounds retained text, charged only for output actually kept: agentMessage deltas, completed texts, orphan payloads.
- Retention keeps head **and** tail (`bound_text_ends`, `bound_text_tail`). The final answer lives at the end of an agent stream, so head-only truncation discarded exactly the part that mattered.
- A turn that completed with truncated text returns `ok`. Truncation is a capture artifact, not a failed turn.
- `output-limit` maps to `OUTPUT_CONTRACT_VIOLATION` instead of `TRANSPORT_UNAVAILABLE`, so a truncated capture no longer makes recovery/resume re-probe a healthy seat and report it dead. This applies to both transports, not just app-server.
- Worker receipts record the observed byte count and the configured cap.

## Invariants deliberately preserved

- A record larger than the ceiling is still rejected before `json.loads`, so it is never fully buffered (#1108), and never signals completion (#1200).
- A record that merely *tips* the accumulated ceiling is parsed first, so a final `turn/completed` is recorded before the child is stopped. Moving the ceiling would otherwise have silently reopened #1200 at the new knob.
- Retention past the cap rolls through one shared bounded tail per turn, not a buffer per `item_id`. Per-item buffers would let N item ids each retain a full cap, and re-bounding on every delta is O(n^2).
- Cap constants resolve at construction rather than at class-definition time. As dataclass field defaults they froze the module value at import, making the cap untestable and misreported in the run record.

## Acceptance criteria from #1144

- [x] Regression test covers a stream larger than the cap followed by a small final result
- [x] The final result remains available
- [x] Output truncation does not poison seat health
- [x] The run record includes the original byte count and configured cap

## Verification

- `./scripts/verify-focused tests/test_proc.py tests/test_codex_appserver.py tests/test_agents.py tests/test_run_transport.py` -> exit 0 (ruff lint, ruff format, version-sync, managed snapshot, template-profile snapshot, mypy, 307 tests)
- `tests/test_worker_failure.py` -> exit 0
- Full suite: 11,100 tests, 16 failures. **All 16 fail identically on the merge base `635d5fcd` with this branch's changes absent** (verified in a pristine worktree), covering `test_aboyeur`, `test_run_cli` cancellation, `test_run_detach`, and the fleet-claim suites. No regressions from this change. Those pre-existing failures are not addressed here.

## Note

The `brigade run` dispatched to implement this failed with `failure.kind=output-limit` after 370s: a live reproduction of #1144 on the run fixing #1144. Its edits survived in the worktree while its report was discarded, exactly as the issue describes.

## Summary by CodeRabbit

- **New Features**
  - Added output byte counts and configured limits to execution results and persisted worker records.
  - Retained both the beginning and end of output when truncation occurs, with a clear truncation marker.
  - Distinguished output capture limits from total stream limits for clearer diagnostics.

- **Bug Fixes**
  - Successful executions remain successful when output is truncated.
  - Corrected classification of output-limit conditions as output contract violations.
  - Preserved completion details and final responses in truncated app-server output.


solomonneas:
@coderabbitai full review

solomonneas:
@coderabbitai full review

solomonneas:
All CodeRabbit findings from both rounds are addressed in `f1651389` and `13582171`:

Round 1
- Truncated zero-exit output no longer returns before `_parse_grok_final_output` / `validate_final_output`; it flows through the normal output contract and carries truncation metadata onto whatever result that produces. Regression test: `test_truncated_exec_output_still_runs_final_output_validation`.
- The app-server output-limit branch forwards `output_bytes` / `output_cap_bytes`.
- Resume carries `failure_kind`, `output_truncated`, `output_bytes`, `output_cap_bytes` into the reconstructed `WorkerResult`.

Round 2
- `ByteBudget.observe` now latches on overflow instead of letting a later small write report True again.
- A partially accepted delta retains the prefix the budget was charged for.
- `_output_limit_result` reports stream volume against `MAX_STREAM_BYTES`, the ceiling that actually fires on that path.
- Removed unreachable code after the unconditional return in `_salvage`.

Verified: `./scripts/verify-focused` over test_proc, test_codex_appserver, test_agents, test_run_transport, test_worker_failure, test_run_resume, exit 0 (Brigade receipt `20260901-121230-work-verify-659073`).

Issue #1445, opened 2026-09-05
brigade run pins the Hub canonical model slug for OpenCode seats; native launch id is unreachable


solomonneas:
## What happens

`brigade run --worker muse13` (roster `cli = "opencode"`, `model = "opencode/muse-spark-1.3-contributor-free"`) fails immediately:

```
Error: {
  "name": "UnknownError",
  "data": { "message": "Unexpected server error. Check server logs for details.", "ref": "err_8babbcae" }
}
```

The run journal shows the seat was admitted as `"model": "opencode-muse-spark-1.3-contributor-free"` (the Hub row's canonical slug), and the adapter pins that value as `opencode run -m opencode-muse-spark-1.3-contributor-free ...`. OpenCode does not know that id.

Direct probes with the same PATH, cwd, setsid, and closed stdin:

- `opencode run -m opencode/muse-spark-1.3-contributor-free "Reply with exactly: OK"` -> `OK`, exit 0
- `opencode run -m opencode-muse-spark-1.3-contributor-free "Reply with exactly: OK"` -> the same `UnknownError`, exit 1

Trying to correct it from the CLI is blocked on both sides:

- `brigade run --worker muse13 --model opencode/muse-spark-1.3-contributor-free` -> `fleet model policy denied seat 'muse13': --model 'opencode/muse-spark-1.3-contributor-free' does not match Hub model 'opencode-muse-spark-1.3-contributor-free'`
- the suggested `brigade fleet models set opencode opencode-muse-spark-1.3-contributor-free muse13 --brigade-cli opencode ...` re-pins the same slug; `fleet models set` has no flag to record a native brigade model binding, and the Hub model name pattern rejects a slash.

## Why it matters

This is the live case of "canonical slugs are not launch IDs" that the fleet authority work is built around. Any provider whose native id contains a slash or differs from the Hub slug (OpenCode Zen models, Cursor `provider/model` ids) cannot be launched through `brigade run` even though the seat is enabled and healthy. Muse Spark 1.3 is one of the two free lanes right now.

## Expected

- The adapter pins the seat's effective brigade binding model (`bindings.brigade.model`, projected through `consumer_launch_bindings` for the `brigade-run` consumer) when present, and only falls back to the canonical slug when no binding exists.
- `brigade fleet models set` accepts `--brigade-model <native id>` so operators can record that binding on the legacy row without waiting for policy activation.
- The `--model` override error names the binding fix instead of a command that re-pins the broken slug.

## Environment

[...]

PR #1451, opened 2026-09-06
feat(fleet): centralize fleet policy enforcement through the hub control plane

solomonneas:
## What

Seat, model, privacy, and machine decisions move out of per-machine rosters into one signed, versioned fleet policy document served by the hub. This branch is the integrated control plane: policy core and migration, the session handshake, launch proof, routing authority, consumer launch bindings, remote transport, model inventory, and the Deck and policy pages.

## Why

Every machine currently carries its own roster, so a seat can be enabled on one host and drifted on another, a canonical model slug gets pinned as if it were a launch id, and nothing proves that the policy a session acknowledged is the policy it launched under. Centralizing enforcement gives one authority, one digest, and a launch proof bound to the exact prepared context.

## How it holds together

- **Handshake.** `session prepare` emits a mandatory `context_hash` (`sha256:` plus 64 hex) over the effective settings, sources, overrides, instruction block, selected canonical and native model, consumer, repo, session, origin, and delegation identity. An applied acknowledgement must carry that exact hash and persists through `record_acknowledged_policy`. A same-version re-prepare with a different override or model changes the hash and invalidates any earlier acknowledgement. Legacy resolve-only receipts stay non-launchable.
- **Launch proof.** `models admit --phase launch` requires `--repo` and `--policy-context-hash`, and runs before any model lease. Enrolled transports that cannot honor the gate deny instead of skipping it.
- **Capacity.** Routing reservations and legacy model leases count against the same seat limit through `fleet_hub_capacity.capacity_usage`. Deduplication needs the exact decision, session, and node execution triple; unlinked leases stay counted.
- **Remote transport.** `brigade run` resolves the route before entering the model lease. A seat placed on another machine never takes a local lease or reaches a provider; work goes to the local `t3-fleet` client under a hub delegation, and a delegated writable result reports `pending-handoff` rather than claiming a local edit.
- **Staged activation.** The migration keeps legacy rosters working until deliberate activation. Nothing in this PR activates the live hub.

## How to verify

Focused gate on the integrated tree, receipt `20260906-000859-work-verify-d7d5c3`:

```
./scripts/verify-focused tests/test_fleet_*.py tests/test_run_transport.py tests/test_run_preference.py tests/test_run_projector.py tests/test_run_shadow.py tests/test_worklore_store.py tests/test_aboyeur*.py
1318 passed in 162.20s
```

Full serial `./scripts/verify` is running through Brigade now; I will post its receipt on this PR when it finishes.

## Look hardest at

- `fleet_hub_policy_api.py` acknowledge path: hash comparison happens inside the `BEGIN IMMEDIATE` transaction against an immutable pending row.
- `fleet_hub_routing._seat_occupancy`: the only exclusion is the caller's own reservation by its exact execution triple.
- `run_transport.dispatch`: the route resolves before `model_lease` is entered.
- Schema bump to 20 in `test_run_preference.py` and `test_worklore_store.py` is intentional (additive fleet policy tables).

## Known follow-ups (not in this PR)

- #1445: the adapter pins the hub canonical slug for OpenCode seats, so native ids with a slash cannot launch until bindings are honored at dispatch.
- T3 side still needs `submit --delegation-id` to adopt the hub delegation instead of creating a second reservation.
- Deck `control_plane_status.authority` still reports a fixed inactive value; it should read the migration activation state.


solomonneas:
Full serial gate, first run (receipt `20260906-002815-work-verify-b75df7`): 15 failed, 12419 passed, 10 skipped in 45m29s.

- 14 of the 15 were the launch environment, not the code. I started the gate with `nohup ... &` from a non-interactive shell, which starts the process with SIGINT ignored (`SigIgn 0x7`), and every child inherits that. The failing tests in `test_fleet_claims` and `test_run_cli` all depend on SIGINT or a main-thread interrupt reaching a subprocess, so they timed out or reported "run kept going". Each passes in isolation (`20260906-010203`-series receipts). The rerun is a user systemd unit with default signal dispositions (`SigIgn 0x4`).
- 1 was real: `test_module_size_ratchet` caught `fleet_policy_page.py` at 2088 lines. Fixed in a21c4cc8 by moving form parsing and scoped document construction into `fleet_policy_form.py`, re-exported from the page. Focused gate `20260906-011845-work-verify-9794f7`, 211 passed.

The PR was `DIRTY` against main (two conflicts: `fleet_model_admission.py` LKG loader, where both `high_water_record` and main's revision-rollback check are kept, and the `PRESERVED_FIELDS` count, now 69), which is why no CI run was created. Merged main in fec6732d; focused gate on the merged tree `20260906-012257-work-verify-aab079`, 151 passed, plus 332 passed across the admission, proof, roster, transport, and routing files.

Full serial gate is rerunning on the a21c4cc8 tree now; receipt to follow.

solomonneas:
CI on fec6732d: everything green except `content-guard` (1 blocking finding, which also skipped `build-and-test`). The audit summary does not name the finding; running the guard library over the tracked tree with the baseline applied pointed at `tests/test_fleet_policy_cli.py:290`, where the scrub test passes `https://user:super-secret-token@github.com/...` on purpose and asserts the token never reaches output. The guard's email rule reads that as an address. Fixed in 18fd0abf with a reviewed `# content-guard: allow email` tag on that line (the assertion is unchanged), plus the fixture docstring no longer embeds a literal port. Local audit with the baseline now reports `blocked: false`; focused gate `20260906-013016`-series, 99 passed.

solomonneas:
Full serial gate on the a21c4cc8 tree, launched as a user systemd unit with default signal dispositions (receipt `20260906-011948-work-verify-d4ae37`, exit 0, status completed):

```
12434 passed, 10 skipped, 7 warnings, 68 subtests passed in 2620.99s (0:43:40)
```

The 13 claim-loss and SIGINT-cancel failures from the first run did not reproduce, confirming they came from the ignored-SIGINT launch environment. Combined with CI green on 18fd0abf (37 success, 0 failure, 5 path-gated skips) and `mergeStateStatus: CLEAN`, this PR is open-green.

Issue #1482, opened 2026-09-06
ci: add a per-file Windows pytest lane with a known-failing allowlist and ratchet


solomonneas:
## Problem

CI runs the pytest suite only on Ubuntu. The Windows lanes on `main` (`windows-native-acceptance`, `quickstart-smoke (windows-latest)`, `notify-build-and-test (windows-latest)`) are green, but they execute only `scripts/windows-native-acceptance.ps1` and a quickstart smoke. None of the Windows-only defects found on 2026-09-06 were visible to CI.

A per-file run on the Windows fleet host covered 229 of 446 test files before it had to be stopped:

| metric | count |
|---|---|
| passed | 4721 |
| failed | 1189 |
| files with failures | 109 |

A full-session run is not possible yet because of #1473, so a Windows pytest lane needs to run one process per test file. The per-file driver used for the audit (`CREATE_NO_WINDOW | CREATE_NEW_PROCESS_GROUP`, a separate `--basetemp` per file, an 8-wide thread pool) finished about half the suite in 20 minutes on a 12-core host.

Two host-side gotchas for whoever wires this:

- Concurrent pytest sessions on one Windows host fight over `%TEMP%\pytest-of-<user>`. The numbered-dir cleanup and the `pytest-current` junction fail with WinError 5 and 32 unless every session gets its own `--basetemp`.
- Test children inherit no console from a no-window parent and each opens a visible CMD window when the runner is an interactive session. Run the lane under a non-interactive logon.

## Expected

- A `windows-pytest` CI job that runs the suite per file on `windows-latest`, initially `continue-on-error: true` with a committed allowlist of known-failing files, and a ratchet that fails the job when a previously passing file regresses.
- The allowlist shrinks as the Windows issues in milestone 0.28.0 close.

## Acceptance

- The job runs on every PR and publishes the per-file summary as an artifact.
- Removing a file from the allowlist without fixing it turns the job red.

Found during the 2026-09-06 Windows audit. Depends on #1473 and #1474.

solomonneas:
Runner design note from the 2026-09-06 killer hunt on the Windows fleet host, for whoever implements this lane:

- With #1473 merged, `tests/test_aboyeur.py` still took down a `powershell -File` wrapper mid-file (no pytest summary, no exit line, scheduled task ends with result 0).
[...]

PR #1499, opened 2026-09-06
fix(proc): keep KeyboardInterrupt from escaping the Windows terminate path (Fixes #1473)

### LINKED ISSUE #1473: tests: pytest suite cannot run as one session on Windows (KeyboardInterrupt escapes through the taskkill path)
## Problem

The pytest suite cannot run as a single session on Windows. Every full-suite attempt on the Windows fleet host (Gandalf, Python 3.14.6, pytest 9.1.1, xdist 3.8.0) died between 1 and 11 percent with no pytest summary, and the PowerShell wrapper around it exited without reaching its own next line. Serial runs of the same files reproduce the mechanism.

## Root cause (confirmed with `--full-trace`)

`tests/test_work_cmd_verification.py::test_run_verify_child_process_catches_keyboard_interrupt` replaces the global `subprocess.Popen` with a wrapper whose `communicate` raises `KeyboardInterrupt`. On POSIX the verify code catches that at `_run_verify_child_process` and terminates the child with `killpg`, so no second Popen is created.

On Windows the terminate path is `proc._terminate_windows_process_tree`, which calls `subprocess.run(["taskkill", "/PID", ...])`. `subprocess.run` creates another Popen through the patched constructor and calls its `communicate`, which raises again. That second raise happens inside `_terminate_verify_child` on the zero-grace retry (`src/brigade/work_cmd/verification.py:350`, via `src/brigade/proc.py:403` and `:366`) where there is no handler, so the interrupt escapes the test.

Trace excerpt:

```
src\brigade\work_cmd\verification.py:388: except KeyboardInterrupt
src\brigade\work_cmd\verification.py:350
src\brigade\proc.py:403
src\brigade\proc.py:366
C:\Python314\Lib\subprocess.py:557
tests\test_work_cmd_verification.py:2667: KeyboardInterrupt
```

Under xdist a worker interrupt tears down the whole session. Deselecting this one test still lost the run at 3 percent, so at least one other test does something equivalent. `tests/test_run_cli.py:1806` raises `SIGTERM` into the test process with `signal.raise_signal`, which on Windows has no Python-level handler installed by `run_lifecycle` and is a likely second source.

## Expected

- `_terminate_windows_process_tree` and the zero-grace retry in `_terminate_verify_child` tolerate `KeyboardInterrupt` the same way the POSIX path does, or the test patches only the verify child Popen instead of the module global.
- Tests that inject signals into the test process are marked POSIX-only or use a Windows-safe equivalent.
- A full-suite run on Windows reaches the pytest summary.

## Acceptance

- `python -m pytest tests/test_work_cmd_verification.py -p no:xdist` exits 0 or 1 on Windows and prints a summary.
- `python -m pytest tests -n 12` on Windows reaches the summary line.

Found during the 2026-09-06 Windows audit. Related: #1213, #1352.

solomonneas:
Fixes #1473

## What

- `src/brigade/proc.py`: `_terminate_windows_process_tree` tolerates `KeyboardInterrupt` from the `taskkill` call, from `poll`, and from `kill`, and never re-raises. The Windows branch of `_terminate_processes` is safe to call twice.
- `src/brigade/work_cmd/verification.py`: both zero-grace retries in `_terminate_verify_child` are wrapped so the function always returns.
- `tests/test_work_cmd_verification.py`: the interrupt test's Popen wrapper returns the real Popen for `taskkill`, so only the verify child gets the raising `communicate`.
- `tests/test_run_cli.py`: the SIGTERM raise test is skipped on Windows, where neither `run_lifecycle` nor the aboyeur terminal handler installs a Python-level SIGTERM handler.
- `tests/test_proc.py`: POSIX-runnable regression test that forces the Windows branch with a `subprocess.run` stub raising `KeyboardInterrupt` and asserts the caller returns and `kill()` ran.

## Verification

Focused gate through Brigade (ruff, format, version sync, snapshots, mypy, selected tests):

```
brigade work verify run --target . --argv-json '["./scripts/verify-focused","tests/test_proc.py","tests/test_work_cmd_verification.py","tests/test_run_cli.py"]' --capture brigade-work
run: 20260906-170547-work-verify-1b8e07
status: completed  exit=0
```

Windows confirmation on the fleet host is recorded in a follow-up comment.

### DIFF
--- a/src/brigade/proc.py
+++ b/src/brigade/proc.py
@@ -371,12 +371,18 @@ def _terminate_windows_process_tree(process: subprocess.Popen[bytes], *, timeout
             check=False,
             timeout=max(timeout, 0.1),
         )
-    except (OSError, subprocess.TimeoutExpired):
+    except (OSError, subprocess.TimeoutExpired, KeyboardInterrupt):
         pass
-    if process.poll() is None:
+    try:
+        alive = process.poll() is None
+    except KeyboardInterrupt:
+        alive = True
+    except OSError:
+        alive = False
+    if alive:
         try:
             process.kill()
-        except OSError:
+        except (OSError, KeyboardInterrupt):
             pass
 
 
@@ -400,8 +406,17 @@ def _terminate_processes(
     if os.name == "nt":
         timeout = terminate_grace + kill_grace
         for process in processes:
-            _terminate_windows_process_tree(process, timeout=timeout)
-        _wait_for_processes(processes, kill_grace)
+            try:
+                _terminate_windows_process_tree(process, timeout=timeout)
+            except KeyboardInterrupt:
+                try:
+                    process.kill()
+                except (OSError, KeyboardInterrupt):
+                    pass
+        try:
+            _wait_for_processes(processes, kill_grace)
+        except KeyboardInterrupt:
+            pass
         return
     for process in processes:
         _signal_process_group(process, signal.SIGTERM)
--- a/src/brigade/work_cmd/verification.py
+++ b/src/brigade/work_cmd/verification.py
@@ -347,13 +347,19 @@ def _terminate_verify_child(process: subprocess.Popen[bytes]) -> None:
     try:
         proc._terminate_processes((process,), terminate_grace=0.5, kill_grace=0.5)
     except KeyboardInterrupt:
-        proc._terminate_processes((process,), terminate_grace=0.0, kill_grace=0.0)
+        try:
+            proc._terminate_processes((process,), terminate_grace=0.0, kill_grace=0.0)
+        except KeyboardInterrupt:
+            pass
     survivors = tuple(verify_reaper._survivor_mocks(b | verify_reaper._collect_descendants(process.pid)))
     if survivors:
         try:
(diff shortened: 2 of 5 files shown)


solomonneas:
Windows confirmation on the fleet host (Python 3.14.6, pytest 9.1.1, basetemp outside the checkout, `BRIGADE_NO_UPDATE_CHECK=1`), running the three affected files serially on 518f780b:

```
83 failed, 219 passed, 6 skipped, 1 warning in 456.02s (0:07:36)
exit=1
```

The point of this PR is the first line: the session now reaches the pytest summary. On `main` the same command dies mid-run with a `KeyboardInterrupt` banner and no summary (trace on #1473). The 83 failures are other Windows gaps and are out of scope here: the `SIGINT`/`SIGTERM` cancel tests in `test_run_cli.py` (`send_signal(SIGINT)` is unsupported on Windows), receipt identity and archive tests in `test_work_cmd_verification.py`, and the process-group tests in `test_proc.py`. They will be triaged under #1482 once a per-file Windows lane exists.

Issue #1498, opened 2026-09-06
cloud: Jules sessions end at planGenerated with no output, and status/sync still report them as pending with no hub lease


solomonneas:
## Problem

Five Jules sessions launched on 2026-09-06 through `brigade run cloud launch --provider jules --auto-create-pr` (labels `windows-1473-terminate-path`, `windows-1478`, `windows-1479`, `windows-1480`, `windows-1481`) each produced exactly one activity, `planGenerated`, and then moved to `state: COMPLETED` about ten minutes after creation with no `outputs`, no branch on the repository, and no pull request. `POST sessions/<id>:approvePlan` returned `{}` and did not change the state.

Two Brigade-side gaps made this invisible:

1. `brigade run cloud status` kept classifying all five as `[pending]` with `branch=None` while the Jules API already reported `COMPLETED`. A terminal provider state with no outputs should classify as `needs_you` or `failed`, not `pending`.
2. `brigade run cloud sync` reported `active: 0 released: 0 needs_you: 0`, so no hub lease was ever created. The fleet hub panel showed `jules 0/15, no active leases` for the whole run, and the conductor only found out by calling the Jules REST API by hand.

## Expected

- The Jules adapter maps `COMPLETED` with empty `outputs` to a distinct classification (`completed-no-output` or `needs_you`) and surfaces the session URL.
- `cloud sync` publishes a hub lease for a Jules session as soon as launch succeeds, and releases it on a terminal state, so the hub reflects cloud work in flight.
- The launch path records whether the session stopped at the plan stage (for example a plan-approval requirement or a source setup failure) so the operator can tell a Jules-side failure from a Brigade-side one.

## Evidence

- Session objects show `sourceContext.githubRepoContext.startingBranch: main`, `environmentVariablesEnabled: true`, and no error field.
- Activities endpoint returned `count 1` (`planGenerated`) for the first session.
- `gh api repos/escoffier-labs/brigade/branches` shows no branch for any of the five labels.

Found while conducting the Windows fixes on milestone 0.28.0. The five issues were re-dispatched to the local `muse13` seat.

solomonneas:
[...]

PR #1504, opened 2026-09-06
fix(proc): keep KeyboardInterrupt from escaping the Windows terminate path (Fixes #1473)

### LINKED ISSUE #1473: tests: pytest suite cannot run as one session on Windows (KeyboardInterrupt escapes through the taskkill path)

solomonneas:
Fixes #1473 by catching `KeyboardInterrupt` in the Windows terminate path to prevent it from escaping.

Changes:
* Caught `KeyboardInterrupt` inside `_terminate_windows_process_tree`.
* Guarded the `taskkill` invocation in `_terminate_processes` with a `_windows_tree_terminated` flag so the Windows branch is safe to invoke twice.
* Added `try-except KeyboardInterrupt` blocks around the retries inside `_terminate_verify_child` to guarantee it returns.
* Excluded `taskkill` commands from the mocked `Popen` in `test_run_verify_child_process_catches_keyboard_interrupt` so it accurately tests the verification interrupt path without poisoning the terminate logic.
* Skipped `test_run_cli_terminalizes_sigterm_during_roster_snapshot` on Windows since the test leverages a POSIX signal.
* Added a new POSIX-runnable regression test to ensure that the Windows process tree terminate path tolerates `KeyboardInterrupt`.

Tested with `./scripts/verify-focused tests/test_proc.py tests/test_work_cmd_verification.py tests/test_run_cli.py`.

---
*PR created automatically by Jules for task [10860442454072640265](https://jules.google.com/task/10860442454072640265) started by @solomonneas*

### DIFF
--- a/src/brigade/proc.py
+++ b/src/brigade/proc.py
@@ -351,7 +351,7 @@ def _bind_suspended_windows_child(
         pass
     try:
         process.wait(timeout=_TIMED_OUT_DRAIN_SECONDS)
-    except (OSError, subprocess.TimeoutExpired):
+    except (OSError, subprocess.TimeoutExpired, KeyboardInterrupt):
         pass
     try:
         job.close()
@@ -371,7 +371,7 @@ def _terminate_windows_process_tree(process: subprocess.Popen[bytes], *, timeout
             check=False,
             timeout=max(timeout, 0.1),
         )
-    except (OSError, subprocess.TimeoutExpired):
+    except (OSError, subprocess.TimeoutExpired, KeyboardInterrupt):
         pass
     if process.poll() is None:
         try:
@@ -400,7 +400,9 @@ def _terminate_processes(
     if os.name == "nt":
         timeout = terminate_grace + kill_grace
         for process in processes:
-            _terminate_windows_process_tree(process, timeout=timeout)
+            if not getattr(process, "_windows_tree_terminated", False):
+                _terminate_windows_process_tree(process, timeout=timeout)
+                process._windows_tree_terminated = True  # type: ignore[attr-defined]
         _wait_for_processes(processes, kill_grace)
         return
     for process in processes:
--- a/src/brigade/work_cmd/verification.py
+++ b/src/brigade/work_cmd/verification.py
@@ -347,13 +347,19 @@ def _terminate_verify_child(process: subprocess.Popen[bytes]) -> None:
     try:
         proc._terminate_processes((process,), terminate_grace=0.5, kill_grace=0.5)
     except KeyboardInterrupt:
-        proc._terminate_processes((process,), terminate_grace=0.0, kill_grace=0.0)
+        try:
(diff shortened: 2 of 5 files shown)


solomonneas:
Closing in favor of #1499, which lands the same change with a Brigade focused-gate receipt (20260906-170547-work-verify-1b8e07) and a Windows confirmation run on the fleet host. This PR arrived from the Jules session about an hour after the session reported COMPLETED; the delay is tracked in #1498. Thanks for the parallel confirmation of the approach.

Issue #1512, opened 2026-09-07
scanners: descriptor-relative inbox and rollback have no Windows path, so the module fails closed on every call


solomonneas:
## Problem

After #1478 routed the POSIX-only `os` attributes through the `dirfd` and `proc` facades, `tests/test_work_cmd_scanners.py` is the one module that still fails broadly on Windows. On the fleet host (Python 3.14.6, basetemp outside the checkout) the seven files touched by that PR report `95 failed, 161 passed, 25 skipped`, and nearly all of the failures are in this module:

| error | count |
|---|---|
| `AssertionError: assert 2 == 0` | 13 |
| `AttributeError: module 'os' has no attribute 'O_NOFOLLOW'` (test helper) | 9 |
| `assert 0 == 1` | 7 |
| `AssertionError: Regex pattern did not match.` | 6 |
| `OSError: descriptor-relative scanner inbox operations are unavailable` | 6 |
| `OSError: scanner run artifact rollback could not restore its retained snapshot` | 5 |
| `assert False` | 5 |

The `OSError` lines are the module behaving as designed: `src/brigade/work_cmd/scanners.py` fails closed when descriptor anchoring is unavailable, which on Windows is always. The remaining assertion failures are tests asserting the POSIX outcome, and the 9 `AttributeError` lines are helpers inside the test file that still open descriptors themselves.

## Expected

Either:

- the scanner inbox and run-artifact operations gain a handle-relative Windows implementation on top of `nt_dirfd`, matching the direction of #1213 and #1352, so `brigade work scanners` works on Windows; or
- the module documents Windows as unsupported with one typed reason, the CLI surfaces that reason instead of a raw `OSError`, and the tests skip with that reason rather than asserting POSIX behaviour.

The first is the real fix; the second is acceptable as an interim step if it is explicit.

## Acceptance

- `python -m pytest tests/test_work_cmd_scanners.py` on Windows reports only passes and reasoned skips.
- No `AttributeError` for a POSIX-only `os` attribute anywhere in the module or its tests.

Found during the 2026-09-06 Windows audit; split out of #1478 so that PR can land. Related: #1213, #1352, #1482.

PR #1518, opened 2026-09-07
fix(windows): close descriptors before the NT rename and keep drive paths whole (Fixes #1480)

### LINKED ISSUE #1480: work: no-follow parent walk rejects temp paths on Windows; component paths crash without LOCALAPPDATA
## Problem

On Windows, 19 tests fail with `PermissionError: path component access denied` in `tests/test_daily_driver_cmd.py` (7), `tests/test_handoff_cmd.py` (4), `tests/test_context_learning_closeout_cmd.py` (2), and others. These are the no-follow parent walks from #1047 rejecting ordinary pytest temp paths.

Two more Windows-only shapes appeared in the same run:

- `ValueError: windows component cache root requires LOCALAPPDATA` (5, `tests/test_claude_hooks_compaction_marker.py`): the tests scrub the environment and `component_paths.cache_root` has no fallback, so the hooks crash instead of degrading.
- `tests/test_daily_driver_cmd.py` took 305 seconds for 40 tests, so a timeout or retry loop is also involved.

## Expected

- The no-follow parent walk accepts a plain directory chain under the user's temp root on Windows; the pytest `pytest-current` junction and reparse points on the path are handled explicitly rather than by refusing the whole path.
- `component_paths` falls back to a documented default when `LOCALAPPDATA` is unset instead of raising.

## Acceptance

- The three listed test files pass on Windows, and `test_daily_driver_cmd.py` finishes in under 60 seconds.

Found during the 2026-09-06 Windows audit. Related: #1047, #1022.

solomonneas:
Fixes #1480

## What

Four Windows-only defects from the 2026-09-06 audit, each with a POSIX-runnable regression test:

- `src/brigade/work_cmd/ledger/inbox_provenance.py` and `src/brigade/work_cmd/nt_dirfd.py`: the write and snapshot-rollback paths held the temporary and existing descriptors open across `authority_store._dirfd_replace`. `NtSetInformationFile` with `FileRenameInformation` returns `STATUS_ACCESS_DENIED` while a handle is open, which surfaced as a bare `PermissionError: path component access denied`. Close the descriptors before the replace, keep the POSIX fsync ordering, and name the component in the NT error.
- `src/brigade/handoff_cmd/sources.py`: `_split_outcome_line` split on the first colon, truncating `C:\Users\x\handoff.md` to `C`. Drive-qualified paths stay whole; the POSIX `path:detail` form is unchanged.
- `src/brigade/daily_cmd/config.py`: `_bounded_status_call` only bounded a section when `signal.SIGALRM` exists, so on Windows a slow section ran to completion and reported `ok` instead of `warn`. It now falls back to a worker thread with `join(timeout)`.
- `tests/test_claude_hooks_compaction_marker.py`: the env helper scrubbed the environment without `LOCALAPPDATA`, so `component_paths.cache_root` raised on Windows. The helper supplies it; `component_paths` keeps no fallback.

Path-separator assertions in `tests/test_handoff_cmd.py` now compare through `Path` instead of literal forward slashes.

## Verification

Focused gate through Brigade (ruff, format, version sync, snapshots, mypy, five test files), green on the first attempt:

```
run: 20260907-033515-work-verify-56ff5a
status: completed  exit=0
```

### DIFF
--- a/src/brigade/daily_cmd/config.py
+++ b/src/brigade/daily_cmd/config.py
@@ -509,17 +509,21 @@ def _bounded_status_call(
     timeout = timeout_seconds or _status_section_timeout_seconds()
     start = time.monotonic()
     use_alarm = threading.current_thread() is threading.main_thread() and hasattr(signal, "SIGALRM")
+    if not use_alarm:
+        # Platforms without SIGALRM (notably Windows) bound the call with a
+        # worker thread so a slow section still reports warn instead of
+        # stalling the whole status run.
+        return _thread_bounded_status_call(label, func, fallback, timeout=timeout, start=start)
     previous_handler = None
     previous_timer = None
-    if use_alarm:
-        previous_handler = signal.getsignal(signal.SIGALRM)
-        previous_timer = signal.getitimer(signal.ITIMER_REAL)
+    previous_handler = signal.getsignal(signal.SIGALRM)
+    previous_timer = signal.getitimer(signal.ITIMER_REAL)
 
-        def _raise_timeout(_signum, _frame):
-            raise _DailyStatusSectionTimeout(label)
+    def _raise_timeout(_signum, _frame):
+        raise _DailyStatusSectionTimeout(label)
 
-        signal.signal(signal.SIGALRM, _raise_timeout)
-        signal.setitimer(signal.ITIMER_REAL, timeout)
+    signal.signal(signal.SIGALRM, _raise_timeout)
+    signal.setitimer(signal.ITIMER_REAL, timeout)
     try:
         result = func()
         elapsed = int((time.monotonic() - start) * 1000)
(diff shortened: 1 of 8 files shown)


solomonneas:
Windows confirmation on the fleet host (Python 3.14.6, pytest 9.1.1, basetemp outside the checkout), same five files, `origin/main` at 207ad9ed vs this branch at a4d53dc5:

| metric | main | this PR |
|---|---|---|
| `path component access denied` | 70 | 0 |
| `windows component cache root requires LOCALAPPDATA` | 10 | 0 |
| failed | 65 | 19 |
| passed | 249 | 300 |

Both defects this PR targets are gone, and 51 more tests pass. The 19 remaining failures are separate Windows gaps in the same files, none of them the errors fixed here: `os.mkfifo` is absent on Windows, three `PermissionError` cases on a different path, and several assertions that still expect POSIX shapes. They belong with the per-file Windows lane in #1482 and the scanner work in #1512.

PR #1519, opened 2026-09-07
test(agents): platform-neutral executable fixtures and Windows shape coverage (Fixes #1479)

### LINKED ISSUE #1479: agents: dispatch tests measure the Windows shim guard instead of dispatch; cover real Windows executable shapes
## Problem

On Windows, 126 of 196 tests in `tests/test_agents.py` fail because dispatch returns `unsupported-command-shim` where the test expects the real outcome (`invalid-dispatch-args`, `non-final-output`, `tool-only-output`, `model-unavailable`, `rate-limit-error`, `output-validation`, and others). The classification comes from `src/brigade/agents.py:1232`, which maps any non-missing executable that is not a recognized native binary to `unsupported-command-shim`.

The test fixtures build executables with POSIX paths (`path="/x/ollama"`, `kind="native"`) or bare names such as `fake-grok`. On Windows those never resolve to a `.exe`, so the shim guard from #281 and #340 rejects them before the code under test runs. The tests are therefore measuring the shim guard instead of the dispatch behaviour.

Separately, the same guard needs a check against real Windows installs. On the Windows fleet host `cursor-agent` resolves to a `.ps1`, `codex` to a node-hosted `.exe` under the npm package, and `claude` to a native `.exe`. Roster doctor accepts all three today, but the tests do not cover those shapes.

## Expected

- Test fixtures create a platform-appropriate executable (a real `.exe` or `.cmd` on Windows, a script on POSIX) or the executable-resolution step is stubbed so dispatch tests exercise dispatch.
- Explicit tests for the three Windows executable shapes above: native `.exe`, npm node wrapper `.exe`, and `.ps1` shim with a per-seat command override (#1115).

## Acceptance

- `python -m pytest tests/test_agents.py -p no:xdist` on Windows: no `unsupported-command-shim` mismatches.

Found during the 2026-09-06 Windows audit. Related: #281, #340, #1100, #1115.

solomonneas:
Fixes #1479

## What

- `tests/test_agents.py`: dispatch fixtures no longer hard-code POSIX executable paths. A shared helper produces a platform-appropriate resolved executable, and tests that do not care about executable resolution stub that seam instead, so they exercise dispatch rather than the shim guard from #281 and #340.
- `tests/test_agents_windows_executable.py`: new classification tests for the three real Windows shapes, runnable on POSIX by feeding the classifier synthetic paths: a native `.exe`, an npm node wrapper `.exe` under `node_modules`, and a `.ps1` shim that is only acceptable with the per-seat command override from #1115.

This branch also carries the `env-file` fix for #1520: `ENV_FILE_REF_RE` required a leading slash, so Windows drive paths never matched and any seat resolving a secret through an `env-file:` reference could not start on Windows.

## Verification

Focused gate through Brigade (ruff, format, version sync, snapshots, mypy, both test files):

```
run: 20260907-141525-work-verify-b90a7e
status: completed  exit=0   (216 passed)
```

Windows confirmation on the fleet host is recorded in a follow-up comment.

### DIFF
--- a/src/brigade/agents.py
+++ b/src/brigade/agents.py
@@ -16,7 +16,7 @@ import unicodedata
 from collections import OrderedDict
 from collections.abc import Mapping
 from dataclasses import dataclass
-from pathlib import Path
+from pathlib import Path, PurePosixPath, PureWindowsPath
 from typing import Callable, List
 
 from . import proc
@@ -27,7 +27,13 @@ _CODEX_CLOUD_PREFIX = "codex-cloud:"
 _CLOUDFLARE_AI_GATEWAY_PREFIX = "cloudflare-ai-gateway/"
 _CLOUDFLARE_AI_GATEWAY_REQUIRED_ENV = ("CLOUDFLARE_ACCOUNT_ID", "CLOUDFLARE_GATEWAY_ID")
 ENV_FILE_REF_PREFIX = "env-file:"
-ENV_FILE_REF_RE = re.compile(r"^env-file:(/[^#]+)#([A-Z][A-Z0-9_]*)$")
+# Accept a POSIX absolute path ("/...") or a Windows drive-absolute path in
+# either separator form ("C:\..." or "C:/..."). UNC paths ("\\server\share" and
+# "//server/share") are deliberately rejected: nothing needs them and they
+# would widen the trust boundary. The pattern is only a first filter;
+# _parse_env_file_reference re-enforces absoluteness with PurePath checks.
+ENV_FILE_REF_RE = re.compile(r"^env-file:((?:/(?!/)[^#]+)|(?:[A-Za-z]:[/\\][^#]+))#([A-Z][A-Z0-9_]*)$")
+_ENV_FILE_DRIVE_ABSOLUTE_RE = re.compile(r"^[A-Za-z]:[/\\]")
 
 
 def is_env_file_reference(value: str) -> bool:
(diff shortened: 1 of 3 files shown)


solomonneas:
Windows confirmation on the fleet host (Python 3.14.6, pytest 9.1.1, basetemp outside the checkout), both files, `origin/main` vs this branch at 56b2e2c3:

| metric | main | this PR |
|---|---|---|
| failed | 126 | 1 |
| passed | 86 | 214 |

The tests were measuring the shim guard rather than dispatch, exactly as the issue describes. One failure remains, `test_run_agent_env_file_ref_uses_runtime_environment_file`, which is a separate Windows gap in environment-file handling rather than executable classification; it is being investigated and will either be fixed here or split into its own issue before merge.

solomonneas:
Updated Windows confirmation on the fleet host at c8c02e38, now including the `env-file` fix (#1520):

| file | main | this PR |
|---|---|---|
| `tests/test_agents.py` | 126 failed, 70 passed | 0 failed |
| `tests/test_agents_windows_executable.py` | 16 passed, 1 skipped | passes |
| `tests/test_roster_cmd.py` | 8 failed, 55 passed | 8 failed, unchanged |

Combined run of the three files: `8 failed, 272 passed, 1 skipped`. Every remaining failure is in `test_roster_cmd.py`, and the per-file baseline on `main` shows the identical 8 failures there, so this branch introduces none. Those are a separate Windows gap in roster doctor output and are not in scope here.

The single `test_run_agent_env_file_ref_uses_runtime_environment_file` failure reported earlier is fixed by the `env-file` commit on this branch: `ENV_FILE_REF_RE` required a leading slash, so a Windows drive path never matched and any seat resolving a secret through an `env-file:` reference could not start. That was a live problem for this fleet, whose Cursor seats use exactly that pattern.

PR #1537, opened 2026-09-07
fix(scanners): declare unsupported Windows descriptor operations

### LINKED ISSUE #1512: scanners: descriptor-relative inbox and rollback have no Windows path, so the module fails closed on every call

solomonneas:
Windows scanner descriptor operations now stop with one typed unsupported reason before binding run authority, opening an inbox transaction, or creating rollback locks. This is the explicit interim boundary until handle-relative inbox and rollback support exists.

Scanner init/run/history/doctor commands report the unsupported reason. General work doctor reports a warning without reading unavailable receipt state or suggesting every scanner is due. Read-only list/show/plan and configuration checks remain available. Operator init and quickstart skip the optional scanner setup with that reason and continue other setup. Work bootstrap reports a warning. Unexpected errors still fail. Descriptor-dependent tests carry individual capability skips using the same reason.

Validation:
- Final Brigade focused gate: 320 passed across operator, work session, scanner, and module-size tests. Receipt: `20260907-230950-work-verify-b4d948`.
- Same-host Windows baseline: 86 failed, 62 passed, 15 skipped. Final: 36 passed, 134 explicit platform skips, receipt 20260907-224710-work-verify-cb7092.
- Independent adversarial review approved after fixing early authority/rollback access and unsupported health scheduling. Conductor probes verified those fixes, receipt 20260907-224615-work-verify-577bb3.
- Minimal and full quickstart plus operator doctor passed on the same Windows host for both unchanged main and the repaired branch. The branch preserves the explicit unsupported reason. Receipts: baseline `20260907-231052-work-verify-dcf4b9`, repaired `20260907-230923-work-verify-00f949`.
- Bootstrap integration passed independent review. Direct probes confirmed unexpected errors still propagate, receipt `20260907-231126-work-verify-df3c41`. Full native acceptance runs in CI.
- scanners.py shrinks by five physical lines; no ratchet exemption.
- A broader doctor check found test_doctor_per_run_verdicts failing with absent instead of live lock state. The exact test also fails on the unchanged baseline branch; it is separate from this scanner change.

This documents unsupported Windows operations; it does not implement Windows descriptor-relative scanner writes.

Fixes #1512.


solomonneas:
@coderabbitai full review

